PR #10, opened 2020-08-10
Feat(Core)/Create did and did document parser 

tensor-programming:
DID parser, DID structure and DID document structure are setup in this PR.   May change DID Parser to Nom in the future when implementing the document parser; it should be more efficient.   Basic abstractions for Public Keys, Services, Subjects, and Context are also included in this PR. 

Unit testing and Property testing was implemented for all of the features with the proptesting focusing on the did parser.


JelleMillenaar (approved):
Quite a lot to read, but I noticed nothing wrong at first glace. Seems like an excellent base to start working from. Good job :)

PR #15, opened 2020-08-17
Feat(VC)/Create verifiable credentials

huhn511:
Implement [Verifiable Credentials Data Model 1.0](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-credentials)


JelleMillenaar:
Feature implemented by #20

PR #19, opened 2020-08-25
feat(core): Add diffing logic

tensor-programming:
Create logic to generate diff structures of the DID document to publish to the tangle.  Create a system for determining order in these diff documents and the original document and create an API to merge the diffed documents with the original document to create a full document which reflects the changes. 

Along with the Diffing logic, added changes to the API to make it smoother.  Also added the Authentication field to the DID document. Metadata stays as a undefined hashmap for now and will be updated as the shape of that data becomes more relevant.


identity_core/tests/authentication.rs
```diff
@@ -0,0 +1,546 @@
+use identity_core::{
+    document::DIDDocument,
+    utils::{Authentication, Context, PublicKey, Subject},
+};
+
+use std::str::FromStr;
+
+#[test]
+fn test_auth() {
+    let json_str = r#"
```
l1h3r, line 10: some of these large strings can go in files and be added through `include_str!`

identity_core/tests/authentication.rs
```diff
@@ -0,0 +1,546 @@
...
+        "publicKeyPem".into(),
+        "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----".into(),
+    )
+    .unwrap();
+
+    let key2 = PublicKey::new(
+        "did:iota:123456789abcdefghi#keys-2".into(),
+        "Ed25519VerificationKey2018".into(),
+        "did:iota:pqrstuvwxyz0987654321".into(),
+        "publicKeyBase58".into(),
```
l1h3r, line 68: maybe an easier API for this is having `KeyEncodingType` store the actual key data. thoughts?

```rust
#[derive(Deserialize, Serialize)]
#[serde(rename_all = "camelCase")]
enum KeyEncodingType {
  PublicKeyBase58(String),
  PublicKeyPem(String),
  ...
}
```

It makes the serialization a bit easier too, eg. `publicKeyJwk` is a map and can be auto-magically handled by serde

identity_core/src/did.rs
```diff
@@ -252,3 +211,11 @@ impl Serialize for DID {
         serializer.serialize_str(s.as_str())
     }
 }
+
+impl From<DIDTuple> for Param {
+    fn from(tuple: DIDTuple) -> Param {
+        let (name, value) = tuple;
```
l1h3r, line 217: 💅 you could use `(name, value)` in the function signaure: `fn from((name, value): DIDTuple) -> Param`

identity_core/src/document.rs
```diff
@@ -56,28 +66,111 @@ impl DIDDocument {
...
+    }
+
+    /// get the ID from the Document as a DID.
     pub fn derive_did(&self) -> crate::Result<DID> {
         self.id.to_did()
     }
+
+    /// Updates the `updated` time for the `DIDDocument`.
+    pub fn update_time(&mut self) {
+        self.updated = Some(Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true));
```
l1h3r, line 155: I've added a `Timestamp` type [here](https://github.com/iotaledger/identity.rs/pull/24/commits/fe047e0666fb43d54752242d69af12d27a55f0a2) to make this a little less repetitive and enforce the format.

That hasn't been merged yet so nothing to change for now

identity_core/Cargo.toml
```diff
@@ -21,6 +21,7 @@ anyhow = "1.0"
 serde = {version = "1.0", features = ["derive"]}
 serde_json = "1.0"
 serde_cbor = "0.11"
+serde-diff = "0.3.1"
```
l1h3r, line 24: The output of this is quite large...but I noticed you mentioned on discord that you were looking into some changes so 👍

l1h3r (approved):

identity_core/tests/authentication.rs
tensor-programming: Already done, in the latest version.

identity_core/tests/authentication.rs
tensor-programming: Have considered it especially since the jwt format has multiple fields associated with it.  For now I just wanted something minimal but will likely be making those changes later.

identity_core/src/did.rs
tensor-programming: Ahh yes, you are right, should have just destructed from the arguments.

identity_core/src/document.rs
tensor-programming: I saw, yes, that's a great addition.

identity_core/Cargo.toml
tensor-programming: The output is large if the json is not minified.  When its minified, its actually about the same as what I've been working on.  That said, I still want to have a bit more control over the diffing logic, so I will go through with my library and we can discuss it as a group this Wednesday.

identity_core/tests/authentication.rs
tensor-programming: Went ahead and did it and later when the common Object is merged, Ill wrap the JWT in it.

PR #28, opened 2020-09-01
feat(core, derive, diff)/ solidify api and diff

tensor-programming:
Diff crate and Diff derive macro crate.  Some refactoring is probably in order but all of the basic behavior is in place. Diff is implemented on all of the important types for our Identity Document structure and it is implemented on the `serde_json::Value` in the case where we need more flexibility.  Some work was done at the end to make sure that serialized empty `Options` (aka `Some(None)` becomes `None`) would be thrown out.  For now, this works completely on Struct structures but not completely on tuple structures.   It is likely that a manual serialize implementation is needed to flesh this out should we need those kinds of data structures in our other libraries.  

Structure and size of a diff can be found in the `document.json` file. When a `DiffDIDDocument` is serialized; parts of the document are converted back into their non-diff types.  When deserialized, they reverse their type back to the Diff Type.  The plan is to eventually try to allow the entire Document to be converted from a `DiffDIDDocument` into a normal `DIDDocument` upon serialization, but for now, there are some blockers to this and very few advantages.


identity_derive/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "identity_derive"
```
l1h3r, line 2: Might be better to name this `identity_diff_derive`, or is this meant to contain a variety of `derive`able behaviours?

l1h3r (commented):
Still going over the diffing stuff but looks pretty solid 👌 

What do you think of diffing the `DID` URLs as strings instead of objects? Would this be easy to do with the current implementation?

identity_core/Cargo.toml
```diff
@@ -21,7 +21,10 @@ anyhow = "1.0"
 serde = {version = "1.0", features = ["derive"]}
 serde_json = "1.0"
 serde_cbor = "0.11"
-serde-diff = "0.3.1"
+# serde-diff = "0.3.1"
```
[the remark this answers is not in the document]
tensor-programming: Yeah, will deal with it.

identity_core/src/did.rs
```diff
@@ -53,18 +51,18 @@ impl DID {
         parse(input)
     }
 
-    /// Method to add params to the DID.  
-    pub fn add_params(&mut self, params: Vec<Param>) {
-        let ps = match &mut self.params {
+    /// Method to add a query or query params to the DID.  The `value` field of the `Param` type can be None.
+    pub fn add_query(&mut self, query: Vec<Param>) {
+        let qur = match &mut self.query {
```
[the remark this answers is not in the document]
tensor-programming: Ahh yes, you are right.

identity_core/src/did_parser.rs
```diff
@@ -43,30 +43,35 @@ fn parse_pairs(pairs: Pairs<Rule>) -> crate::Result<DID> {
...
                 did.id_segments.push(pair.as_str().to_string());
             }
-            Rule::param => {
-                let mut inner = pair.into_inner();
-                let name = inner.next().expect("No name for this value");
+            Rule::path_segment => {
+                path_segs.push(pair.as_str().to_string());
+            }
+            Rule::query => {
+                let pairs = pair.clone().into_inner();
```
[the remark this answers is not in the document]
tensor-programming: Will look into it but not sure.

identity_core/src/did_parser.rs
```diff
@@ -43,30 +43,35 @@ fn parse_pairs(pairs: Pairs<Rule>) -> crate::Result<DID> {
...
+                for pair in pairs {
+                    let mut param = Param { ..Default::default() };
```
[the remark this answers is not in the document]
tensor-programming: No its not, bad habit of mine.

identity_core/src/document.rs
```diff
@@ -67,7 +71,16 @@ impl DIDDocument {
 
     /// sets a new `service` of type `Service` into the `DIDDocument`.
     pub fn update_service(&mut self, service: Service) {
-        self.services.push(service);
+        let mut services: Vec<Service> = self
```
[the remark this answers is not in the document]
tensor-programming: Little lost, is there an ordering that needs to occur here?

identity_derive/Cargo.toml
tensor-programming: It may have other derivation macros if we decide to add any.  At least that was the intent.

tensor-programming:
> Still going over the diffing stuff but looks pretty solid 👌
> 
> What do you think of diffing the `DID` URLs as strings instead of objects? Would this be easy to do with the current implementation?

That is something that I talked about in the meeting that I am currently working on.  Basically, I just need to pass a `serialize_with` expression into the macro in a `#[serde(serialize_with = ]` block and it should work.

identity_core/src/document.rs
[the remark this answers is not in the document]
l1h3r: Sorry, I made this comment the other day and thought I had deleted it - seemed a bit unnecessary

The general idea was a trait to express some comparable/hashable ID

```
trait HasId {
  type Id: Hash + PartialEq + Eq;
  fn id(&self) -> &Self::Id;
}
```

and a newtype to wrap values implementing that trait

```
pub struct IdCompare<T: HasId>(T);
```

These traits would be implemented using the value of `self.id()` through the `HasId` trait

```
impl Hash for IdCompare {}
impl PartialEq for IdCompare {}
impl Eq for IdCompare {}
```

and finally use it in `DIDDocument` definition

```
struct DIDDocument {
  ...
  services: HashSet<IdCompare<Service>>,
}
```

l1h3r:
> That is something that I talked about in the meeting that I am currently working on

🤦 ... please carry on 👍

identity_core/src/document.rs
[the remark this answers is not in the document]
tensor-programming: Oh I see, the purpose of the trait is to verify that the service has a unique id in the set.  I guess the fragment on the end of the service id must be unique.  Aye this is a good idea; I will implement it for this field.

tensor-programming:
> > That is something that I talked about in the meeting that I am currently working on
> 
> 🤦 ... please carry on 👍

I suspect that I am going to have to get your builder macro and use it in this lib.  Using the newtype/trait pattern with the init struct pattern is just too awkward.

identity_derive/src/impls/structs.rs
```diff
@@ -0,0 +1,662 @@
...
+                .enumerate()
+                .map(|(idx, field)| {
+                    let marker = &field_markers[idx];
+                    let typ = field.typ();
+
+                    if field.should_ignore() {
+                        quote! { Default::default() }
+                    } else {
+                        quote! {
+                            <#typ>::from_diff(
```
l1h3r, line 557:
```
#marker.map(|v| <#typ>::from_diff(v)).unwrap_or_default()
```

What about converting fields like this? Avoids the `to/from` diff conversion for default values.

identity_derive/src/impls/structs.rs
```diff
@@ -0,0 +1,662 @@
...
+                    }
+
+                    fn diff(&self, other: &Self) -> Self::Type {
+                        #diff
+                    }
+
+                    #[allow(unused)]
+                    fn from_diff(diff: Self::Type) -> Self {
+                        match diff {
+                           #diff => Self,
```
l1h3r, line 649: redundant `match`? also below

identity_derive/src/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use syn::{DeriveInput, Field, Meta, MetaList, NestedMeta, Path, PathSegment};
+
+use quote::format_ident;
+
+const PARENS: Delimiter = Delimiter::Parenthesis;
+
+/// checks to see if a field's type is `Option`.  This logic is necessary to find cases where fields contain nested
+/// Options and avoid a `Some(None)` case.
+pub fn extract_option_segment(path: &Path) -> Option<&PathSegment> {
```
l1h3r, line 11: `path.segments.last().filter(|segment| segment == "Option")`

I think this does the same and avoids the two allocations?

identity_derive/src/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+        .into_iter()
+        .find(|s| idents_of_path == *s)
+        .and_then(|_| path.segments.last())
+}
+
+/// checks to see if the `should_ignore` attribute has been put before a field.
+pub fn should_ignore(field: &Field) -> bool {
+    let mut attr_exists = false;
+
+    field.attrs.iter().for_each(|field| {
```
l1h3r, line 27: `find(|field| { ... })` should give you what you want and halts early

identity_derive/src/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        attr_exists = attr_exists || diff_attr && should_ignore
+    });
+
+    attr_exists
+}
+
+pub fn parse_from_into(input: &DeriveInput) -> bool {
+    let mut attr_exists = false;
+    input.attrs.iter().for_each(|a| {
```
l1h3r, line 50: `find(..)`

identity_diff/src/traits.rs
```diff
@@ -0,0 +1,20 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+
+/// The primary `Diff` Trait type.
+pub trait Diff: Clone + Debug + PartialEq {
+    /// The Corresponding Diff Type for the implemented Type.
+    type Type: Sized + Clone + Debug + PartialEq + for<'de> Deserialize<'de> + Serialize;
+
+    /// Finds the difference between two types; `self` and `other` and returns `Self::Type`
+    fn diff(&self, other: &Self) -> Self::Type;
```
l1h3r, line 10: These should probably all return `Result<T>` so `DiffError`s can be handled by the user (rather than panic).

something simple

`enum Error { DiffError, MergeError, ConvertError }`

identity_diff/src/vec.rs
```diff
@@ -0,0 +1,281 @@
...
+impl<T: Diff> Debug for InnerVec<T> {
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        match &self {
+            Self::Change { index, item } => f
+                .debug_struct("Change")
+                .field("index", index)
+                .field("item", item)
+                .finish(),
+            Self::Remove { count } => f.debug_struct("Remove").field("count", count).finish(),
+            Self::Add(diff) => write!(f, "Add({:#?})", diff),
```
l1h3r, line 136: Is there a `f.debug_` fn that could do this? Just to preserve formatting options

l1h3r (commented):
Overall this is really solid - nice job 💯

JelleMillenaar (commented):
Very happy with how this turned out. I only have a few questions. Does identity_derive have to be a top level folder? I am assuming derive is only a requirement for diff? Wouldn't it work better if derive is part of the diff folder then?  

Secondly, when I look at the diff examples in core, I would like to understand why some fields have an "Add" object, while others don't (For example the "id" or "context" objects).

tensor-programming:
> Very happy with how this turned out. I only have a few questions. Does identity_derive have to be a top level folder? I am assuming derive is only a requirement for diff? Wouldn't it work better if derive is part of the diff folder then?
> 
> Secondly, when I look at the diff examples in core, I would like to understand why some fields have an "Add" object, while others don't (For example the "id" or "context" objects).

1) The `identity_derive` crate needs to be its own crate since its a proc macro.  You can not unfortunately, add a proc macro to a different existing library in Rust and embedding the library into the `identity_diff` crate would be too complicated. 

2) The "Add" and "Remove" keys come from the fact that these fields are inside of a `HashSet` rather than a `Vec`. I could untag them but the problem is that the deserializer wouldn't be able to tell which field is a Remove and which is an Add.  I could try to change the shape of each and then untag them if that's what you want.

identity_core/src/document.rs
```diff
@@ -66,8 +73,17 @@ impl DIDDocument {
     }
 
     /// sets a new `service` of type `Service` into the `DIDDocument`.
-    pub fn update_service(&mut self, service: Service) {
-        self.services.push(service);
+    pub fn update_service(&mut self, service: IdCompare<Service>) {
+        let mut services: HashSet<IdCompare<Service>> = self
```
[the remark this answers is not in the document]
tensor-programming: Right you are.

identity_derive/src/utils.rs
tensor-programming: Issue is that the path segments need to be converted into a `String` before they can be compared to a `&str` since theres no `PartialEq` for `PathSegment` to `&str` or `String`.

JelleMillenaar:
> 1. The `identity_derive` crate needs to be its own crate since its a proc macro.  You can not, unfortunately, add a proc macro to a different existing library in Rust and embedding the library into the `identity_diff` crate would be too complicated.

Alright fair enough. No problem then.

> 2. The "Add" and "Remove" keys come from the fact that these fields are inside of a `HashSet` rather than a `Vec`. I could untag them but the problem is that the deserializer wouldn't be able to tell which field is a Remove and which is an Add.  I could try to change the shape of each and then untag them if that's what you want.

Its more important for the deserializer to understand it than humans, so that's alright. Similarly, please don't add fields to make it consistent, I'd rather keep it small. Could you document the ruleset on how the diff is generated though? This will be needed for the MethodSpec as others have to be able to generate the exact same diffs.

JelleMillenaar (approved):

tensor-programming:
> Its more important for the deserializer to understand it than humans, so that's alright. Similarly, please don't add fields to make it consistent, I'd rather keep it small. Could you document the ruleset on how the diff is generated though? This will be needed for the MethodSpec as others have to be able to generate the exact same diffs.

Yes, I can write them down and make a spec for them.   One thing though, might be worth putting the auth fields into the IdCompare newtype and as a result into HashSets, in which case all of the fields would have the Add and Remove keys (unless I changed the format which would be reasonable to do).  The main question is whether or not the auth fields need to have unique IDs.  I didn't do this because I knew that there could be a reference ID in the auth field, but am not sure if this reference field could have the same ID as another Auth Key in there.

PR #33, opened 2020-09-07
feat(crypto): Identity Proofs

l1h3r:
This PR introduces two new main crates:
  - `identity_crypto`
    - The place for all cryptographic utilities. Common traits, types, etc.
    - All other `identity_*` crates should rely on this crate for their crypto needs (prefer common traits over any concrete implementations)
  - `identity_ldp`
    - An implementation of Linked Data Proofs and various types/traits to implementation signature suites
    - Missing a few canonicalization algorithms but still usable with [canonical JSON](https://github.com/gibson042/canonicaljson-spec)

Concrete implementations of various behaviours can be implemented as needed (several LD signature suites will follow in a subsequent PR) and `identity_crypto` types and traits can be used immediately in definitions.

Supported proofs will likely extend beyond the scope of Linked Data Proofs; the `identity_crypto::Proof` trait is intended to reflect this.

---

<details>
<summary>Notes on LD Proofs</summary>
<p>

The [LD Proofs standard](https://w3c-ccg.github.io/ld-proofs/) is still changing at a fairly frequent pace and no concrete version has been released yet.

Some examples of unclear/inconsistent behaviour in the current standard:
  - input options expected by various signature suites are not clearly defined
  - properties in LD proof JSON object are not clearly defined/implemented
  - reference implementations using incorrect jws `alg` param (eg. RsaSignature2018)

[Cryptographic Suite Registry](https://w3c-ccg.github.io/ld-cryptosuite-registry/) and [Security Vocabulary](https://w3c-ccg.github.io/security-vocab/) are used to determine requirements/structure.

[Security Vocabulary v3 (unstable)](https://w3c-ccg.github.io/security-vocab/contexts/security-v3-unstable.jsonld)

Most registered suites are "provisional" but several settled on a JWS signatured-based proof structure - JWS is NOT the only way going forward and future JWS LD proofs will most likely all use `JwsSignature2020`.

> The JWS stuff was an experiment to see if we could unify the JOSE community and the Linked Data Security community. The attempt has failed for this cryptosuite, but the work is being carried on in JwsSignature2020... so we'd like to push all JOSE stuff to that cryptography suite and keep the Ed25519 stuff simpler and more auditable.

So there will be requirements to support JWS and non-JWS versions of the same cryptographic functions. Not too big of a deal as the `SignatureSuite` trait in this PR is flexible enough to support a variety of signature suite algorithms.

~Note: A few things are missing to fully support legacy signature suites~
  - ~`proofValue` should be one of `jws`, `proofValue`, or `signatureValue`. Signature suites need more control here~
  - ~Additional proof properties should be supported and "flattened" in the JSON representation. Trivial with `identity_common::Object`~

A high-level example of how this might be used with the `SignatureProof` wrapper and an implementation of `Ed25519Signature2018`.

```rust
// signer

let credential = ...
let seckey = ...

let klass = SignatureProof::new(Ed25519Signature2018);
// OR
let klass = SignatureProof::with_options(Ed25519Signature2018, my_options);

let proof = klass.create_proof(credential, seckey);
let json = serde_json::to_string(json);

// verifier

let credential ...
let pubkey = ...

let suite = ProofManager::get(credenital.proof["type"]);
let klass = SignatureProof::new(suite);
let valid = klass.verify(credential, credential.proof, pubkey);
```

</p>
</details>

---

<details>
<summary>Notes on cryptography libraries</summary>
<p>

- [Ursa](https://github.com/hyperledger/ursa)

  - Lots of useful functionality that **could** be used to build our own suites or for other use-cases besides LD Proofs. Only provides cryptographic support for 3 of the existing signature suites
  - Fairly large dependency chain and some versions outright failed to compile with stable rust
  - Functionality is heavily aimed towards supporting Hyperledger projects (eg. Indy)

Overall `ursa` is nice and easy to use but probably not worth the dependencies

- [Rust Crypto](https://github.com/RustCrypto)

  - Lots of great functionality in pure Rust
  - Missing some more exotic algorithms but provides solid building blocks

#### JSON Web Tokens

The state of JSON web tokens in Rust is not great...

[rust-jwt](https://github.com/mikkyang/rust-jwt) - Missing `ES256K` and more
[biscuit](https://github.com/lawliet89/biscuit) - Missing `ES256K` and more; based on `rust-jwt`
[jsonwebtoken](https://github.com/Keats/jsonwebtoken) - Missing `ES256K` and more
[josekit](https://github.com/hidekatsu-izuno/josekit-rs) - Incredibly new, `openssl`-based, and supports all algorithms currently needed

There are many more implementations, but these seemed to be the most complete/supported. I personally thought `josekit` looked like the most promising option.

---

### Update

Spent some time getting familiar with [`ring`](https://github.com/briansmith/ring) - it's pretty nice and I suspect it would cover most needs alongside [`Rust Crypto`](https://github.com/RustCrypto).

`jsonwebtoken` was missing support for a few minor things (the `crit` header param, signing non-`str` values) but already have open PRs; also nice that it also uses `ring`. Unfortunately it's missing three of the algorithms needed for some of the older jws signature suites - without modifications it would only provide support for `RsaSignature2018`. `ES256K` is missing due to lacking support for `secp256k1` in `ring` so that's probably a *bit* more involving.

~`jsonwebtoken` and only `RsaSignature2018` should be sufficient for now..~

</p>
</details>


tensor-programming:
Regarding potential cryptography libraries; when I was building stronghold, I found `sodiumoxide` to be pretty solid.  Uses the lib-sodium C library as its core.  `rust-crypto` is also pretty battle tested, but again not sure if it contains all of the algorithms we will need.  Another would be `ring` which is the most widely used rust crypto lib to my knowledge.  I've heard decent things about `josekit` but as you pointed out its pretty naive and immature right now.

l1h3r:
> I found `sodiumoxide` to be pretty solid

Thanks, will check it out - nice to see it supports `no_std`.

> `rust-crypto` is also pretty battle tested

Forgot to add a note on this but we'll definitely use it for building blocks and hopefully more

> Another would be `ring`

Cool, I'll check this out today 👍

l1h3r:
@tensor-programming This is ready for review now - Signatures are in another PR (https://github.com/iotaledger/identity.rs/pull/35) but that's still a draft. Ended up using `ring` and `sodiumoxide` instead of a JWT library

tensor-programming:
Alright, will look at it tomorrow.

identity_crypto/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+#[macro_use]
+extern crate identity_common;
```
tensor-programming, line 2: perhaps should import the macros with use.

identity_ldp/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+    InvalidLDSignature(String),
+    #[error(transparent)]
+    Common(#[from] identity_common::Error),
+    #[error(transparent)]
+    Crypto(#[from] identity_crypto::Error),
+    #[error(transparent)]
+    Custom(#[from] anyhow::Error),
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
```
tensor-programming, line 21: should use `anyhow::Result<T, E>` instead of the `std::result::Result<T, E>`. Might as well use anyhow since its a dependency.

identity_ldp/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+#[macro_use]
+extern crate anyhow;
```
tensor-programming, line 2: again, probably should use `use` for macros.

tensor-programming (commented):
everything looks good aside from a few structural issues.

JelleMillenaar (commented):
Looking at the DID standard, it wants to remove the dependency on LD-Proof. Yet the VC standard mentions it heavily. The VC Impl guide mentions both LD-Proof and JWT as separate ways to create proofs. If this implementations is ment to support both types of proofs, I recommend renaming this folder to identity_proofs and add support for JWT in this folder at a later stage. While the DID standard doesn't require it anymore, we do need the proofs for both DID and VC. Happy to discuss this further over chat.

tensor-programming:
Fixed the merge conflicts for you.

tensor-programming (approved):

JelleMillenaar (approved):

PR #38, opened 2020-09-24
Feat(core)/create method specificic id

Thoralf-M:


Thoralf-M:
Moved into https://github.com/iotaledger/identity.rs/pull/44

PR #42, opened 2020-09-30
Feat(bindings)/add wasm binding

huhn511:


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
rajivshah3, line 1: This file can be removed

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rajivshah3, line 1: This file can also be removed

libraries/wasm/package-lock.json
```diff
@@ -0,0 +1,5870 @@
+{
```
rajivshah3, line 1: This file should also be removed since it looks like you're using Yarn instead of NPM

tensor-programming:
Should add a `.gitignore` for all of the node stuff.

Cargo.toml
```diff
@@ -4,6 +4,10 @@ members = [
     "identity_crypto",
     "identity_derive",
     "identity_diff",
+    "identity_derive",
     "identity_iota",
     "identity_proof",
+    "libraries/wasm"
 ]
+exclude = [
```
l1h3r, line 12: I think `libraries/wasm` should be excluded from the workspace

identity_iota/src/did/document.rs
```diff
@@ -235,7 +235,7 @@ impl IotaDocument {
         create_address_from_trits(Self::create_diff_address_hash(public_key))
     }
 
-    fn check_authentication_key_id(authentication: &PublicKey, did: &IotaDID) -> Result<()> {
+    pub fn check_authentication_key_id(authentication: &PublicKey, did: &IotaDID) -> Result<()> {
```
l1h3r, line 238: does this **need** to be public?

libraries/wasm/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+authors = ["Team Identity"]
+edition = "2018"
+repository = "https://github.com/iotaledger/identity.rs"
+homepage = "https://www.iota.org"
+license = "MIT"
+
+[lib]
+crate-type = ["cdylib"]
+
+# TODO: Discover this:
```
l1h3r, line 13: ??

libraries/wasm/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+console_error_panic_hook = "0.1"
+iota-constants = { version = "0.2", default-features = false }
```
l1h3r, line 21: unused dependency?

libraries/wasm/src/did.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    #[wasm_bindgen(js_name = toString)]
+    pub fn to_string(&self) -> String {
+        self.0.to_string()
+    }
+}
+
+impl DID {
+    pub fn parse_from_did(did: CoreDID) -> Result<DID, JsValue> {
```
l1h3r, line 77: This can be part of the same `impl DID` block above

libraries/wasm/src/did.rs
```diff
@@ -0,0 +1,80 @@
...
+        Ok(Self(IotaDID::parse(did.to_string()).map_err(js_err)?))
```
l1h3r, line 78: any reason this converts the DID to a string before running method-specific validation via `IotaDID`?

```suggestion
        IotaDID::try_from_did(did).map_err(js_err).map(Self)
```

libraries/wasm/src/doc.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub fn update_service(&mut self, did: WasmDID, url: String, service_type: String) -> Result<Doc, JsValue> {
+        let endpoint = ServiceEndpoint::Url(url.parse().map_err(js_err)?);
+        let service = ServiceBuilder::default()
+            .id(DID::parse(did.0).map_err(js_err)?)
+            .service_type(service_type)
+            .endpoint(endpoint)
+            .build()
+            .map_err(js_err)?;
+        self.0.update_service(service);
+        Ok(Doc(self.0.clone()))
```
l1h3r, line 180: Any reason this modifies the current document AND returns a *new* document? (same with functions below)

It might be nicer to `clone` first and leave the original document unmodified

libraries/wasm/src/doc.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    #[wasm_bindgen]
+    pub fn update_time(&mut self) {
+        self.0.update_time();
+    }
+
+    #[wasm_bindgen]
+    pub fn resolve_key(&mut self, key_relation: &str) -> Result<PubKey, JsValue> {
+        let relation = match key_relation {
```
l1h3r, line 238: Should be able to use `serde` for this if you add `#[derive(Deserialize)]` to `KeyRelation`

libraries/wasm/src/doc.rs
```diff
@@ -0,0 +1,260 @@
...
+            "VerificationMethod" => KeyRelation::VerificationMethod,
+            "Authentication" => KeyRelation::Authentication,
+            "AssertionMethod" => KeyRelation::AssertionMethod,
+            "KeyAgreement" => KeyRelation::KeyAgreement,
+            "CapabilityInvocation" => KeyRelation::CapabilityInvocation,
+            "CapabilityDelegation" => KeyRelation::CapabilityDelegation,
+            _ => panic!("Unknown KeyRelation"),
+        };
+
+        match self.0.resolve_key(0, relation) {
```
l1h3r, line 248: you can use `expect` to "unwrap or panic" with a message

`self.0.resolve_key(0, relation).expect("Key not found")`

libraries/wasm/src/iota.rs
```diff
@@ -0,0 +1,97 @@
...
+        .await
+        .map_err(js_err)
+        .and_then(|response| JsValue::from_serde(&response).map_err(js_err))
+}
+
+#[wasm_bindgen(js_name = checkCredential)]
+pub async fn check_credential(data: String, params: JsValue) -> Result<bool, JsValue> {
+    let client = client(params)?;
+    let validator: CredentialValidator<'_> = CredentialValidator::new(&client);
+    let validation: CredentialValidation = validator.check(&data).await.map_err(js_err)?;
```
l1h3r, line 95: Can this return the entire validation object? Might need to add `Serialize` to the type

libraries/wasm/src/did.rs
Thoralf-M: I tried it, but then I got ```error[E0277]: the trait bound `identity_core::did::DID: wasm_bindgen::convert::FromWasmAbi` is not satisfied
  --> libraries\wasm\src\did.rs:18:1
   |
18 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::FromWasmAbi` is not implemented for `identity_core::did::DID`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)```

l1h3r: ahh yeah, that make sense - we don't have `#[wasm_bindgen]` on `identity_core` types

👍 ignore my bad suggestion

libraries/wasm/src/doc.rs
Thoralf-M: Right, thanks for the suggestion

libraries/wasm/src/iota.rs
Thoralf-M: Yes, needed also `Serialize` for the other types in there

libraries/wasm/src/doc.rs
Thoralf-M: How should `serde` be used there then?

l1h3r: I should have suggested `FromStr` - it would look like this:

```rust
use core::str::FromStr;

impl FromStr for KeyRelation {
  type Err = crate::Error;

  fn from_str(string: &str) -> Result<Self, Self::Err> {
    match string {
      "VerificationMethod" => Ok(Self::VerificationMethod),
      "Authentication" => Ok(Self::Authentication),
      "AssertionMethod" => Ok(Self::AssertionMethod),
      "KeyAgreement" => Ok(Self::KeyAgreement),
      "CapabilityInvocation" => Ok(Self::CapabilityInvocation),
      "CapabilityDelegation" => Ok(Self::CapabilityDelegation),
      _ => Err(...)
    }
  }
}
```

and you would use it like:

```rust
let relation: KeyRelation = key_relation.parse()?;
```

libraries/wasm/Cargo.toml
Thoralf-M: Probably the wasm-opt thing, I tried it with other values, but it looks like it's not supported atm and it's currently recommended just to set it to false

tensor-programming (approved):

libraries/wasm/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "files": [
+    "web",
+    "node",
+    "web.js"
+  ],
+  "devDependencies": {
+    "mocha": "^7.1.2",
+    "@wasm-tool/wasm-pack-plugin": "^1.1.0",
+    "copy-webpack-plugin": "^5.0.3",
```
JelleMillenaar, line 41: This is an old version

libraries/wasm/README.md
```diff
@@ -0,0 +1,80 @@
...
+$ npm install copy-webpack-plugin --save-dev
+// or using yarn
+$ yarn add copy-webpack-plugin --dev
+```
+
+- Add the copy plugin to the `plugins` array of your webpack config:
+```js
+new CopyWebpackPlugin([
+  {
+    src: 'node_modules/iota-identity-wasm-test/web/iota_identity_wasm_bg.wasm',
```
JelleMillenaar, line 64: Please update syntax to support CopyWebpackPlugin 6.x+

libraries/wasm/README.md
```diff
@@ -0,0 +1,80 @@
...
+- Install `copy-webpack-plugin`:
+```bash
+$ npm install copy-webpack-plugin --save-dev
+// or using yarn
+$ yarn add copy-webpack-plugin --dev
+```
+
+- Add the copy plugin to the `plugins` array of your webpack config:
+```js
+new CopyWebpackPlugin([
```
JelleMillenaar, line 62: Missing line of code:
const CopyWebPlugin= require('copy-webpack-plugin');

libraries/wasm/README.md
```diff
@@ -0,0 +1,80 @@
...
+  {
+    src: 'node_modules/iota-identity-wasm-test/web/iota_identity_wasm_bg.wasm',
+    to: 'iota_identity_wasm_bg.wasm'
+  }
+])
+```
+
+### Usage
+```js
+import identity from "iota-identity-wasm-test/web";
```
JelleMillenaar, line 72: import * as IotaIdentity from "iota-identity-wasm-test/web"; 

Works for me without the need to do identity().then((lib)) stuff. This is in a .ts file though.

libraries/wasm/src/vc.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct VerifiableCredential(pub(crate) IotaVC);
+
+#[wasm_bindgen]
+impl VerifiableCredential {
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        issuer_doc: &Doc,
+        issuer_key: &Key,
+        credential_id: String,
```
JelleMillenaar, line 20: Please make the Id optional.

libraries/wasm/examples/node.js
```diff
@@ -0,0 +1,160 @@
...
+
+  doc.sign(key)
+
+  console.log("doc (signed)", doc.toJSON())
+
+  console.log("doc valid?", doc.verify())
+
+  const json = JSON.stringify(doc.toJSON())
+
+  console.log("From JSON >", Doc.fromJSON(json))
```
JelleMillenaar, line 40: fromJSON is not actually loading from a JSON object, but from a string that contains a valid JSON. Causes confusion as I did JSON.parse() before calling the function. Either rename or rework the logic to be more consistent with expectations.

libraries/wasm/README.md
```diff
@@ -0,0 +1,80 @@
...
+## Web Setup
+
+The library loads the WASM file with an HTTP GET request, so the .wasm file must be copied to the root of the dist folder.
+
+### Rollup
+- Install `rollup-plugin-copy`:
+```bash
+$ npm install rollup-plugin-copy
+// or using yarn
+$ yarn add rollup-plugin-copy
```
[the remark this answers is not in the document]
Thoralf-M: Yes it can

l1h3r (approved):

PR #44, opened 2020-10-06
Organize and consolidate crates

l1h3r:
### TODO

https://hackmd.io/2IVv4WgyR0SgOZ5C9IT29g


identity_iota/src/io/writer.rs
```diff
@@ -0,0 +1,189 @@
...
+            address: create_address(&did)?,
+            value: 0,
+            message: Some(to_string(&data).map_err(core::Error::EncodeJSON)?),
+            tag: None,
+        };
+
+        // Dispatch the transfer to the network
+        let mut bundle: Vec<BundledTransaction> = self.client.send(None).transfers(vec![transfer]).send().await?;
+
+        // Extract the response transaction - there should be exactly one.
```
Thoralf-M, line 123: Why should there only be one transaction? If the document has a lot content, the size can exceed a single transaction

JelleMillenaar: Agreed, even in Chrysalis P2, where transactions can be pretty massive. You can't make any assumptions.

l1h3r: 😄  fixed - This was just supposed to grab the tail transaction from the list

PR #46, opened 2020-10-08
fix(ci): Use env files for setting env vars

rajivshah3:
The `set-env` command is now deprecated (see [GitHub blog](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/))﻿


l1h3r (approved):

l1h3r:
Thanks @rajivshah3 👍

PR #47, opened 2020-10-13
add resolve example, have diffs on another address + small changes

Thoralf-M:


identity_core/src/did/did.rs
```diff
@@ -67,7 +67,7 @@ impl DID {
         })
     }
 
-    pub fn parse_from_str<T>(input: T) -> crate::Result<Self>
+    pub fn from<T>(input: T) -> crate::Result<Self>
```
l1h3r, line 70: Maybe `try_from_str` or `parse` but not `from`. [`From`](https://doc.rust-lang.org/core/convert/trait.From.html) is an extremely common trait and this will be confusing.

Also, from the docs of `From` -> `This trait must not fail`
It's not really expected that this conversion will return `Result<Self>` (but that would be expected of `try_from/try_into`)

You could add an implementation of `TryFrom<T>` that delegates to `parse_from_str` (that would be useful to pair with `derive_builder`s `try_setter` attribute) - https://rust-lang.github.io/api-guidelines/interoperability.html#conversions-use-the-standard-traits-from-asref-asmut-c-conv-traits

There is also currently an implementation of `FromStr` so it's already possible to do `"did:iota:main:1234".parse()?`

Thoralf-M: I wanted to remove `str` because it also works with `String` and not only with `str` or is this always the case if `str` is supported?

l1h3r: They are similar. [`str`](https://doc.rust-lang.org/std/primitive.str.html) is a primitive type; usually has to be borrowed so you see it as `&str`.

[`String`](https://doc.rust-lang.org/std/string/struct.String.html) is a struct with ownership of the contents. This is all it is:
```rust
pub struct String {
  vec: Vec<u8>,
}
```

and it implements [`Deref<Target = str>`](https://doc.rust-lang.org/std/string/struct.String.html#deref-methods) so you can use `String` in most places where `&str` is required by de-referencing and re-borrowing -> `&*string` (You can see this in `did::parser::parse` with `input_str`).

l1h3r: It really looks like `did::parser::parse` should use `T: AsRef<str>` instead of `T: ToString` (and also the public `DID` parsing method(s)) - This will avoid an allocation (from `ToString`) and use the string slice via `.as_ref()`.

l1h3r: It's probably not worthwhile to make too many parser changes as I was planning on replacing it with a new lib when moving to an updated DID standard - hopefully in the near future

Thoralf-M: I changed the name to `parse` now

identity_iota/src/did/proof.rs
```diff
@@ -16,7 +16,7 @@ pub struct DIDProof {
 impl DIDProof {
     pub fn new(id: DID) -> Self {
         Self {
-            id,
+            id: id.fragment.expect("DID without fragment provided"),
```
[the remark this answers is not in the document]
l1h3r: Yes, it should

JelleMillenaar (changes requested):
We decided to keep the deactivate DID Document functionality at the High-level library. I'd suggest moving this to the iota-client PR.

JelleMillenaar (approved):

tensor-programming (approved):

PR #51, opened 2020-10-19
Feat/iota client

l1h3r:
This PR adds a `Client` to replace `TangleReader`, `TangleWriter`, and `TangleResolver`.

Note: diff chain logic was removed while the spec is being finalized


identity_iota/src/did/did.rs
```diff
@@ -17,8 +17,8 @@ pub fn method_id(did: &DID) -> Result<&str> {
 
 /// Creates an 81 Trytes IOTA address from the DID
 pub fn create_address_hash(did: &DID) -> Result<String> {
-    let digest: &[u8] = &Blake2b256::digest(method_id(did)?.as_bytes());
-    let encoded: String = encode(digest).into_string();
+    let digest: MultihashGeneric<_> = Blake2b256::digest(method_id(did)?.as_bytes());
```
Thoralf-M, line 20: I think we should remove this hashing part here and use the method_id right away

identity_iota/src/did/did.rs
```diff
@@ -31,17 +31,17 @@ pub fn create_address(did: &DID) -> Result<Address> {
 }
 
 /// Creates an 81 Trytes IOTA address from public key bytes for a diff
-pub fn create_diff_address_hash(public_key: &[u8]) -> Result<String> {
-    let hash = &Blake2b256::digest(public_key);
-    let second_hash = &Blake2b256::digest(hash.digest());
-    let encoded: String = encode(second_hash.digest()).into_string();
+pub fn create_diff_address_hash(did: &DID) -> Result<String> {
```
Thoralf-M, line 34: And here we should rather use the authentication key instead of the method_id from the DID, because otherwise the address/tag will be the same for every diff, even if we change the authentication key

Thoralf-M:
publish_read example fails because the created DIDDocument doesn't have an `updated` value

tensor-programming (approved):

tensor-programming:
Nothing stands out as something that could be implemented better and I don't see any issues in the code.  The only minor thing would be to remember that eventually some of these functionalities will end calling through stronghold and the accounts module. I guess also the DerefMut implementation should be avoided if possible though you've got a `TODO` to remove it anyhow. It also may be worth using `derive_builder` for the client if it won't increase the size significantly.

PR #53, opened 2020-10-21
Add FromJson/ToJson convenience traits

l1h3r:
This is just splitting up `AsJson` so traits can be used with more relaxed bounds


JelleMillenaar (approved):
LGTM

PR #56, opened 2020-10-21
JCS Ed25510 Signature 2020

l1h3r:


JelleMillenaar (approved):
LGTM. Would it be useful to back up some of this code if we ever do want to support JSON-LD? I suppose Github does this for us.

l1h3r:
Yeah, we can always refer to the old PR and most of that code was moved to `crypto.rs` or `libjose`.

PR #58, opened 2020-10-22
Add IotaDocument type

l1h3r:


identity_iota/src/did/document.rs
```diff
@@ -1,145 +1,193 @@
...
+        }
 
-        Ok(())
+        Ok(Self { document })
     }
 
-    fn verify_diff_unchecked(&self, diff: &DIDDiff) -> Result<()> {
-        // TODO: Validate diff.id
-        // TODO: Validate diff.prevMsg
+    pub fn new(did: IotaDID, authentication: PublicKey) -> Result<Self> {
```
l1h3r, line 48: `authentication` should be validated and `id/controller` should line up with `did`

PR #60, opened 2020-10-27
Chore/vc cleanup

l1h3r:


identity_iota/examples/credential.rs
```diff
@@ -0,0 +1,154 @@
...
+    key: KeyPair,
+    name: String,
+}
+
+impl User {
+    async fn new(name: impl Into<String>, client: &Client) -> Result<Self> {
+        // Generate a DID/keypair
+        let (mut did, key): (IotaDID, KeyPair) = IotaDID::generate_ed25519()?;
+
+        did.id_segments.insert(0, "com".into());
```
JelleMillenaar, line 50: Please use mainnet in examples

identity_iota/examples/credential.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        // Create a signed DID document
+        let pkey: PublicKey = create_ed25519_key(&did, key.public().as_ref())?;
```
JelleMillenaar, line 53: What does this do? Format it? Either needs a dedicated comment or new name. We have generate ed25519 and create ed25519, how do they differ?

identity_iota/examples/credential.rs
```diff
@@ -0,0 +1,154 @@
...
+struct User {
+    doc: IotaDocument,
+    key: KeyPair,
+    name: String,
+}
+
+impl User {
+    async fn new(name: impl Into<String>, client: &Client) -> Result<Self> {
+        // Generate a DID/keypair
+        let (mut did, key): (IotaDID, KeyPair) = IotaDID::generate_ed25519()?;
```
JelleMillenaar, line 48: Rename key to keypair

identity_iota/examples/credential.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        doc.sign(key.secret())?;
+
+        // Publish the document
+        let response: PublishDocumentResponse = client.create_document(&doc).trace(true).send().await?;
+
+        let printer = TransactionPrinter::hash(&response.tail);
+
+        println!("[+] Doc > {:#}", doc);
+        println!("[+]   https://comnet.thetangle.org/transaction/{}", printer);
```
JelleMillenaar, line 64: Please use the new explorer.iota.org

identity_iota/examples/credential.rs
```diff
@@ -0,0 +1,154 @@
...
+#[smol_potat::main]
+async fn main() -> Result<()> {
+    let client: Client = ClientBuilder::new()
+        .network(Network::Comnet)
+        .node("https://nodes.comnet.thetangle.org:443")
+        .build()?;
+
+    let issuer: User = User::new("Issuer", &client).await?;
+    let subject: User = User::new("Subject", &client).await?;
+    let credential: VerifiableCredential = issuer.issue(&subject)?;
```
JelleMillenaar, line 118: rename to verifiableCredential or vc (Just to remain accurate)

JelleMillenaar (changes requested):

identity_iota/examples/credential.rs
l1h3r: This creates a verification method object - the name isn't great.

After https://github.com/iotaledger/identity.rs/pull/64 we'll have `IotaDocument::generate_ed25519` (or similar) and this function will be removed.

I'll fix the comment since it also seems to be wrong and the function will be removed soon

PR #62, opened 2020-10-27
Feat(core): Add manual implementation of Diff on DID. 

tensor-programming:
Adds a manual implementation of the diff trait on the DID struct to allow the method name to pass through each diff.


l1h3r (changes requested):
Still have a few clippy warnings but otherwise 👍

tensor-programming:
Will get to them.

l1h3r (approved):

PR #64, opened 2020-10-28
Update IotaDID/IotaDocument constructors

l1h3r:


identity_iota/src/did/document.rs
```diff
@@ -30,8 +30,21 @@ pub struct IotaDocument {
 }
 
 impl IotaDocument {
-    pub fn generate_ed25519_keypair() -> KeyPair {
-        jcsed25519signature2020::new_keypair()
+    pub fn generate_ed25519<'a, T>(tag: &str, network: T) -> Result<(Self, KeyPair)>
```
Thoralf-M, line 33: Why do we want the tag there? Shouldn't everyone just use key-1 for the first key?

Thoralf-M: At least if this is intended for the high level usage I would remove the tag

l1h3r: > Shouldn't everyone just use key-1 for the first key?

I don't think we need to impose a restriction on the actual naming of the identifier? Naming your key isn't really a "low-level" action, is it?

It could be optional and this could be callable like `generate_ed25519(None, None)`

Maybe the builder pattern would've been better here, unfortunately this configuration can't be derived but that's a minor loss.

```
let _ = DocumentBuilder::new().build()?;

let _ = DocumentBuilder::new().tag("my-custom-tag").build()?;
```

Thoralf-M: I just think that usually one probably don't care what's the name of it, but if you want to use the lib and aren't really familiar how it works you don't know for what the tag is right away, so I prefer to have it as Option or with the builder pattern
Didn't want to have this as restriction, of course you should still be able to set it on your own, just wanted to have it as simple as possible

JelleMillenaar: Low-level the tag can indeed be a requirement. High-level can keep track of an index and automate the naming logic.

Thoralf-M (approved):

Issue #72, opened 2020-11-10
Handle "main" network as default more consistently


JelleMillenaar:
A DID Document that is created without a network defaults to the mainnet called "main". The main part is left out of the id and keys. Such that they are did:iota:uuid instead of did:iota:main:uuid. However, if I input "main" into the generate key function, it does add main into the did's such that they render did:iota:main:uuid. Even though they work the same, this creates an inconsistent DID Document. I suggest we handle main as a default input better such that no input or "main" as network input is handled the same.

l1h3r:
Which version is considered the canonical version, `did:iota:uuid` or `did:iota:main:uuid`? (which should be in the document)

JelleMillenaar:
Inside a DID Document it should be stored as `did:iota:uuid` with the assumption that with no network found the mainnet is assumed as is described in the DID Method Specification.

l1h3r:
This was resolved in #73

PR #73, opened 2020-11-11
Don't add ":main" to a DID

Thoralf-M:
#72


l1h3r (changes requested):
I think it would be easier to normalize the DIDs whenever appropriate (we already have a function for DIDs and could add one to normalize documents as well).

If going down the route of preventing people from explicitly specifying `main`, this needs to:

- Remove `normalize/normalized` from `IotaDID` since these will add it back
- Make sure `main` is removed from DIDs created through `try_from_did/parse`
- Note: You can ignore `DerefMut` for now, it will be removed soon (allows direct modification of the `id_segments` field)

Thoralf-M:
https://github.com/iotaledger/identity.rs/blob/87c9ab6c26ff77717bcfbae74973e56c897bc522/identity_iota/src/did/did.rs#L134
So remove "main:" in the `normalize` method and call `normalized` in `try_from_did`?

l1h3r:
> So remove "main:" in the normalize method and call normalized in try_from_did?

Pretty much, it would be better to use `normalize` though and avoid the clone.

Might actually be a good idea to remove `normalized` (returns new clone) and keep `normalize` (mutates self). The two methods can get confusing and it's trivial to call `.clone()` when you need the behaviour of `normalized`.

identity_iota/src/did/did.rs
```diff
@@ -131,8 +131,19 @@ impl IotaDID {
 
     pub fn normalize(&mut self) {
         match &*self.id_segments {
-            [_] => self.id_segments.insert(0, Self::NETWORK.into()),
-            [_, _] | [_, _, _] => {}
+            [_] => {}
+            [network, _] => {
```
l1h3r, line 135: you can use the same style of pattern as before -> `[network, _] | [network, _, _] => {`

Thoralf-M: Thanks, forgot the shard DID the first time that's why I had them separated

l1h3r (approved):

identity_iota/src/did/did.rs
```diff
@@ -133,18 +137,16 @@ impl IotaDID {
 
     pub fn normalize(&mut self) {
         match &*self.id_segments {
-            [_] => self.id_segments.insert(0, Self::NETWORK.into()),
-            [_, _] | [_, _, _] => {}
+            [_] => (),
+            [network, _] | [network, _, _] => {
+                if let "main" = network.as_str() {
```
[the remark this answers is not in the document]
Thoralf-M: Yes

PR #82, opened 2020-11-16
fix/identity_iota DID creation

Thoralf-M:


Thoralf-M:
remove main: also when with_network_and_shard is used

l1h3r (approved):

PR #101, opened 2020-12-17
libjose updates #2

l1h3r:
* JWS Signers/Verifies
* JWE Encrypters/Decrypters
* JWK thumbprint


nothingismagick:
I still feel like this entire lib would be better hosted in crypto.rs, or you would source your needs from there - here it seems that no one is really reviewing your work, and MOST of the primitives / algos you are using are already available in crypto.rs

l1h3r:
I definitely agree that it's time to start using crypto.rs here (will start soon) - I haven't thought too much about moving the lib into crypto.rs (as feature) but that's a nice idea 👍

PR #103, opened 2021-01-07
Add documentation page

huhn511:
Documentation page for IOTA Identity

Deploys `/docs` content to http://identity.docs.iota.org on `dev` branch changes.


JelleMillenaar (approved):

nothingismagick:
Something I have been wondering is why its called libraries and not bindings?

huhn511:
Good question! My thoughts about this: 

A "binding" is used to "bind" a function from one programming languages to another, right?

So "bindings" are a collection of many of these functions. A library is also defined, as a collection of functions - and an often used term.

So the idea here is to tell the readers: "hey developers, here are the libraries for your programming languages". They don't need to know, that we did binded them in the background.

What do you think?

nothingismagick:
Compare this: https://github.com/shadowmint/rust-plibsys-sys with https://github.com/saprykin/plibsys

# plibsys-sys
This is a rust binding for the excellent plibsys library:

My point is that there is an original library, that is in this case the Rust Library. It is the canonical source of code truth. Then you have foreign function interfaces or bindings that are technically libraries, but your approach here would put rust and nodejs and java and python all on the same level, which in my opinion is incorrect.

nothingismagick:
Further complicating matters, at least in the case of stronghold (sorry for threadjacking @JelleMillenaar ) is that we have a number of libraries within the project. So I as a user would expect to find documentation about the individual libraries. In the case of stronghold, the bindings documentation will be MUCH, MUCH leaner than the libraries documentation, for the simple purpose that we will never expose all the library internals to the bindings (or the top level client for that matter.)

huhn511:
`but your approach here would put rust and nodejs and java and python all on the same level, which in my opinion is incorrect.`

I agree, technically its not correct to bring it on the same level - but for the usage level, there is no difference, right? So developers just need to know, which programming languages we are supporting (not how we do this).

For Example: As an javascript developer, i just install it via "npm install" and want to get started and use the library. I don't want to know, how it's works in the background or even how we build it (and if they really want, it's described in the repository README's).

If i did miss something here, i just will replace "libraries" with "bindings" to prevent further complications :)

nothingismagick:
I think you need both, and my point stands. Just because something is documented in a library does not mean that a binding is required to have the same interface. That is a design decision and not prescriptive.

Just look here: https://github.com/iotaledger/identity.rs/blob/c1e7b2024ca3f480778cff0dd1551d9a1bdced68/identity-core/Cargo.toml#L2

That is a library. And it should be explicitly documented in `/docs/libraries`

Currently, and I am assuming its just temporary, you are not, in fact, pulling in the `identity-diff` crate here: https://github.com/iotaledger/identity.rs/blob/c1e7b2024ca3f480778cff0dd1551d9a1bdced68/bindings/wasm/Cargo.toml#L18-L19

So my argument stands: 
- one folder for bindings
- one folder for libraries

huhn511:
I just found this: https://github.com/rust-lang/rust-bindgen

> `bindgen` automatically generates Rust FFI bindings to C (and some C++) libraries.

`rust-lang` also use the wording "libraries" for the result of the Rust FFI bindings. 

--- 

Okay, having both is nice :) 

But not sure yet, what i should add into the `bindings` folder. 

Where comes the docs.rs documentation into the game? Is this maybe the missing part of documentation here?

nothingismagick:
I think you are misunderstanding a little bit here how they are using the word library. In a C library you would attach to the rust bindings prepared for C.

But I do agree, I think that having the rustdocs in the /libraries section would be an easy win. Not sure how best to wire that up though.

l1h3r (approved):
👍 for some distinction between bindings and libraries - better done sooner to avoid breaking URLs

Issue #110, opened 2021-01-19
Add Merkle Tree Verification Method


l1h3r:
Provide a method for representing large quantities of cryptographic keys inside a DID Document.

#### TODO

- [ ] [Spec](https://github.com/iotaledger/identity.rs/pull/112)
- [x] [Merkle Tree Utility](https://github.com/iotaledger/identity.rs/pull/111)
- [ ] Sign/Verify with DID Document
  - [x] [Part 1](https://github.com/iotaledger/identity.rs/pull/109)
  - [ ] [Part 2](https://github.com/iotaledger/identity.rs/pull/112)

PR #112, opened 2021-01-19
Add Merkle Key Collection

l1h3r:
This PR adds the [specification](https://github.com/iotaledger/identity.rs/blob/578ec6e4119263f32260043298043fbe35eb5e7c/docs/specs/merkle-key-collection.md) and implementation of the "Merkle Key Collection" DID verification method.

This is essentially a single verification method representing an arbitrary number of cryptographic keys with built-in support for individual key revocation.

Closes https://github.com/iotaledger/identity.rs/issues/110


examples/merkle_key.rs
```diff
@@ -0,0 +1,146 @@
...
+    .controller(controller.clone())
+    .key_type(MethodType::MerkleKeyCollection2021)
+    .key_data(MethodData::new_b58(&key_data))
+    .build()
+    .unwrap();
+
+  println!("method: {:#}", method);
+
+  // Create a DID Document with the `MerkleKeyCollection2021` method
+  // as the sole verification method
```
JelleMillenaar, line 70: The example currently doesn't write to or read from the Tangle. While it does test all the appropriate functionalities, this is not how it would be utilized in the real world. I suggest we update the example to make use of the Tangle. This also means we can't add this as the "sole verification method", but would need to add a normal authentication-key for signing the DID Documents.

l1h3r: 👍 I'll add some more realistic Tangle usage to the example

JelleMillenaar (approved):

PR #125, opened 2021-02-18
fix(bindings/wasm): Use npmignore so that WASM binaries are included

rajivshah3:
When `.npmignore` does not exist, NPM will use the `.gitignore` instead. Since the `web` and `node` folders are gitignored, they won't be present in the NPM package. This PR creates a `.npmignore` so that they are included.


l1h3r (approved):
🚀

Issue #138, opened 2021-03-01
Stronghold Wrapper


l1h3r:
Add a higher-level API for Stronghold that handles actor management and provides conveniences for commonly used patterns

- Based on [tauri-plugin](https://github.com/tauri-apps/tauri-plugin-stronghold)
- Based on [wallet.rs](https://github.com/iotaledger/wallet.rs)
- [x] Add high-level Stronghold wrapper (see tauri-plugin)
- [x] Add `Store`-like API for indexed records (see wallet.rs)

Ideally we would be able to share this with wallet.rs and possibly tauri-plugin-stronghold

PR #148, opened 2021-03-02
Stronghold Wrapper

l1h3r:
Closes https://github.com/iotaledger/identity.rs/issues/138


identity-account/src/stronghold/result.rs
```diff
@@ -0,0 +1,44 @@
...
+  SLIP10Generate,
+  SLIP10Derive(ChainCode),
+  BIP39Recover,
+  BIP39Generate,
+  BIP39MnemonicSentence(String),
+  Ed25519PublicKey([u8; 32]),
+  Ed25519Sign([u8; 64]),
+}
+
+impl PleaseDontMakeYourOwnResult<ProcedureResult> for ProcResult {
```
nothingismagick, line 32: Love the name!

JelleMillenaar:
Overall question for the Reviewers (@nothingismagick @tensor-programming) and @l1h3r. Is the current implementation fit for bindings? We want to communicate with the Stronghold actor in such a way that Stronghold doesn't have to have bindings in that language. Is the current design fitting for that? Would the current architecture work with IOTA Identity WASM for example?

tensor-programming (commented):
Looks solid.  I would recommend maybe combining the imports into blocks to clean up the size a bit but that's about all.

tensor-programming (approved):

PR #149, opened 2021-03-02
Ensure consistent timestamp precision

l1h3r:
The code below was failing because `Timestamp`s had sub-second precision when created with `Timestamp::now` but changed to whole seconds when serialized as JSON. This was partly a regression from https://github.com/iotaledger/identity.rs/pull/122 - there was also a case where timestamps could be parsed directly with sub-second precision.

```rust
/// Deserializes a Document from a JSON string.
pub fn from_json(json: &String) -> Result<Document, CoreError> {
    match Document::from_json(&json) {
        Ok(document) => Ok(document),
        Err(e) => Err(e),
    }
}

/// Serializes an Document into a JSON string.
pub fn to_json(document: &Document) -> Result<String, CoreError> {
    match document.to_json() {
        Ok(document) => Ok(document),
        Err(e) => Err(e),
    }
}

/// Creates a new Document
pub fn create_did_document() -> Result<Document, IotaError> {
    let keypair: KeyPair = KeyPair::new_ed25519()?;
    let mut document: Document = Document::from_keypair(&keypair)?;
    match document.sign(keypair.secret()) {
        Ok(()) => Ok(document),
        Err(e) => Err(e),
    }
}

#[test]
fn test_functions() {
    let doc1 = create_did_document().unwrap();
    let doc2 = from_json(&to_json(&doc1).unwrap()).unwrap();
    assert_eq!(doc1, doc2);
}
```

### DIFF
--- a/identity-core/src/common/timestamp.rs
+++ b/identity-core/src/common/timestamp.rs
@@ -3,6 +3,7 @@
 
 use chrono::DateTime;
 use chrono::SecondsFormat;
+use chrono::Timelike;
 use chrono::Utc;
 use core::convert::TryFrom;
 use core::fmt::Debug;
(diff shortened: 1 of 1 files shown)


JelleMillenaar:
We are dropping ms precision for this, right? I don't think we need it, but it's good to be aware of. As per the DID standard: 

> The value of the property MUST be a string formatted as an XML Datetime normalized to UTC 00:00:00 and without sub-second decimal precision. For example: 2020-12-20T19:17:47Z.

So this is actually more aligned with the standard, right?

JelleMillenaar (approved):

Issue #143, opened 2021-03-01
Add support for using Blake2b256 with Merkle Key Collections


l1h3r:
Add support for using `blake2b` with a 256-bit output as a Merkle Key Collection digest algorithm

Depends on [moving to crypto.rs](https://github.com/iotaledger/identity.rs/issues/132)

Issue #132, opened 2021-02-22
Move Cryptographic logic to Crypto.rs


JelleMillenaar:
Migrate all Cryptographic primitives and complex primitive uses to the [Crypto.rs](https://github.com/iotaledger/crypto.rs) repository.

- [x] Add HMAC-SHA384 to Crypto.rs [PR](https://github.com/iotaledger/crypto.rs/pull/33)
- [x] Add SHA384 to Crypto.rs [PR](https://github.com/iotaledger/crypto.rs/pull/34)
- [ ] Add AES-128-GCM and AES-192-GCM to Crypto.rs [PR](https://github.com/iotaledger/crypto.rs/pull/35) [Issue](https://github.com/iotaledger/crypto.rs/issues/43)
- [x] Add AES Key Wrap to Crypto.rs [PR](https://github.com/iotaledger/crypto.rs/pull/40)
- [ ] Add AES-CBC to Crypto.rs [PR](https://github.com/iotaledger/crypto.rs/pull/41) [Issue](https://github.com/iotaledger/crypto.rs/issues/43)
- [x] Add pbkdf hmac sha256/sha384 to Crypto.rs [PR](https://github.com/iotaledger/crypto.rs/pull/44)
- [ ] Update Identity.rs to use Crypto.rs

PhilippGackstatter:
Issue is done as all the required additions to crypto.rs were merged. The unchecked PRs are no longer relevant for us due to a changed scope.

PR #152, opened 2021-03-04
Use crypto.rs

l1h3r:
Part of https://github.com/iotaledger/identity.rs/issues/132 - (Update Identity.rs to use Crypto.rs)

Also closes https://github.com/iotaledger/identity.rs/issues/143 - (Blake2b256 with Merkle Key Collections)

### DIFF
--- a/bindings/wasm/Cargo.toml
+++ b/bindings/wasm/Cargo.toml
@@ -31,7 +31,7 @@ features = [
 wasm-bindgen-test = { version = "0.3" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-rand = { version = "0.7", default-features = false, features = ["wasm-bindgen"] }
+getrandom = { version = "0.2", features = ["js"] }
 
 [package.metadata.wasm-pack.profile.release]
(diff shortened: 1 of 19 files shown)


JelleMillenaar (changes requested):
[Crypto.rs also has a random function](https://github.com/iotaledger/crypto.rs/blob/c3bf565eba62d0b81144174c2ff917bfde282e49/Cargo.toml#L52) that we can use instead of having our own dependencies on rand.

l1h3r:
> [Crypto.rs also has a random function](https://github.com/iotaledger/crypto.rs/blob/c3bf565eba62d0b81144174c2ff917bfde282e49/Cargo.toml#L52) that we can use instead of having our own dependencies on rand.

We only depend on `rand` for examples and tests. `rand` has a fairly large high-level API and crypto.rs doesn't include the traits that integrate with `core/std`. Specifically crypto.rs only uses the low-level `getrandom` dependency and follows the same pattern of using `rand` in `dev-dependencies`

JelleMillenaar (approved):

PR #155, opened 2021-03-08
Chrysalis Phase 2

l1h3r:
https://github.com/iotaledger/identity.rs/issues/131

PR #161, opened 2021-03-15
Chrysalis Phase 2 completion

DeppLearning:
Successor of https://github.com/iotaledger/identity.rs/pull/155
Includes https://github.com/iotaledger/identity.rs/pull/160


examples/Cargo.toml
```diff
@@ -7,8 +7,11 @@ publish = false
 [dependencies]
 identity = { path = "../identity" }
 rand = { version = "0.8" }
-smol = { version = "0.1", features = ["tokio02"] }
-smol-potat = { version = "0.3" }
+iota-core = { git = "https://github.com/iotaledger/iota.rs", rev = "86f5bb24727266e13abd772616d0700abe866892" }
+#iota-core = { git = "https://github.com/iotaledger/iota.rs", rev = "c993159e2a0d4eb250dc82e43e233219144e5931" }
```
l1h3r, line 11: should remove all these unused deps

examples/create_did.rs
```diff
@@ -3,14 +3,17 @@
...
 use identity::iota::Result;
 use identity::iota::TangleRef;
 
-#[smol_potat::main]
+#[tokio::main]
 async fn main() -> Result<()> {
-  let client: Client = Client::new()?;
+  // Create a new client connected to the Testnet (Chrysalis).
+  // Node-syncing has to be disabled for now.
+  let client: Client = ClientBuilder::new().node_sync_disabled().build().await?;
```
l1h3r, line 16: There should be a `Client::builder` function. very minor but can avoid an extra import

identity-iota/Cargo.toml
```diff
@@ -12,12 +12,15 @@ homepage = "https://www.iota.org"
...
 [dependencies]
 async-trait = { version = "0.1", default-features = false }
+futures = { version = "0.3" }
 identity-core = { version = "=0.2.0", path = "../identity-core" }
 identity-credential = { version = "=0.2.0", path = "../identity-credential" }
 identity-did = { version = "=0.2.0", path = "../identity-did" }
 iota-constants = { version = "0.2", default-features = false }
 iota-conversion = { version = "0.5", default-features = false }
-iota-core = { git = "https://github.com/Thoralf-M/iota.rs", rev = "d7c8c64fc3ac2340f0148708a916c245f42fd454" }
+#iota-core = { git = "https://github.com/Thoralf-M/iota.rs", rev = "d7c8c64fc3ac2340f0148708a916c245f42fd454" }
```
l1h3r, line 21: unused deps

identity-iota/src/tangle/message_index.rs
```diff
@@ -143,55 +142,55 @@ mod tests {
   }
 
   #[rustfmt::skip]
```
l1h3r, line 144: these `#[rustfmt::skip]` attributes should have been removed a long time ago. probably best to do now

identity-iota/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub mod did;
 pub mod error;
 pub mod tangle;
 
-pub(crate) mod utils;
+// pub(crate) mod utils;
```
l1h3r, line 34: unused?

identity-iota/src/client/client_builder.rs
```diff
@@ -41,9 +43,14 @@ impl ClientBuilder {
     self
   }
 
+  pub fn node_sync_disabled(mut self) -> Self {
```
l1h3r, line 46: `pub fn node_sync_enabled(mut self, value: bool) -> Self` leaves much more flexibility for the caller

identity-iota/src/client/client_builder.rs
```diff
@@ -1,23 +1,25 @@
...
 
 impl ClientBuilder {
   /// Creates a new `ClientBuilder`.
-  pub const fn new() -> Self {
+  pub fn new() -> Self {
     Self {
-      network: Network::Mainnet,
+      network: Network::default(),
       nodes: Vec::new(),
+      node_sync_enabled: true,
```
l1h3r, line 22: can this default to `false` since it seems to be required at the moment?

probably also good to use a constant

identity-iota/src/client/client_builder.rs
```diff
@@ -1,23 +1,25 @@
...
-use crate::client::Client;
-use crate::client::Network;
-use crate::error::Result;
+use crate::{
+  client::{Client, Network},
+  error::Result,
+};
 
 /// A `ClientBuilder` is used to generated a customized `Client`.
-#[derive(Clone, Debug)]
```
l1h3r, line 9: why this change? `bool` should be debuggable

identity-iota/src/client/client.rs
```diff
@@ -1,44 +1,38 @@
...
-use crate::error::Result;
-use crate::tangle::Message;
-use crate::tangle::MessageId;
-use crate::utils::bundles_from_trytes;
-use crate::utils::create_address_from_trits;
-use crate::utils::encode_trits;
-use crate::utils::txn_hash_trytes;
+use futures::stream::FuturesUnordered;
+use futures::stream::StreamExt;
+use identity_core::{common::Url, convert::ToJson};
```
l1h3r, line 6: should consistently use the same style of import.

some form of consistency will be enforced through `rustfmt.toml` eventually

l1h3r:
Looks like CI is failing because the wasm bindings don't compile after these changes - just need to handle the change to `async` here:

https://github.com/iotaledger/identity.rs/blob/dev/bindings/wasm/src/iota.rs#L45

identity-iota/src/client/client.rs
l1h3r: ahh I see this was the style from my old PR 😄 should move it to the new style

identity-iota/src/client/client_builder.rs
```diff
@@ -1,23 +1,25 @@
...
 pub struct ClientBuilder {
   pub(crate) network: Network,
   pub(crate) nodes: Vec<String>,
+  pub(crate) node_sync_enabled: bool,
 }
 
 impl ClientBuilder {
   /// Creates a new `ClientBuilder`.
-  pub const fn new() -> Self {
```
l1h3r, line 17: any reason `const` was dropped?

identity-iota/src/client/network.rs
```diff
@@ -2,15 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use identity_core::common::Url;
-use iota::client::builder;
 
 use crate::did::DID;
 
 lazy_static! {
-  static ref EXPLORER_MAIN: Url = Url::parse("https://explorer.iota.org/mainnet").unwrap();
+  static ref EXPLORER_MAIN: Url = Url::parse("https://explorer.iota.org/chrysalis").unwrap();
```
l1h3r, line 9: We are dropping the `Comnet` variant I thought?

I don't really see a reason to keep the old networks. Definitely think it's a bit confusing to mix them both.

I think just `enum Network { Main, Test }` will do?

cc @JelleMillenaar

identity-iota/src/client/client_builder.rs
DeppLearning: Removed this to be able to call Network::default. Reverted to Network::main now and added that const again.

identity-iota/src/client/client_builder.rs
DeppLearning: Opps, this is a left over change from when we briefly went with something like:
```
pub struct ClientBuilder {
  pub(crate) builder: iota::ClientBuilder,
  pub(crate) network: Network,
}
``` 
before noticing that it's probably better not to wrap it in the struct. Reverted this

l1h3r:
Currently blocked by lack of wasm support in iota.rs

https://github.com/iotaledger/iota.rs/issues/444

l1h3r (approved):

Issue #134, opened 2021-02-22
Add conceptual documentation


JelleMillenaar:
Add conceptual documentation for the DID, VC, and IOTA Identity topics on the documentation website. This should include in-depth descriptions of IOTA Identity's unique features. 

- [X] DID and IOTA Method spec
- [X] Verifiable Credentials
- [X] DID Communications
- [X] Account Module / Actor
- [X] Detailed roadmap 
- [X] Future plans: Beyond the framework
- [X] Glossary
- [X] Contribute

PR #162, opened 2021-03-16
Updated Mdbook structure

JelleMillenaar:
# Description of change

Updated the MDbook structure to reflect a basis on how the conceptual documentation should look like.

The basis for #134 

## Type of change
Remove options that are not relevant to your changes.

- Documentation Fix


Aconitin (approved):
Looks good, but maybe we should talk again about the extra "Programming Languages" section, might actually be better to include that in the other chapters after all.

PR #166, opened 2021-03-16
Rust week cleanup

l1h3r:
Addressing some suggestions from "Rust week"

- Removed `AsJson` trait (just use `ToJson`/`FromJson`)
- Renamed ed25519 helpers
- Misc. doc fixes


identity-credential/src/credential/builder.rs
```diff
@@ -234,17 +233,29 @@ mod tests {
 
     assert_eq!(credential.context.len(), 2);
     assert_eq!(credential.context.get(0).unwrap(), Credential::<Object>::base_context());
-    assert_eq!(credential.context.get(1).unwrap(), "https://www.w3.org/2018/credentials/examples/v1");
+    assert_eq!(
```
JelleMillenaar, line 236: I am curious why these specifically are spread over multiple lines, but others aren't. Is it based on the length?

JelleMillenaar (approved):
Nothing huge I see.

identity-credential/src/credential/builder.rs
l1h3r: this is just due to `rustfmt`, it was previously disabled so this code wasn't being formatted

PR #170, opened 2021-03-18
Combine verifiable types with base types

### LINKED ISSUE #167: Combine Credential and VerifiableCredential types
## Description

Combine the `Credential` and `VerifiableCredential` types. We only need to have `VerifiableCredential`.

The same should be done for presentations

## Motivation

The current approach is tedious to work with and doesn't provide any security benefits.

You currently have to wrap a `Credential` with a `VerifiableCredential` to make it signable...

```rust
let mut target = VerifiableCredential::new(credential, Vec::new());
document.sign_data(&mut credential, secret)?;
```

...instead, you should be able to sign a `Credential` directly:

```rust
document.sign_data(&mut credential, secret)?;
```

## To-do list

- [ ] Merge `VerifiableCredential` -> `Credential`
  - [ ] `proof`/`proof_mut` methods
  - [ ] `identity_core::crypto` trait implementations
  - [ ] Change `proof` type: `OneOrMany<Signature>` -> `Option<Signature>`

- [ ] Merge `VerifiablePresentation` -> `Presentation`
  - [ ] `proof`/`proof_mut` methods
  - [ ] `identity_core::crypto` trait implementations
  - [ ] Change `proof` type: `OneOrMany<Signature>` -> `Option<Signature>`

## Change checklist

DeppLearning:
# Description of change
- Merged VerfiablePresentations into Presentations
- Merged VerifiableCredentials into Credentials
- Adjusted test-fixtures and rust-docs

- fixes #167 

## Type of change

- [ ] Documentation Fix

## Change checklist

<del> - [ ] I have added tests that prove my fix is effective or that my feature works </del>


examples/credential.rs
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! A basic example that generates and publishes subject and issuer DID
-//! Documents, creates a VerifiableCredential specifying claims about the
+//! Documents, creates a Credential specifying claims about the
```
l1h3r, line 5:
```suggestion
//! Documents, creates a Verifiable Credential specifying claims about the
```

I think it's better to use the W3C terminology here

examples/credential.rs
```diff
@@ -53,13 +54,13 @@ async fn main() -> Result<()> {
   // Create an unsigned Credential with claims about `subject` specified by `issuer`.
   let credential: Credential = issue_degree(&doc_iss, &doc_sub)?;
 
-  // Sign the Credential with the issuer secret key - the result is a VerifiableCredential.
-  let credential: VerifiableCredential = credential.sign(&doc_iss, "#authentication".into(), key_iss.secret())?;
+  // Sign the Credential with the issuer secret key - the result is a Credential.
+  let credential: Credential = credential.sign(&doc_iss, "#authentication".into(), key_iss.secret())?;
```
l1h3r, line 58: This is also possible now:

```rust
doc_iss.sign_data(&mut credential, key_iss.secret())?;
```

I think it makes sense to drop the `sign` function from the actual `Credential` type now. All types that implement `Serialize` and `SetSignature` can be signed through the `sign_data` function (this implicitly uses the authentication method)

identity-credential/src/credential/credential.rs
```diff
@@ -34,7 +37,7 @@ lazy_static! {
 
 /// A `Credential` represents a set of claims describing an entity.
 ///
-/// `Credential`s can be signed with `Document`s to create `VerifiableCredential`s.
+/// `Credential`s can be signed with `Document`s to create `Credential`s.
```
l1h3r, line 40: This line can probably be removed now

identity-credential/src/credential/credential.rs
```diff
@@ -79,6 +82,9 @@ pub struct Credential<T = Object> {
   /// Miscellaneous properties.
   #[serde(flatten)]
   pub properties: T,
+  /// Proof(s) used to verify a `Credential`
+  #[serde(skip_serializing_if = "Option::is_none")]
+  pub proof: Option<Signature>,
```
l1h3r, line 87: Does this actually get populated when deserialized or is the `proof` field "merged" into `properties` because of `#[serde(flatten)]`?

identity-credential/src/credential/credential.rs
```diff
@@ -150,25 +157,33 @@ impl<T> Credential<T> {
...
 }
 
 impl<T> Credential<T>
 where
   T: Serialize,
 {
-  /// Creates a new [`VerifiableCredential`] by signing `self` with
+  /// Creates a new [`Credential`] by signing `self` with
   /// [`document`][`Document`] and [`secret`][`SecretKey`].
   pub fn sign<D1, D2, D3>(
```
l1h3r, line 178: See comment above about just removing this completely

identity-credential/src/presentation/presentation.rs
```diff
@@ -10,19 +10,22 @@ use identity_core::common::Object;
...
-use crate::credential::VerifiableCredential;
 use crate::error::Error;
 use crate::error::Result;
 use crate::presentation::PresentationBuilder;
 
-/// A `Presentation` represents a bundle of one or more `VerifiableCredential`s.
+/// A `Presentation` represents a bundle of one or more `Credential`s.
 ///
-/// `Presentation`s can be signed with `Document`s to create `VerifiablePresentation`s.
+/// `Presentation`s can be signed with `Document`s to create verifiable `Presentation`s.
```
l1h3r, line 28: This line can probably be removed now

identity-credential/src/presentation/presentation.rs
```diff
@@ -49,6 +52,9 @@ pub struct Presentation<T = Object, U = Object> {
   /// Miscellaneous properties.
   #[serde(flatten)]
   pub properties: T,
+  /// Proof(s) used to verify a `Presentation`
+  #[serde(skip_serializing_if = "Option::is_none")]
+  pub proof: Option<Signature>,
```
l1h3r, line 57: See comment above regarding `flatten`. I think ordering is important here (for the generated code)

l1h3r (changes requested):

identity-credential/src/presentation/presentation.rs
```diff
@@ -10,19 +10,22 @@ use identity_core::common::Object;
...
 use crate::credential::Credential;
 use crate::credential::Policy;
 use crate::credential::Refresh;
-use crate::credential::VerifiableCredential;
 use crate::error::Error;
 use crate::error::Result;
 use crate::presentation::PresentationBuilder;
 
-/// A `Presentation` represents a bundle of one or more `VerifiableCredential`s.
+/// A `Presentation` represents a bundle of one or more `Credential`s.
```
l1h3r, line 26: This was an existing issue but this crate is never touched so might as well do now 😄 . The documentation can link directly to other items by using ```[`SomeStruct`]``` but this was incorrectly using ``` `SomeStruct` ```.

The [sign function](https://github.com/iotaledger/identity.rs/pull/170/files#diff-f866bde21b7bc14af5faca62954b5990f570cf2dafb67ae1853b3025c316bf81R176) has an example of this

DeppLearning: Ah thought this was another convention since I thought I only saw it in places were it was self-referential, where it makes sense to not link..

I will adjust this where I can find this pattern

identity-credential/src/credential/credential.rs
DeppLearning: I thought the same when diagnosing why the tests that deserialize Credentials and Presentations from Json failed. However, those run through, i.e. here: 
https://github.com/iotaledger/identity.rs/blob/58608f86f2c9c494784ce2f856bcb343fa9880e2/identity-credential/src/presentation/presentation.rs#L173

When I noticed there seems to be no difference, in terms of flatten, I put it in the last position since this would have been the original layout after the inner type was flattened in the Verifiable* version which also matches the specs.

The docs on flatten also mention that you can use flatten multiple times, if our initial suspicion was correct, this would never be possible?

l1h3r: looking at the generated code it doesn't seem to be an issue:

```rust
let __field14 = match __field14 {
    _serde::__private::Some(__field14) => __field14,
    _serde::__private::None => {
        match _serde::__private::de::missing_field("proof") {
            _serde::__private::Ok(__val) => __val,
            _serde::__private::Err(__err) => {
                return _serde::__private::Err(__err);
            }
        }
    }
};
let __field13: T = match _serde::de::Deserialize::deserialize(
    _serde::__private::de::FlatMapDeserializer(
        &mut __collect,
        _serde::__private::PhantomData,
    ),
) {
    _serde::__private::Ok(__val) => __val,
    _serde::__private::Err(__err) => {
        return _serde::__private::Err(__err);
    }
};
```

but we should definitely have more tests 👍

identity-credential/src/presentation/presentation.rs
l1h3r: Not a convention but just written before docs became a bigger focus - the self-referential ones are **also** not needed so those could be updated as you encounter them.

examples/credential.rs
l1h3r: Should be able to use the `DocumentSigner`:

```rust
document.signer(keypair.secret()).method("#key-1").sign(&mut credential)?;
```

l1h3r (approved):
nice job on the documentation cleanup 🤩

PR #174, opened 2021-03-19
Include README.md as crate documentation

l1h3r:
A quick test using the nightly `extended_key_value_attributes` feature to cut down on documentation duplication.

Files can be included with a simple attribute: `#![doc = include_str!("relative/path/to/file.md")]`.

We already build documentation with nightly rust so this pattern doesn't really add any overhead.

You need use multiple layers of `cfg_attr` at the moment (compiler bug when `doc = include_str!` is conditionally included)

```rust
#![cfg_attr(docsrs, cfg_attr(docsrs, doc = include_str!("../README.md")))]
```

```rust
pub mod example {
  #[cfg_attr(docsrs, cfg_attr(docsrs, doc = include_str!("../example.md")))]
  pub struct Example;
}
```

### DIFF
--- a/identity-core/Cargo.toml
+++ b/identity-core/Cargo.toml
@@ -34,3 +34,9 @@ features = ["blake2b", "ed25519", "random", "sha"]
 
 [dev-dependencies]
 rand = { version = "0.8" }
+
+[package.metadata.docs.rs]
+# To build locally:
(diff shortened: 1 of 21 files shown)


JelleMillenaar (approved):
Awesome improvement for how we handle documentation!

Issue #141, opened 2021-03-01
Replace `immutable` property with DID URL query param


l1h3r:
The `immutable` property currently determines whether we fetch updates from the diff-chain of a particular auth-chain record. This property should be removed from the DID Document type as the same optimization can be accomplished with query params, eg. `did:iota:1234?diff=false`.

- [ ] Remove `immutable` property from `Document` type
- [ ] Add support for `diff=true|false` DID query param in `Client::read_document_chain`
  - [ ] Add the query param to the [DID Method Specification](https://github.com/iotaledger/identity.rs/issues/130)

PR #180, opened 2021-03-23
Remove immutable property and support for query param

marcianos:
# Description of change
Removed the immutable property from the DID Document type and related methods.
Add support for query parameter diff=true|false.

fixes issue #141 

## Type of change
- [X] Documentation Fix

Customization and addition of new tests in diff_chain example.
Test with no query parameter, test with query parameter diff=true and diff=false.

## Change checklist

- [X] I have added tests that prove my fix is effective or that my feature works


identity-iota/src/client/client.rs
```diff
@@ -160,16 +160,22 @@ impl Client {
 
     let auth: AuthChain = AuthChain::try_from_messages(did, &messages)?;
 
-    let diff: DiffChain = if auth.current().immutable() {
-      DiffChain::new()
-    } else {
+    // Check if there is any query given and return
+    let diff_option = match did.query() {
```
l1h3r, line 164: You can use [`query_pairs`](https://docs.rs/did_url/0.1.0/did_url/struct.DID.html#method.query_pairs) for this - it's pretty explicit but it handles encoding automatically

```rust
let skip_diff: bool = did.query_pairs().any(|(key, _)| key == "diff" && value == "false").unwrap_or_default();
```

examples/diff_chain.rs
```diff
@@ -179,5 +179,54 @@ async fn main() -> Result<()> {
...
+  println!("Chain (R) {:#}", remote);
+  println!();
+
+  let a: &Document = chain.current();
+  let b: &Document = remote.current();
+
+  // The current document in the resolved chain should not be identical to
+  // the current document in our local chain because you read an outdated
+  // version of the document
+  assert_ne!(a, b);
```
l1h3r, line 229: nice! always forget about `assert_ne!`

identity-iota/src/client/client.rs
marcianos: Ah thank you! I didn't see query_pairs. But is has to be:
`let skip_diff: bool = did.query_pairs().any(|(key, value)| key == "diff" && value == "false");`
Is that right?

l1h3r: Yeah, that looks right 👍

JelleMillenaar (approved):
Approved based on preview review from @l1h3r's feedback being implemented.

Issue #172, opened 2021-03-18
Improve identity::iota::Document type consistency


l1h3r:
## Description

The `Deref` implementation for `identity::iota::Document` leads to situations where unexpected "core" types are returned when IOTA-specific types are expected.

This trait can be removed and the relevant methods from `iota::did::Document` can be re-implement in `identity::iota::Document` while also performing the necessary type conversions.

## Motivation

This can be confusing and often leads to type inconsistencies

## Resources

https://rust-unofficial.github.io/patterns/anti_patterns/deref.html

> There is no one good alternative

## To-do list

- [x] Remove `Deref` from `identity::iota::Document`
- [x] Validate all DID Document verification methods in `iota::Document::try_from_core`**
  ```rust
  for method in document.methods() { Method::check_validity(method)?; }
  ```

- [x] Re-implement the following methods for `identity::iota::Document`***:
  - [ ] controller
  - [ ] methods
  - [ ] resolve
  - [ ] try_resolve

** This is required for `unsafe` casts between references (safe alternatives require a clone and refs are needed for iterators, etc.)

*** This should just call the "core" document method and convert the output type (see `iota::Document::id(...)`)

## Change checklist

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

m-renaud:
I can pick this one up.

m-renaud:
The following also need to be re-implemented:

- `proof()`
- `verifier()`
- `signer()`
- `service()`

Additionally, removing the `Deref` impl didn't actually cause any failures due to missing `controller()`, `methods()`, `resolve()`, or `try_resolve()` calls. So, I think I'll start with a PR to expand test coverage, and then send a follow-up to remove the `Deref` impl.

m-renaud:
Hmm, ran into another issue: `impl From<BaseDocument> for IotaDocument`

This appears to be "unsafe" since there is no validation that the underlying `BaseDocument` (ie. `CoreDocument<Properties, Object, Object>` actually coforms to the IOTA DID spec.

[...]

PR #233, opened 2021-05-02
Return IOTA specific types for methods on IotaDocument

m-renaud:
# Description of change

## Summary

Remove the `impl Deref for IotaDocument` and re-implement `CoreDocument` methods on `IotaDocument` returning the correct (ie. IOTA-specific) types.

## Details

Specifically, the following functions have been linked to the `IotaDocument` type:

- `IotaDocument::controller()` - "unsafe", see Safety section below
- `IotaDocument::also_known_as()` - "safe"
- `IotaDocument::proof()` - "safe"
- `IotaDocument::methods()` - "unsafe", see Safety section below
- `IotaDocument::resolve()` - "unsafe", see Safety section below
- `IotaDocument::try_resolve()` - "unsafe, see Safety section below
- `IotaDocument::verifier()` - "safe"
- `IotaDocument::signer()` - "safe"
- `IotaDocument::services()` - "safe"

## Safety / Verification

The `IotaDocument::try_from_core()` method has been updated to include validity checks for:
- `document.controller()` - needed for `IotaDocument::controller()`
- `document.methods()` - needed for `IotaDocument::methods()`, `IotaDocument::resolve()`, `IotaDocument::try_resolve()`

Reviewers, please triple check that there aren't other ways non-conformant documents can slip in.

As a side note, I noticed that the `impl From<BaseDocument> for IotaDocument` instance may not be "safe", since `BaseDocument` is simply `CoreDocument<Properties, Object, Object>` which I think someone could construct themself (and not conform to the spec), and then convert this to an `IotaDocument`.

## Breaking Change

This is a breaking change, any code that was previously relying on the `Deref` instance for `IotaDocument`->`CoreDocument` will need to update. Hopefully the breakages are minimal since the most common functions that are used via `Deref` have been re-implemented on `IotaDocument`.

fixes issue #172

## Type of change

`cargo test`

## Change checklist

### DIFF
--- a/bindings/wasm/src/wasm_document.rs
+++ b/bindings/wasm/src/wasm_document.rs
@@ -11,8 +11,8 @@ use identity::crypto::PublicKey;
 use identity::crypto::SecretKey;
 use identity::did::verifiable;
 use identity::did::MethodScope;
-use identity::did::VerificationMethod;
 use identity::iota::DocumentDiff;
+use identity::iota::Error;
 use identity::iota::IotaDID;
 use identity::iota::IotaDocument;
 use identity::iota::IotaVerificationMethod;
@@ -220,7 +220,7 @@ impl WasmDocument {
         let merkle_key: Vec<u8> = self
           .0
           .try_resolve(&*method)
-          .and_then(|method| method.key_data().try_decode())
+          .and_then(|method| method.key_data().try_decode().map_err(Error::InvalidDoc))
           .map_err(err)?;
 
         let public: PublicKey = decode_b58(&public).map_err(err).map(Into::into)?;
@@ -260,11 +260,7 @@ impl WasmDocument {
 
   #[wasm_bindgen(js_name = resolveKey)]
   pub fn resolve_key(&mut self, query: &str) -> Result<WasmVerificationMethod, JsValue> {
-    let method: VerificationMethod = self.0.try_resolve(query).map_err(err)?.clone();
-
-    IotaVerificationMethod::try_from_core(method)
-      .map_err(err)
-      .map(WasmVerificationMethod)
+    Ok(WasmVerificationMethod(self.0.try_resolve(query).map_err(err)?.clone()))
   }
 
   #[wasm_bindgen(js_name = revokeMerkleKey)]
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use core::convert::TryFrom;
+use core::convert::TryInto;
 use core::fmt::Debug;
 use core::fmt::Display;
 use core::fmt::Formatter;
@@ -101,7 +102,8 @@ impl IotaDocument {
       .authentication(method)
       .build()
       .map(CoreDocument::into_verifiable)
-      .map(Into::into)
+      .map(TryInto::try_into)
+      .map(Result::unwrap)
       .unwrap() // `unwrap` is fine - we provided all the necessary properties
   }
 
@@ -111,6 +113,29 @@ impl IotaDocument {
   ///
   /// Returns `Err` if the document is not a valid IOTA DID Document.
   pub fn try_from_core(document: CoreDocument) -> Result<Self> {
+    IotaDocument::validate_core_document(&document)?;
+
+    Ok(Self {
+      document: document.serde_into()?,
+      message_id: MessageId::null(),
+    })
+  }
+
+  /// Converts a generic DID [`Document`][`BaseDocument`] to an IOTA DID Document.
+  ///
+  /// # Errors
+  ///
+  /// Returns `Err` if the document is not a valid IOTA DID Document.
+  pub fn try_from_base(document: BaseDocument) -> Result<Self> {
+    IotaDocument::validate_core_document(&document)?;
+
+    Ok(Self {
+      document: document.serde_into()?,
+      message_id: MessageId::null(),
+    })
+  }
+
+  fn validate_core_document<T, U, V>(document: &CoreDocument<T, U, V>) -> Result<()> {
     // Validate that the DID conforms to the IotaDID specification.
     // This check is required to ensure the correctness of the `IotaDocument::id()` method which
     // creates an `IotaDID::new_unchecked_ref()` from the underlying DID.
@@ -130,7 +155,7 @@ impl IotaDocument {
       IotaVerificationMethod::check_validity(method)?;
     }
 
-    let method: &VerificationMethod = document
+    let method = document
       .authentication()
       .head()
       .and_then(|method| document.resolve_ref(method))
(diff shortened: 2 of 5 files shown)


bindings/wasm/src/wasm_document.rs
```diff
@@ -236,11 +236,7 @@ impl WasmDocument {
 
   #[wasm_bindgen(js_name = resolveKey)]
   pub fn resolve_key(&mut self, query: &str) -> Result<WasmVerificationMethod, JsValue> {
```
m-renaud, line 238: @l1h3r previously this code relied on `IotaVerificationMethod::try_from_core()` (because `IotaDocument.resolve()` returns a `VerificationMethod`), should we also change the signature of this to just return `WasmVerificationMethod`? It should be impossible for this to fail now (since `WasmDocument` wraps an already verified `IotaDocument`. lmk what you think.

Can also change the signature in a follow-up PR since that may be more involved.

l1h3r: The conversion doesn't fail but the query can still fail...but it probably makes more sense to just use `resolve` and return `null` in JavaScript.

This is probably best to do in a separate PR; may try to remove some other `Result`s that don't offer much value in JavaScript.

m-renaud: SG, will address in a follow-up.

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -173,6 +218,16 @@ impl IotaDocument {
     unsafe { IotaDID::new_unchecked_ref(self.document.id()) }
   }
 
+  /// Returns a reference to the `IotaDocument` controller.
+  pub fn controller(&self) -> Option<&IotaDID> {
+    unsafe { self.document.controller().map(|d| IotaDID::new_unchecked_ref(d)) }
```
l1h3r, line 223: This should have a `SAFETY` comment to explain how this is considered safe. doesn't really need to be as verbose as the others and could just reference `check_authentication`

*edit:* noticed there were quite a few so definitely feel free to keep it short 😅

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -99,19 +102,65 @@ impl IotaDocument {
...
+    // This check is required to ensure the correctness of the `IotaDocument::controller()` method which
+    // creates an `IotaDID::new_unchecked_ref()` from the underlying controller.
+    document.controller().map_or(Ok(()), |c| IotaDID::check_validity(c))?;
+
+    // Validate that the validation methods conform to the IotaDID specification.
+    // This check is required to ensure the correctness of the `IotaDocument::methods()`,
+    // `IotaDocument::resolve()`, `IotaDocument::try_resolve()`, IotaDocument::resolve_mut()`,
+    // and IotaDocument::try_resolve_mut()` methods which creates an `IotaDID::new_unchecked_ref()`
+    // from the underlying controller.
+    for method in document.methods() {
```
l1h3r, line 159: This only checks the embedded verification methods and not ones referenced by `DID`.

Currently the only way to **add** a non-IOTA DID verification method is by converting to/from JSON but I think a simple fix would be to add `Document::method_refs` that returns the unfiltered `MethodRef` type.

You could then `match` in the iterator and either call `IotaVerificationMethod::check_validity` or `IotaDID::check_validity`

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -230,10 +285,25 @@ impl IotaDocument {
...
+    self.document.proof()
+  }
+
   // ===========================================================================
   // Verification Methods
   // ===========================================================================
 
+  /// Returns an iterator over all verification methods in the DID Document.
+  pub fn methods(&self) -> impl Iterator<Item = &IotaVerificationMethod> {
+    unsafe {
```
l1h3r, line 299: See `SAFETY` comment above

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -250,6 +320,39 @@ impl IotaDocument {
     Ok(())
   }
 
+  /// Returns the first verification [`method`][`IotaverificationMethod`] with an `id` property
+  /// matching the provided `query`.
+  pub fn resolve<'query, Q>(&self, query: Q) -> Option<&IotaVerificationMethod>
+  where
+    Q: Into<MethodQuery<'query>>,
+  {
+    unsafe {
```
l1h3r, line 329: See `SAFETY` comment above

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -250,6 +320,39 @@ impl IotaDocument {
...
+  /// matching the provided `query`.
+  ///
+  /// # Errors
+  ///
+  /// Fails if no matching verification `IotaVerificationMethod` is found.
+  pub fn try_resolve<'query, Q>(&self, query: Q) -> Result<&IotaVerificationMethod>
+  where
+    Q: Into<MethodQuery<'query>>,
+  {
+    unsafe {
```
l1h3r, line 347: See `SAFETY` comment above

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -404,6 +523,8 @@ impl IotaDocument {
   }
 }
 
+impl<'a, 'b, 'c> IotaDocument {}
```
l1h3r, line 526: 🤔

l1h3r (changes requested):
This is awesome 🤩 left a few comments but looks good otherwise

identity-iota/src/did/doc/iota_document.rs
m-renaud: Done.

identity-iota/src/did/doc/iota_document.rs
m-renaud: Done.

identity-iota/src/did/doc/iota_document.rs
m-renaud: Ahh good catch, I glossed over the `__filter_ref` in `CoreDocument::methods()`.

So, the one tricky thing is the `self.verification_method` part of that iterator `.chain()` in `CoreDocument::methods()` because this is a `VerificationMethod<T>`, not a `MethodRef<T>` (like all the others). This means you can't construct a `MethodRef::Embed` from a reference, so (afaik) we'd need to `Clone` the underlying `VerificationMethod` to construct the iterator, which seems non-ideal from a performance standpoint. lmk if I'm missing something though.

identity-iota/src/did/doc/iota_document.rs
m-renaud: Done.

identity-iota/src/did/doc/iota_document.rs
m-renaud: Done

identity-iota/src/did/doc/iota_document.rs
m-renaud: Ummmmm I think this was from something else I was playing around with, plz disregard haha (and removed, of course).

identity-iota/src/did/doc/iota_document.rs
l1h3r: > This means you can't construct a MethodRef::Embed from a reference

😵 I glossed over this as well. you're right that `Clone` would be non-ideal.

Maybe this is better suited as just an iterator over the [verification relationships](https://www.w3.org/TR/2020/WD-did-core-20200731/#verification-relationships) and you could use that in combination with `document.verification_method()`.

```
for method_ref in document.verification_relationships() {
```

m-renaud: Seems resonable, done!

l1h3r (approved):
👍

PR #235, opened 2021-05-04
use iota_client

Thoralf-M:
# Description of change
use iota-client instead of iota-core, because we will remove the latter probably soon


l1h3r (approved):
Looks good 🚀 originally wanted to hold off because I thought it was a version change 😄

Issue #264, opened 2021-05-11
[Bug] (node:91512) UnhandledPromiseRejectionWarning: Chain Error: Invalid Root Document


sdellava:
## Bug description
identity.resolve("did:iota:HVbziVthxfpUMhv2xuo3Dhr1BtQqTaq373EXbi42vGKL") on testnet always return the error  Invalid Root Document

iota node: https://api.lb-0.testnet.chrysalis2.com:443

did.document: https://explorer.iota.org/testnet/message/1c2de96b49cbc293c84f6077b404cd98bf240486f72d9d403d56680e2ac79a16

## Language version

- Language: Node.js
- versione: v14.15.4.

## IOTA Identity version
Which version of IOTA Identity are you using?  Version: 0.3.1

- Version (version number, commit, or branch):  from npm i @iota/identity-wasm

## Steps To reproduce the bug

> npm i @iota/identity-wasm
> node
const identity = require('@iota/identity-wasm/node')
const run = async () => {
const key = await new identity.KeyPair(identity.KeyType.Ed25519)
const doc = await identity.Document.fromKeyPair(key)
doc.sign(key)
message = await identity.publish(doc.toJSON(), { node: "https://api.lb-0.testnet.chrysalis2.com:443" })
console.log(message)
document = await identity.resolve(doc.id.toString())
console.log(document)
}
run()

sdellava:
update: here you can find two dids. The first is resolved, the second return error.

did:iota:2rXkhC7f4i62WLf3t62ngFSTEt9m6q4kHMTTFmK9JtLY   - OK !

did:iota:8gpmPxYWx7GVqY6pdsh4rcJZhZRmgPfm1cTZXnPzVx1P  - NO OK !!!

Looking in the explorer I can't find any significant error in the JSON.

JonasHiltl:
I'm getting the same error while validating a credential with `checkCredential()`. I gues that the `checkCredential` function also resolves the Did document under the hood so the reason for the error is probably the same.
Do you have any news on the issue?

m-renaud:
@sdellava in the bug description you mention that the failure occurs when on the testnet, not the mainnet, but looking at the DIDs that are created, neither of them have a non-`main` network specified (such as `did:iota:test:XYZ`). Is it possible that the node is publishing to the testnet but `resolve()` is attempting to find the message on the main net?

[...]

PR #276, opened 2021-05-21
Add IotaDocument::from_keypair_with_network() and WASM binding.

m-renaud:
# Description of change

Add `IotaDocument::from_keypair_with_network(key, network)` and expose in WASM bindings (via optional `network` param in `fromKeyPair`).

This was a missing API which required a workaround (see https://github.com/iotaledger/identity.rs/issues/264#issuecomment-843364038).

This PR also include an example in the WASM examples README alongside the `identity.Document.fromKeyPair()` call (in comments).

#264 root cause was from a mismatch in network between document and node, and it wasn't obvious how to create a document from a keypair on the test network.

## Type of change

- [x] Documentation Fix

`cargo test`

## Change checklist

### DIFF
--- a/bindings/wasm/src/wasm_document.rs
+++ b/bindings/wasm/src/wasm_document.rs
@@ -86,6 +86,14 @@ impl WasmDocument {
     IotaDocument::from_keypair(&key.0).map_err(err).map(Self)
   }
 
+  /// Creates a new DID Document from the given KeyPair on the specified network.
+  #[wasm_bindgen(js_name = fromKeyPairWithNetwork)]
+  pub fn from_keypair_with_network(key: &KeyPair, network: &str) -> Result<WasmDocument, JsValue> {
+    IotaDocument::from_keypair_with_network(&key.0, &network)
+      .map_err(err)
+      .map(Self)
+  }
+
   /// Creates a new DID Document from the given verification [`method`][`Method`].
   #[wasm_bindgen(js_name = fromAuthentication)]
   pub fn from_authentication(method: &WasmVerificationMethod) -> Result<WasmDocument, JsValue> {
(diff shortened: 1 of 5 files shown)


JelleMillenaar:
TIL Rust has no real optional or default arguments to a function, that's weird TBH. 
Anywho, Typescript does and I think we can cut a function if we upgrade the WASM fromKeypair function to also handle the network.

m-renaud:
> TIL Rust has no real optional or default arguments to a function, that's weird TBH.

Heh, yeah, I don't miss it in most cases personally. Many of the situations where default arguments are used it's either, a) better replaced by a builder, or b) ambiguous what the various omitted arguments are (especially when they're the same type).

> Typescript does and I think we can cut a function if we upgrade the WASM fromKeypair function to also handle the network

SGTM. Using default arguments is more idiomatic typescript so agree we should follow the convention of the bindings language.

m-renaud:
Updated. Although, made the changes from the GitHub editors, :crossed_fingers: no typos/compile errors haha

bindings/wasm/src/wasm_document.rs
```diff
@@ -80,10 +80,16 @@ impl WasmDocument {
...
+  ///
+  /// If unspecified, network defaults to the mainnet.
   #[wasm_bindgen(js_name = fromKeyPair)]
-  pub fn from_keypair(key: &KeyPair) -> Result<WasmDocument, JsValue> {
-    IotaDocument::from_keypair(&key.0).map_err(err).map(Self)
+  pub fn from_keypair(key: &KeyPair, network: Option<String>) -> Result<WasmDocument, JsValue> {
+    let doc = match network {
+      Some(net) => IotaDocument::from_keypair_with_network(&key.0, &net),
+      None => IotaDocument::from_keypair(&key.0)
+    }
```
l1h3r, line 91:
```suggestion
    };
```

l1h3r (approved):

PR #291, opened 2021-07-02
Add features to account to allow wasm compilation

PhilippGackstatter:
# Description of change

It should be possible to compile the upcoming identity actor to WebAssembly. It will use types from the account crate, and the account crate is not Wasm compatible due to a dependency on stronghold. This PR adds features to the account crate to opt-in to stronghold support - which is enabled by default - but can be opted out of to enable compilation to Wasm.

There is a weird issue with `tokio`, where even if we specify it to only use feature `sync` for example, it will still depend on the `mio` crate, even though it shouldn't. This may be happening due to how cargo selects features, and I've described the issue in some detail in the `toml` file. Fortunately, this can be worked around by specifying dependencies via git rather than crates.io. Makes it a bit less ergonomic to work with, i.e. we have to explicitly specify when we want to update `tokio` by updating the revision, but that shouldn't be too much of a hassle.

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/identity-account/Cargo.toml
+++ b/identity-account/Cargo.toml
@@ -17,16 +17,19 @@ hashbrown = { version = "0.9", features = ["serde"] }
 identity-core = { version = "=0.3.0", path = "../identity-core" }
 identity-credential = { version = "=0.3.0", path = "../identity-credential" }
 identity-did = { version = "=0.3.0", path = "../identity-did" }
-identity-iota = { version = "=0.3.0", path = "../identity-iota" }
+identity-iota = { version = "=0.3.0", path = "../identity-iota", default-features = false }
 itoa = { version = "0.4" }
 log = { version = "0.4", default-features = false }
 once_cell = { version = "1.7", default-features = false, features = ["std"] }
 paste = { version = "1.0" }
-riker = { version = "0.4" }
-serde = { version = "1.0", default-features = false, features = ["alloc", "derive"] }
+riker = { version = "0.4", optional = true }
+serde = { version = "1.0", default-features = false, features = [
+    "alloc",
+    "derive"
+] }
 slog = { version = "2.7" }
 thiserror = { version = "1.0" }
-tokio = { version = "1.5", default-features = false, features = ["rt", "rt-multi-thread", "sync"] }
+tokio = { version = "1.5", features = ["sync"] }
 zeroize = { version = "1.3" }
 
 [dependencies.iota-crypto]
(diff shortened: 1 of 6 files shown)


identity-account/Cargo.toml
```diff
@@ -17,16 +17,19 @@ hashbrown = { version = "0.9", features = ["serde"] }
...
+identity-iota = { version = "=0.3.0", path = "../identity-iota", default-features = false }
 itoa = { version = "0.4" }
 log = { version = "0.4", default-features = false }
 once_cell = { version = "1.7", default-features = false, features = ["std"] }
 paste = { version = "1.0" }
-riker = { version = "0.4" }
-serde = { version = "1.0", default-features = false, features = ["alloc", "derive"] }
+riker = { version = "0.4", optional = true }
+serde = { version = "1.0", default-features = false, features = [
+    "alloc",
```
l1h3r, line 27: 💅 strange formatting 😛

identity-account/Cargo.toml
```diff
@@ -17,16 +17,19 @@ hashbrown = { version = "0.9", features = ["serde"] }
...
+    "derive"
+] }
 slog = { version = "2.7" }
 thiserror = { version = "1.0" }
-tokio = { version = "1.5", default-features = false, features = ["rt", "rt-multi-thread", "sync"] }
+tokio = { version = "1.7", features = ["sync"] }
```
l1h3r, line 32: was the change to 1.7 required to fix wasm compilation? maybe better to update as a separate PR.

identity-account/Cargo.toml
```diff
@@ -36,15 +39,30 @@ features = ["blake2b", "ed25519", "hmac", "pbkdf", "sha", "slip10", "std"]
...
 [features]
 mem-client = []
+stronghold = [
+    "iota_stronghold",
+    "stronghold_engine",
+    "riker",
+    "tokio/rt-multi-thread",
+]
+wasm = ["identity-iota/wasm"]
+default = ["stronghold", "identity-iota/async"]
```
l1h3r, line 68: better to have an `async` feature that enables this feature in `identity-iota` (`default = ["stronghold", "async"]`). at the moment it's a bit awkward to disable stronghold and enable the `async` feature

identity-account/Cargo.toml
l1h3r: in any case, the tokio version in the examples should be kept in sync with this one.

PhilippGackstatter: No that wasn't required for Wasm, I just upgraded it. But should be a separate PR, I agree.

JelleMillenaar (approved):

PR #292, opened 2021-07-07
Example Cleanup

l1h3r:
# Description of change

## Type of change

- [x] Documentation Fix

Examples and tests

## Change checklist


identity-iota/src/error.rs
```diff
@@ -51,4 +51,10 @@ pub enum Error {
   CannotRemoveAuthMethod,
   #[error("Cannot Revoke Verification Method")]
   CannotRevokeMethod,
+  /// Caused by attempting to read a poisoned shared resource.
+  #[error("Shared resource poisoned: read")]
+  SharedReadPoisoned,
```
PhilippGackstatter, line 56: I do wonder, would anyone actually handle such errors? I don't know that panicking is the better solution, either, though. Avoiding it would be best. Maybe instead of `RwLock<HashMap<...>>` we could use an actually concurrent HashMap like `dashmap` or `flurry`, and avoid the potential poisoning. I've used `dashmap` before, and it works well with stored `Arc`-ed types, such as the `Client` here. I think the deadlock caveats for [`DashMap::insert`](https://docs.rs/dashmap/4.0.2/dashmap/struct.DashMap.html#method.insert), for example, don't apply here, since we only hold references in the `client` method and not across `await` points. Might be worth a shot. What do you think?

l1h3r: Yes, looks like the concurrent map options have improved dramatically since I last looked 😄 I'll check out `DashMap` since we should avoid that deadlock caveat

PR #322, opened 2021-07-20
Implement the Identity Agent

### LINKED ISSUE #299: [Task] Implement Basic Agent
## Description

Implement a basic agent for the IOTA Identity framework for performing operations related to a stronghold, such as key creation and signing. The agent should be able to set up secure communication via the Stronghold P2P Communications crate. Lastly, it should allow developers to overwrite the agent or insert pre and post hooks to the default logic. 

## Motivation

Agents are a common idea in the SSI world. It allows different Identity-enabled entities to identify each other and speak in a common language to build trust and exchange trusted information. In addition, it allows a single entity to out-source its operations to another location (web -> desktop or x -> server), which may be more secure. This way we enable secure identity usage in unsecure environment such as the web. 

## Resources

- [Hyperledger Agent Javascript](https://github.com/hyperledger/aries-framework-javascript)
- [Hyperledger Agent Python](https://github.com/hyperledger/aries-cloudagent-python)
- [Stronghold Actors](https://github.com/iotaledger/stronghold.rs/tree/dev/client/src/actors)

## To-do list

- [ ] Implement agent for stronghold operations
- [ ] Set up secure communications between two agents via P2P
- [ ] Rust <-> WASM agent test
- [ ] Add hooks

## Change checklist

PhilippGackstatter:
# Description of change

The identity agent is a peer-to-peer communication framework for building SSI agents on IOTA Identity. It is intended to host implementations of the [DIDComm protocols](https://wiki.iota.org/identity.rs/specs/didcomm/overview) with future updates. Together with these protocols, this will, for example, allow for did-authenticated communication between two identities to exchange verifiable credentials or presentations.

This description serves as a technical introduction to the agent. For a high-level and less technical introduction, see the [blog post](https://blog.iota.org/the-iota-identity-actor-explained/) on the agent (formerly known as identity actor).

The most important dependency of the agent is libp2p. [What is libp2p?](https://docs.libp2p.io/introduction/what-is-libp2p/)

> The one-liner pitch is that libp2p is a modular system of *protocols*, *specifications* and *libraries* that enable the development of peer-to-peer network applications.

We use libp2p because it can easily secure transports using the noise protocol, is agnostic of transports so agents could conceivably communicate over TCP, websockets or Bluetooth, and because of how flexible it is, we can make it suit the agent nicely.

## Building an agent

```rust
let id_keys: IdentityKeypair = IdentityKeypair::generate_ed25519();
let addr: Multiaddr = "/ip4/0.0.0.0/tcp/0".parse()?;

let mut agent: Agent = AgentBuilder::new()
  .keypair(id_keys)
  .build()
  .await?;

agent.start_listening(addr).await?;
```

To build a minimal working agent, we generate a new `IdentityKeypair` from which the `AgentId` of the agent is derived. The `AgentId` is an alias for a `libp2p::PeerId`, which allows for cryptographically verifiable identification of a peer. This decouples the identity concept from the underlying network address, which is important if the agent roams across networks. If we want the agent to have the same `AgentId` across program executions, we need to store this keypair. Next we create the address for the agent to listen on. A `Multiaddr` is the address format in libp2p to encode addresses of various transports. Finally, we build the agent with a default transport, that supports DNS resolution and can use TCP or websockets.

## Processing incoming requests

To make the agent do something useful, we need handlers. A handler is some state with associated behavior that processes incoming requests. It will be invoked if the agent is able to deserialize the incoming request to the type the handler expects. The `Handler` is a trait that looks like this:

```rust
#[async_trait::async_trait]
pub trait Handler<REQ: HandlerRequest>: Debug + 'static {
  async fn handle(&self, request: RequestContext<REQ>) -> REQ::Response;
}
```

- It takes `&self` so it can modify its state, through appropriate mechanisms, such as locks. A handler will thus typically implement a shallow copy mechanism (e.g. using `Arc`) to share state.
- It takes the request it wants to handle, which needs to implement the `HandlerRequest` trait and needs to return the defined response type.
- This trait can be implemented multiple times so the same handler can process different request types.

Here is an example of a handler being attached on an `AgentBuilder`. We implement `RemoteAccounts`, an exemplary type that manages `Account`s remotely.

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub enum RemoteAccountsError {
  IdentityNotFound,
}

/// The struct that will be sent over the network.
/// When received by an agent, the contained DID is looked up.
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct RemoteAccountsGet(pub IotaDID);

impl HandlerRequest for RemoteAccountsGet {
  /// The result of the lookup procedure, either the corresponding IotaDocument, or an error.
  type Response = Result<IotaDocument, RemoteAccountsError>;

  /// `Endpoint`s are identifiers for requests which lets the remote agent determine the appropriate handler to invoke.
  fn endpoint() -> Endpoint {
    "remote_accounts/get".try_into().unwrap()
  }
}

/// Our thread-safe type that holds accounts that can be looked up by their DID.
#[derive(Debug, Clone, Default)]
pub struct RemoteAccounts {
  accounts: Arc<dashmap::DashMap<IotaDID, Account>>,
}

#[async_trait::async_trait]
impl Handler<RemoteAccountsGet> for RemoteAccounts {
  /// To handle the request, we take the HandlerRequest type wrapped in a RequestContext, which provides some
  /// useful information about the caller like their `AgentId`.
  async fn handle(&self, request: RequestContext<RemoteAccountsGet>) -> Result<IotaDocument, RemoteAccountsError> {
    self
      .accounts
      .get(&request.input.0)
      .map(|account| account.document().to_owned())
      .ok_or(RemoteAccountsError::IdentityNotFound)
  }
}

/// To build the agent with our custom functionality, we first build the agent itself
/// and attach the handler.
async fn build_agent() {
  let mut builder = AgentBuilder::new();
  builder.attach(RemoteAccounts::default());
}
```

An agent that receives a request will check whether a handler is attached that can handle the request's `Endpoint` and if so, will invoke it. In our case, the agent will call the `handle` function of the handler when a `RemoteAccountsGet` request is received. If we wanted, we could attach more handlers to the same agent, and even implement `Handler` for `RemoteAccounts` multiple times, in order to handle different request types.

## Sending requests

To invoke a handler on a remote agent, we send a type that implements `HandlerRequest`, such as `RemoteAccountsGet`.

```rust
let mut agent: Agent = builder.build().await?;

agent.add_agent_address(remote_agent_id, addr).await?;

let result: Result<IotaDocument, RemoteAccountsError> = agent
  .send_request(remote_agent_id, RemoteAccountsGet("did:iota:...".parse()?))
  .await?;
```

After building the agent and adding the address of the remote agent, we can send a request. The agent takes care of serializing the request, and attempts to deserialize the response into `<RemoteAccountsGet as HandlerRequest>::Response`.

## Agent modes

We've just seen an example of a synchronous request, one where we invoke a handler on a remote agent and wait for it to finish execution and return a result. Next to the `Agent` type we also have a `DidCommAgent` type. The latter additionally supports an asynchronous mode, where we send a request without waiting for the result of the handler invocation. Instead, we can explicitly await a request:

```rust
async fn didcomm_protocol(agent_id: AgentId, didcomm_agent: &DidCommAgent) -> AgentResult<()> {
  let thread_id: ThreadId = ThreadId::new();

  didcomm_agent
    .send_didcomm_request(agent_id, &thread_id, PresentationOffer::default())
    .await?;

  let request: DidCommPlaintextMessage<PresentationRequest> = didcomm_agent.await_didcomm_request(&thread_id).await?;

  Ok(())
}
```

This request mode is implemented to support the implementation of [DIDComm](https://identity.foundation/didcomm-messaging/spec/) protocols, which is why a separate `DidCommAgent` is defined that extends the `Agent`s functionality and handles the specifics of DIDComm. Note that the base `Agent` doesn't support the asynchronous mode, but the `DidCommAgent` supports the sychronous mode.

Here, the protocol expects us to first send a `PresentationOffer` request to the remote agent. This method call returns successfully if the request can be deserialized properly and if an appropriate handler exists on the remote agent, but the call might return before the handler on the remote has finished. According to the protocol we implement, we should expect the remote to send us a `PresentationRequest` so we explicitly call `await_didcomm_request` to await the incoming request on the same `ThreadId` that we sent our previous request on. This allows for imperative protocol implementations within a single handler. This is nice to have, because the alternative would be that each request invokes a separate handler in an agent, which would force protocol implementors to hold the state in some shared state, rather than implicitly in the function (such as the `thread_id` here). This setup is intended for DIDComm protocols, as it directly implements DIDComm concepts such as threads.

## Hooks

While the asynchronous mode of operation allows for implementation of DIDComm protocols, it does not yet allow users to hook in their own logic. This might be required to ask a user for consent before proceeding with the exchange of a credential.

Hooks are no longer part of this PR, and will be implemented later. The hook approaches thus far were unsatisfactory, but we keep the original text for reference.

<details>
<summary>Previous text</summary>
To that end, the `DidCommActor` has the concept of hooks, which are similar to handlers. There are various approaches to hooks:

- Explicit hooks: The protocol defines the hooks explicitly within the protocol. This is complementary to the implicit approach.
- Hook closures: The function that implements an async mode protocol takes hooks as explicit closure arguments. The actor itself has no concept of hooks in this model.
- Hook trait: Similar to the closures approach, but defining all hooks as methods on a trait and requiring users to implement that trait.

The actor currently implements the implicit hooks approach, mostly as an exploratory option, not necessarily because it was deemed the best one. Suppose we want to insert custom logic before `await_didcomm_request` returns the `PresentationRequest` in the above eample. One registers a hook like so:

```rust
async fn receive_presentation_request_hook(
    state: HookState,
    _actor: DidCommActor,
    req: RequestContext<DidCommPlaintextMessage<PresentationRequest>>,
) -> StdResult<DidCommPlaintextMessage<PresentationRequest>, DidCommTermination> {
    // Custom logic, like modifying the input or obtaining user consent.
    Ok(req.input)
}

builder
    .add_state(HookState::new())
    .add_hook(receive_presentation_request_hook)?;
```

The signature of a hook function is very similar to a handler, except for its return type. It needs to return the same type that it received as input or an instruction for the actor to terminate the connection. The endpoint of a hook includes the `/hook` postfix, and indicates to the actor that this hook should be called either before an `HandlerRequest` with a `didcomm/presentation_request` endpoint is sent, or before it is returned from `await_didcomm_request`, as is the case in this example.

### Limitations

The limitation of this hook implementation is thus that the user can only insert custom logic in between network requests, but not at arbitrary points in the protocol. With the current implementation (and some small modifications), it would also be possible to allow protocol implementors to define arbitrary hook points and call them.

In theory hooks can be attached to the same state as a handler, so they can access and modify the handlers state. However, in practice we will likely have a function that adds certain protocol handlers to an `ActorBuilder` and a user will add their hooks at other points. At this point, however, the ephemeral `HandlerBuilder` will have gone out of scope and no more typing guarantees could be given that a hook uses the same state object as a handler, e.g. there could be an incompatibility. Because of this lack of typing guarantees, it is currently not possible at all to attach a hook to some previously added state, as it seems very error-prone.

An alternative would be to use `std::any::TypeId` for identification of a state object, so if a handler and a hook take a certain `OBJ` type as state object, then the actor can determine the `TypeId` of the object and guarantee internally, that it will be called with the same object. The downside is, that users cannot add two different state objects that have the same `TypeId`.

</details>

## Implementation Details

This section goes into some details of agent internals.

The overall architecture can be seen as four layers. A libp2p layer, a commander layer to interact with the libp2p layer, the raw agent layer (which uses the commander) and the `DidCommAgent` on top. This architecture is strongly inspired by [stronghold-p2p](https://github.com/iotaledger/stronghold.rs/tree/dev/p2p).

- The p2p layer consists of a `libp2p::RequestResponse` protocol, which enforces on a type level that each request has a response. This naturally maps to the sync mode of the identity agent where each request has some response, as well as to the async mode where each request will be acknowledged.
- This layer has an `EventLoop` that concurrently polls the libp2p `Swarm` to handle its events as well as commands that are sent to it from the `NetCommander`.
- The commander layer, or `NetCommander` communicates with the event loop via channels and is thus the interface for the `EventLoop`.
- When the agent is built, it spawns an `EventLoop` in the background and interacts with it using the `NetCommander`.
- On incoming requests, the `EventLoop` spawns a new task and injects a clone of the agent into it (see `EventLoop::run` and its argument).

### Examples

This PR does not add examples to the `examples` directory. This is mostly due to the instability of the agent. Still, there are two "examples" for each mode of operation as part of the `tests` module, the remote account as a synchronous example, and the IOTA DIDComm presentation protocol as an asynchronous example (this doesn't implement the actual protocol, it just asserts that requests can be sent back and forth as expected). The DIDComm example in particular is very simple and minimal and mostly exists as a proof of concept for the async mode, but it also serves as an example for how a DIDComm protocol could potentially be implemented.

### DIDComm example setup

The async mode didcomm examples are worth explaining a little more. The implementation difficulty for these protocols comes mostly because of how flexible they are. In the presentation protocol for example, both the holder and verifier can initiate the exchange. On the agent level this means either calling the protocol explicitly to initiate it, or attaching a handler to let the agent handle the protocol in the background when a remote agent initiates. Thus, there is one function that implements the actual protocol for each of the roles (i.e. _holder_ and _verifier_ in the `presentation` protocol). As an example, this is what the signature of the holder role would look like:

```rust
pub(crate) async fn presentation_holder_handler(
  mut agent: DidCommAgent,
  agent_id: AgentId,
  request: Option<DidCommPlaintextMessage<PresentationRequest>>,
) -> AgentResult<()> { ... }
```

The holder can call this function to initiate the protocol imperatively by passing `None` as `request`. On the other hand, if the verifier initiates, the holder defines a handler that will inject the received `request`:

```rust
#[async_trait::async_trait]
impl DidCommHandler<DidCommPlaintextMessage<PresentationRequest>> for DidCommState {
  async fn handle(&self, agent: DidCommAgent, request: RequestContext<DidCommPlaintextMessage<PresentationRequest>>) {
    let result = presentation_holder_handler(agent, request.agent_id, Some(request.input)).await;

    if let Err(err) = result {
      log::error!("presentation holder handler errored: {err:?}");
    }
  }
}
```

and attaches it:

```rust
didcomm_builder.attach::<DidCommPlaintextMessage<PresentationRequest>, _>(DidCommState::new());
```

`DidCommState` holds the state for one or more DIDComm protocols. When a `PresentationRequest` is received, it calls the protocol function (`presentation_holder_handler`) to run through the protocol. This allows us to nicely reuse the `presentation_holder_handler` function as the core protocol implementation and only requires defining a thin handler method. The verifier can follow the same pattern for their side of the protocol.

### DIDComm agent internals

- In async mode, the `DidCommAgent` returns an acknowledgment if 1) a handler for the endpoint or a thread exists and 2) if the request can be deserialized into the expected type for the handler or thread (e.g. a DIDComm plaintext message)
- Timeouts can occur in two ways and both are configured via `AgentBuilder::timeout`.
  - A request sender can receive an `InboundFailure::Timeout` if the peer did not respond within the configured timeout. This happens on the event loop level and is handled by the `RequestResponse` protocol.
  - `DidCommAgent::await_didcomm_request` can time out. This is the same timeout value as for the underlying `RequestResponse` protocol. In such a case, the event loop will receive a timeout error, but since no entry in the thread hash map is waiting for a response, it is silently dropped. Thus, `await_didcomm_request` implements its own timeout, and automatically uses the same duration as the underlying protocol to ensure consistent behaviour. For this reason, the `await_didcomm_request` timeout is a per-agent configuration value, and not a parameter on the function, although that would also be possible if desired.

## Open Questions

- Should we replace `RequestMessage::from_bytes` with some serialization that's faster or more compact? For now, it seems good enough to use json. Some crate like rkyv would make the agent potentially less easily portable, if it would ever receive an implementation in a foreign language, while json is highly compatible. How much that downside matters with our single source code of truth approach is debatable. But this particular `RequestMessage` type is fairly simple anyway, and so the serialization performance gains might not be that significant?
- Who is responsible for, and when does cleanup of handler state or thread channels happen?

fixes #299 

## Type of change

- [ ] Documentation Fix

- A synchronous and asynchronous example implementation was added as `remote_account` and `didcomm` respectively.
- Tests were added in `identity-agent/src/tests` which use those examples.

## Change checklist


identity-actor/src/actor/actor_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl<'builder, OBJ> HandlerBuilder<'builder, Synchronous, OBJ>
+where
+  OBJ: Clone + Send + Sync + 'static,
+{
+  /// Add a synchronous handler function that operates on a shared state object and some
+  /// [`ActorRequest`]. The function will be called if the actor receives a request
+  /// on the given `endpoint` and can deserialize it into `REQ`. The handler is expected
+  /// to return an instance of `REQ::Response`.
+  pub fn add_sync_handler<REQ, FUT, FUN>(self, endpoint: &'static str, handler: FUN) -> Result<Self>
```
elenaf9, line 237: Do you really need the generic `FUN` closure here or wouldn't it also work with a normal function pointer `fn(OBJ, Actor, RequestContext<REQ>) -> FUT`? None of your test cases ever capture anything, and therefore can be converted to function pointers.

Apart from that, you can also consider to get rid of `FUT` and instead do `fn(..) -> BoxedFuture<'static, REQ::Response>`, which itself directly implements `Send`, `Sync` and a `'static` lifetime.

elenaf9 (approved):
Great work! I really like your how you solved the handling of inbound requests via the states/ handlers that are registered in the `ActorBuilder`. I do wonder though if there is a way to reduce the overall number of generics.
Couple of ideas below. I did not review everything in dept, but instead focused on the p2p part, which looks good to me!

Edit:
One question: is the `identity-actor` only used to communicate with between peers on the same system, or also between different devices/ networks. In case of the latter: how do you plan to connect peers that can not dial each other directly, e.g. because of firewalls or NAT?

identity-actor/src/actor/actor_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+/// Used to attach handlers and hooks to an [`ActorBuilder`].
+pub struct HandlerBuilder<'builder, MOD: SyncMode, OBJ>
+where
+  OBJ: Clone + Send + Sync + 'static,
+  MOD: 'static,
+{
+  pub(crate) object_id: ObjectId,
+  pub(crate) handler_map: &'builder mut HandlerMap,
+  _marker_obj: PhantomData<&'static OBJ>,
+  _marker_mod: PhantomData<&'static MOD>,
```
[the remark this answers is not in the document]
PhilippGackstatter: > Why are you using `PhantomData<&'static T>` here instead of just `PhantomData<T>`?

The `PhantomData` documentation says:

> ## Ownership and the drop check
> Adding a field of type PhantomData<T> indicates that your type owns data of type T. This in turn implies that when your type is dropped, it may drop one or more instances of the type T. This has bearing on the Rust compiler's [drop check] analysis.
>
> If your struct does not in fact own the data of type T, it is better to use a reference type, like PhantomData<&'a T> (ideally) or PhantomData<*const T> (if no lifetime applies), so as not to indicate ownership.

I just followed that suggestion. Not sure if there's an actual difference between `*const T` and `&'static T` in this case.

PhilippGackstatter:
> One question: is the `identity-actor` only used to communicate with between peers on the same system, or also between different devices/ networks. In case of the latter: how do you plan to connect peers that can not dial each other directly, e.g. because of firewalls or NAT?

The latter, indeed. We definitely want to add relay support in the future to enable those use cases, but it's not part of this initial, minimal version.

identity-actor/src/actor/actor_builder.rs
PhilippGackstatter: Thanks, I was able to get rid of all `FUN` generics. Very nice!

I kept `FUT` because requiring `BoxFuture` as a return type means users cannot simply provide a function pointer to some `async fn`, because such a function does not return `BoxFuture`. Instead, I think they'd have to implement a sync `fn` and return `BoxFuture` manually, which is too much of an ergonomic hit.

identity-account/src/types/identity_setup.rs
```diff
@@ -4,9 +4,11 @@
 use identity_core::crypto::PrivateKey;
 
 /// Configuration used to create a new Identity.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
 pub struct IdentitySetup {
   /// Use a pre-generated Ed25519 private key for the DID.
+  // TODO: Quick hack to get `IdentitySetup` to implement Serialize, by always using `None`
```
olivereanderson, line 10: Leftover TODO?

identity-actor/src/actor/actor.rs
```diff
@@ -0,0 +1,466 @@
...
+/// Handlers are registered at actor build time, using the [`ActorBuilder`](crate::ActorBuilder).
+///
+/// After shutting down the event loop of an actor using [`Actor::shutdown`], other clones of the
+/// actor will receive [`Error::Shutdown`] when attempting to interact with the event loop.
+#[derive(Clone)]
+pub struct Actor {
+  #[cfg(not(feature = "primitives"))]
+  pub(crate) commander: NetCommander,
+  #[cfg(feature = "primitives")]
+  pub commander: NetCommander,
```
olivereanderson, line 62: The type of `commander` seems to be the same regardless of the `primitives` feature being active or not, did you forget to change something during a refactor?

identity-account/src/types/identity_setup.rs
PhilippGackstatter: This is needed for the `RemoteAccount` and I currently see that more as an example of a synchronous protocol that we or our users might want to build at some point. We don't really have to merge that in with this PR. I suppose a better approach would then be to duplicate the `IdentitySetup` so it doesn't affect the existing types in crates outside `identity-actor`.

identity-actor/src/actor/actor.rs
PhilippGackstatter: It's only about visibility, enabling `primitives` users to access the commander directly for more low-level control. Regular users shouldn't really need to access the `NetCommander` so I left it crate-private.

identity-actor/src/actor/endpoint.rs
```diff
@@ -0,0 +1,111 @@
...
+///
+/// An endpoint can also represent a `hook` function, indicated by a `/hook` postfix.
+///
+/// Endpoints are separated by slashes and only allow alphabetic ascii characters and `_`.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct Endpoint {
+  name: String,
+  handler: String,
+  pub(crate) is_hook: bool,
+}
```
olivereanderson, line 24: As we discussed the `(name, handler)` split does currently not seem to be used after initialization and in many cases the endpoint will be known at compile time. Hence it might be useful to change the internal representation to the following: 
```suggestion 
#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
pub struct Endpoint {
  name: Cow<'static, str>, 
  pub(crate) is_hook: bool
}
```

identity-actor/src/p2p/net_commander.rs
```diff
@@ -0,0 +1,140 @@
...
+
+  pub async fn shutdown(&mut self) -> ActorResult<()> {
+    let (sender, receiver) = oneshot::channel();
+    self
+      .send_command(SwarmCommand::Shutdown {
+        response_channel: sender,
+      })
+      .await?;
+    receiver.await.map_err(|_| Error::Shutdown)
+  }
```
olivereanderson, line 103: All the public methods in this `impl` block are missing documentation.

identity-actor/src/remote_account/handler.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    Ok(Self {
+      builder: Arc::new(Mutex::new(builder)),
+      accounts: Arc::new(DashMap::new()),
+    })
+  }
+
+  pub async fn create(
+    self,
+    _actor: Actor,
```
olivereanderson, line 40: Please add a comment about why `_actor` is necessary.

identity-actor/src/remote_account/handler.rs
```diff
@@ -0,0 +1,66 @@
...
+    Ok(doc)
+  }
+
+  pub async fn list(self, _actor: Actor, _request: RequestContext<IdentityList>) -> Vec<IotaDID> {
+    self.accounts.iter().map(|entry| entry.key().to_owned()).collect()
+  }
+
+  pub async fn get(
+    self,
+    _actor: Actor,
```
olivereanderson, line 55: Same as https://github.com/iotaledger/identity.rs/pull/322/files#r857628566

identity-actor/src/remote_account/handler.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// A proof-of-concept implementation of a remote `Account` with very basic operations
+/// and disabled tangle interaction.
+#[derive(Clone)]
+pub struct RemoteAccount {
+  builder: Arc<Mutex<AccountBuilder>>,
+  accounts: Arc<DashMap<IotaDID, Account>>,
+}
+
+impl RemoteAccount {
```
olivereanderson, line 28: The public methods in this `impl` block should ideally be documented.

identity-actor/src/remote_account/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+mod requests;
+
+pub use handler::RemoteAccount;
+pub use requests::IdentityCreate;
+pub use requests::IdentityGet;
+pub use requests::IdentityList;
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[non_exhaustive]
+pub enum RemoteAccountError {
```
olivereanderson, line 14: Consider adding some documentation explaining that this is part of a PoC (which is why it is OK not to implement the Error trait here).

identity-actor/src/remote_account/requests.rs
```diff
@@ -0,0 +1,46 @@
...
+impl SyncActorRequest for IdentityList {
+  type Response = Vec<IotaDID>;
+
+  fn endpoint() -> Endpoint {
+    "remote_account/list".parse().unwrap()
+  }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct IdentityGet(pub IotaDID);
```
olivereanderson, line 38: Consider adding some (simple) documentation for the public types here.

identity-core/src/common/one_or_many.rs
```diff
@@ -203,6 +204,51 @@ impl<'a, T> Iterator for OneOrManyIter<'a, T> {
...
+// =============================================================================
+// IntoIterator
+// =============================================================================
+
+enum Either<L, R> {
+  Left(L),
+  Right(R),
+}
+
+pub struct OneOrManyIterator<T> {
```
olivereanderson, line 216: nit: Maybe `OneOrManyIntoIter` would be more consistent with the naming scheme from the standard library (see https://doc.rust-lang.org/std/vec/struct.IntoIter.html and https://doc.rust-lang.org/std/slice/struct.Iter.html where the later is returned from `Vec::iter`). 

Also consider adding some minimal documentation here like: 
```
/// This struct is created by the `into_iter` method on [`OneOrMany`] (provided by the [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) trait).
```

Finally it would be great with a simple test that confirms that works as expected for a `OneOrMany` collection with zero, one and two (or more) elements.

identity-actor/src/actor/config.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::time::Duration;
+
+pub struct ActorConfig {
```
olivereanderson, line 6: Again add some documentation here.

identity-actor/src/actor/errors.rs
```diff
@@ -0,0 +1,117 @@
...
+pub enum Error {
+  #[error("{context}: {source}")]
+  TransportError {
+    context: &'static str,
+    source: libp2p::TransportError<std::io::Error>,
+  },
+  #[error("invalid endpoint")]
+  InvalidEndpoint,
+  #[error("{0}")]
+  OutboundFailure(#[from] OutboundFailure),
```
olivereanderson, line 21: The `From` implementation brings a type from an unstable external crate into our public API. Furthermore using `#[from]` in this error implies that it is also `source` which violates the following part of our [error handling guidelines](https://github.com/iotaledger/identity.rs/issues/534):
> Return source errors via Error::source unless the source error's message is included in your own error message in Display.

olivereanderson: I would suggest the following here: 
1. Change `#[from]` with `#[source]` . 
2. Decorate this variant with `#[non_exhaustive]`. Together with 1. this will prevent the unstable `OutboundFailure` external type from being part of our public API. 
3. Change the error message to something that does not include the error message of the source error. 

Regarding point 2. If one needs to construct this variant from other types then wrap a `Box<dyn Error +'static + Send + Sync>` and remove the `#[non_exhaustive]` annnotation.

identity-actor/src/actor/errors.rs
```diff
@@ -0,0 +1,117 @@
...
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, thiserror::Error)]
+/// Errors that can occur during the actor operation.
+pub enum Error {
+  #[error("{context}: {source}")]
+  TransportError {
+    context: &'static str,
+    source: libp2p::TransportError<std::io::Error>,
```
olivereanderson, line 16: Unfortunately this exposes a type from an unstable external crate in our public API. Also the error message should not contain the error message of source (when a field is named source `thiserror` will automatically return it via `Error::source`). 
 
```suggestion 
#[non_exhaustive]
#[error("transport error: {0}")]
TransportError(&'static str, #[source] libp2p::TransportError<std::io::Error>), 
```

identity-actor/src/actor/actor.rs
```diff
@@ -0,0 +1,279 @@
...
+  /// request to this [`PeerId`].
+  pub async fn add_addresses(&mut self, peer_id: PeerId, addresses: Vec<Multiaddr>) -> ActorResult<()> {
+    self
+      .commander()
+      .add_addresses(peer_id, OneOrMany::Many(addresses))
+      .await
+  }
+
+  /// Sends a synchronous request to a peer and returns its response.
+  pub async fn send_request<REQ: SyncActorRequest>(
```
olivereanderson, line 143: It might be useful to explain the reason why this method is or needs to be async given that the request itself is synchronous.

identity-actor/src/actor/actor.rs
```diff
@@ -0,0 +1,279 @@
...
+      })??;
+
+    serde_json::from_slice::<REQ::Response>(&response).map_err(|err| Error::DeserializationFailure {
+      location: ErrorLocation::Local,
+      context: "send request".to_owned(),
+      error_message: err.to_string(),
+    })
+  }
+
+  pub fn handle_request(self, request: InboundRequest) {
```
olivereanderson, line 179: Why does this method consume `self`?

identity-actor/src/actor/actor_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+    TRA::Output: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    TRA::Dial: Send + 'static,
+    TRA::Listener: Send + 'static,
+    TRA::ListenerUpgrade: Send + 'static,
+    TRA::Error: Send + Sync,
+  {
+    let executor = Box::new(|fut| {
+      cfg_if::cfg_if! {
+        if #[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))] {
+          tokio::spawn(fut);
```
olivereanderson, line 150: Question: Would it be possible to make this work for an arbitrary async runtime?

identity-actor/src/actor/actor_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /// Grants low-level access to the object map for use in bindings.
+  #[cfg(feature = "primitives")]
+  pub fn objects(&mut self) -> &mut ObjectMap {
+    &mut self.objects
+  }
+
+  /// Add a new shared state object and returns a [`ActorHandlerBuilder`] which can be used to
+  /// attach handlers and hooks that operate on that object.
```
olivereanderson, line 107:
```suggestion 
  /// Add a new shared state object and returns an [`ActorHandlerBuilder`] which can be used to
  /// attach handlers and hooks that operate on that object.

File: identity-actor/src/tests/actor.rs [hunk 20]
```diff
@@ -0,0 +1,343 @@
...
+use crate::actor::ErrorLocation;
+use crate::actor::RequestContext;
+use crate::actor::Result as ActorResult;
+use crate::actor::SyncActorRequest;
+use crate::remote_account::IdentityGet;
+use crate::remote_account::IdentityList;
+use crate::tests::try_init_logger;
+
+use super::default_listening_actor;
+use super::default_sending_actor;
```
olivereanderson, line 25: mega nit: It is more idiomatic not to import functions.

identity-actor/src/actor/actor.rs
PhilippGackstatter: Thanks for this comment. It shows that the current language around sync and async actor requests is not ideal, because devs tend to think about it in terms of sync/async functions in Rust.
The synchronicity of an actor request is a higher-level concept than that of Rust functions. Such an actor request is still sent using async I/O, and hence the function needs to be async. But conceptually, the request is synchronous in that it wait for the peer's handler to finish execution. Does that make sense?

To my knowledge, sync/async is the most appropriate terminology to use, but it's apparently still confusing. If you have alternative ideas for what language to use here, I'd be happy to change it if it fits.

identity-actor/src/tests/actor.rs
PhilippGackstatter: What would you say is the alternative, use the fully qualified path wherever they're needed?

Generally I would agree, but these functions specifically are used very often (in every test function), and so it's just more convenient to refer to them explicitly. I suppose this is a similar argument as to why the `Result::Ok` and `Result::Err` enum variants are in the default Rust prelude, rather than requiring the `Result` prefix.

olivereanderson: > What would you say is the alternative, use the fully qualified path wherever they're needed?

I would refer to them by `super` (i.e. `super::default_listening_actor`), but if they are really used in every single test I understand that it is less ideal to always type super so I can agree to use a full import in this instance (one typically does `use super::*` in tests anyway).

identity-account/src/types/identity_setup.rs
PhilippGackstatter: Removed the TODO and cloned the type for the `RemoteAccount` instead.

identity-actor/src/actor/endpoint.rs
PhilippGackstatter: Change the type to your suggestion, except without `is_hook`.

identity-actor/src/actor/actor_builder.rs
PhilippGackstatter: At compile-time? Yes, probably. At runtime? Tried it before, but wasn't able to get it working.
We could support `async_std` and `tokio` with feature flags. Based on this flag, we'd also have to use different transports in `libp2p`, so this is not completely trivial. Definitely something to look into for stabilization, but would not focus on it right now.

PhilippGackstatter:
Addressed all your comments so far, @olivereanderson, thanks for taking a detailed look!

olivereanderson:
> Addressed all your comments so far, @olivereanderson, thanks for taking a detailed look!

Thanks!

abdulmth (approved):
Looks good to me, minor comments, I like the recent refactor, it also made the code easier to understand IMO. 
I think it would be nicer if we have better documentation (maybe similar to actix) or maybe an activity diagram that can quickly give an idea how everything works. Also, tests require more documentation. 
It also might be necessary after the recent refactor to change the name of the concept and crate name to Agent since "Actor" doesn't fit anymore.

> How do actors clean up state? Is that supposed to be done in actors themselves, and/or do we need to expose additional disconnect handlers/events that actors can react to?

Is this still relevant to this PR after the refactor? I think it depends on the use case if the actor state can be cleared remotely or locally.

> Is the assumption that any await_message(thread_id) is preceded by a send_message on the same thread_id correct?

This is something that we have to include in the documentation. I think `await_message` already has that. I'm not sure when multiple messages are useful, but for now we might keep it as is IMO.

> Should we replace RequestMessage::from_bytes with some serialization that's faster or more compact?

I don't think this is a concern for now, we might add something like that under a feature flag in the future if that would work with wasm and if the performance gain is significant.

identity-actor/src/actor/endpoint.rs
```diff
@@ -0,0 +1,187 @@
...
+      self.name = Cow::Owned(name + "/" + HOOK);
+    }
+    self
+  }
+
+  /// Validates the endpoint.
+  fn validate(string: &str) -> ActorResult<()> {
+    let mut split: Split<char> = string.split('/');
+
+    let name: &str = split.next().expect("split should have at least one element");
```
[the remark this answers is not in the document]
PhilippGackstatter: `str::split` always returns an array with at least one element, so we don't have to handle the empty case. (Splitting the empty string returns the empty string).

I'm changing the message to be clearer, though, thanks for pointing it out.

```
let name: &str = split.next().expect("split always returns at least one element");
```

identity-actor/src/actor/endpoint.rs
```diff
@@ -0,0 +1,187 @@
...
+#[serde(try_from = "String", into = "String")]
+pub struct Endpoint {
+  name: Cow<'static, str>,
+}
+
+impl Endpoint {
+  /// Returns whether this endpoint is a hook.
+  pub fn is_hook(&self) -> bool {
+    let split: Split<char> = self.name.as_ref().split('/');
+    split.last().expect("split should have at least one element") == HOOK
```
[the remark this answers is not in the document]
PhilippGackstatter: See below, but this method was removed because it references hooks which will most likely no longer be part of the actor itself but implemented on a higher level.

identity-actor/src/actor/system.rs
```diff
@@ -0,0 +1,243 @@
...
+
+  pub fn commander(&self) -> &NetCommander {
+    &self.commander
+  }
+
+  /// Start listening on the given `address`. Returns the first address that the actor started listening on, which may
+  /// be different from `address` itself, e.g. when passing addresses like `/ip4/0.0.0.0/tcp/0`. Even when passing a
+  /// single address, multiple addresses may end up being listened on. To obtain all those addresses, use
+  /// [`System::addresses`]. Note that even when the same address is passed, the returned address is not deterministic,
+  /// and should thus not be relied upon.
```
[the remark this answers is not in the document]
PhilippGackstatter: Yes, that's due to libp2p.

identity-actor/src/didcomm/system.rs
```diff
@@ -0,0 +1,295 @@
...
+    Self {
+      actors,
+      threads_receiver: DashMap::new(),
+      threads_sender: DashMap::new(),
+      identity,
+    }
+  }
+}
+
+/// An actor system can be used to send requests to remote actors, and fowards incoming requests
```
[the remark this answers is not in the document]
PhilippGackstatter: Rewrote it entirely:

> /// An actor system fowards incoming requests to attached actors and can be used to send requests to remote actors.

Is that better?

PhilippGackstatter:
> Also, tests require more documentation.

Can you be more specific what parts you would like to see documented? I now documented one of the end-to-end tests in more detail and most of them follow a similar pattern. Documenting tests thoroughly increases the maintenance burden by a lot, so I would refrain from just documenting them all. So perhaps you can point out more specifically what parts are unclear.

identity-actor/src/didcomm/system.rs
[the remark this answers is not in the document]
olivereanderson: > Rewrote it entirely:
> 
> > /// An actor system fowards incoming requests to attached actors and can be used to send requests to remote actors.
> 
> Is that better?

That is definitely better in my opinion. What do you think @abdulmth?

identity-actor/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+serde = { version = "1.0", default-features = false, features = ["derive"] }
+serde_json = { version = "1.0", default-features = false }
+thiserror = { version = "1.0", default-features = false }
+tokio = { version = "1.15", default-features = false, features = ["rt", "time"] }
+uuid = { version = "0.8", default-features = false, features = ["v4", "serde"] }
+
+[target.'cfg(all(target_arch = "wasm32", not(target_os = "wasi")))'.dependencies]
+libp2p = { version = "*", features = ["wasm-bindgen", "wasm-ext", "wasm-ext-websocket"] }
+uuid = { version = "*", features = ["wasm-bindgen"] }
+wasm-bindgen-futures = { version = "0.4.29", default-features = false }
```
olivereanderson, line 31: Since we decided to go with something like gRPC to support actors/agents for different programming languages I think this can be removed?

identity-actor/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+criterion = { version = "0.3", default-features = false, features = ["stable"] }
+identity-account = { version = "=0.5.0", path = "../identity-account", default-features = false, features = ["send-sync-storage"] }
+pretty_env_logger = { version = "0.4", default-features = false }
+
+[[bench]]
+name = "benchmark"
+harness = false
+
+[features]
+default = []
```
olivereanderson, line 43: Can this be removed?

identity-actor/Cargo.toml
olivereanderson, line 31: Since we decided to go with something like gRPC to support calling actors/agents from different programming languages I think this can be removed?

identity-actor/src/actor/endpoint.rs
```diff
@@ -0,0 +1,141 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::actor::Error;
+use crate::actor::Result as ActorResult;
+
+/// A path-like identifier for an actor request. As an example, `identity/create`
+/// could be an endpoint of the request "create" within the "identity" namespace.
+///
+/// Endpoints are separated by slashes and only allow alphabetic ascii characters and `_`.
```
olivereanderson, line 17: `slashes` in plural is a bit misleading as an endpoint may only contain a single slash.

identity-actor/src/actor/errors.rs
```diff
@@ -0,0 +1,121 @@
...
+  InvalidEndpoint,
+  #[non_exhaustive]
+  #[error("failure during sending an outbound request and receiving the response")]
+  OutboundFailure(#[source] OutboundFailure),
+  #[error("unexpected request `{0}`")]
+  UnexpectedRequest(String),
+  #[error("handler invocation error: {0}")]
+  HandlerInvocationError(String),
+  #[error("hook invocation error: {0}")]
+  HookInvocationError(String),
```
olivereanderson, line 28: Hooks are no longer part of this PR per the new description so perhaps this variant can be removed for now?

identity-actor/src/actor/errors.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// Errors that can occur on the remote actor.
+#[derive(Debug, thiserror::Error, serde::Serialize, serde::Deserialize)]
+pub enum RemoteSendError {
+  #[error("unexpected request: {0}")]
+  UnexpectedRequest(String),
+  #[error("handler invocation error: {0}")]
+  HandlerInvocationError(String),
+  #[error("hook invocation error: {0}")]
+  HookInvocationError(String),
```
olivereanderson, line 60: Perhaps this variant can be removed for now?

identity-actor/src/actor/errors.rs
```diff
@@ -0,0 +1,121 @@
...
+  AwaitTimeout(ThreadId),
+  #[error("actor was shutdown")]
+  Shutdown,
+  #[error("actor identity missing")]
+  IdentityMissing,
+}
+
+/// Errors that can occur on the remote actor.
+#[derive(Debug, thiserror::Error, serde::Serialize, serde::Deserialize)]
+pub enum RemoteSendError {
```
olivereanderson, line 54: Are we sure we don't want to mark this as `#[non_exhaustive]`? I know everything in this crate will be marked as unstable, but if we already don't know whether more variants might be needed we should prepare for that now.

identity-actor/src/actor/system.rs
```diff
@@ -0,0 +1,235 @@
...
+impl System {
+  pub(crate) fn new(commander: NetCommander, state: Arc<SystemState>) -> System {
+    Self { commander, state }
+  }
+
+  pub(crate) fn state(&self) -> &SystemState {
+    self.state.as_ref()
+  }
+
+  /// Returns the [`PeerId`] that other peers can securely identify this actor with.
```
olivereanderson, line 63: > /// Returns the [`PeerId`] that other peers can securely identify this **actor** with

Is `actor` still the right word here?

identity-actor/src/actor/system.rs
```diff
@@ -0,0 +1,235 @@
...
+    // this function will return while the event loop keeps running. Ideally we could then join on the background task
+    // to wait for all handlers to finish gracefully. However, not all spawn functions return a JoinHandle,
+    // such as wasm_bindgen_futures::spawn_local. The current alternative is to use a non-graceful exit,
+    // which breaks the event loop immediately and returns an error through all open channels that require a result.
+    self.commander_mut().shutdown().await
+  }
+
+  /// Associate the given `peer_id` with an `address`. This needs to be done before sending a
+  /// request to this [`PeerId`].
+  pub async fn add_address(&mut self, peer_id: PeerId, address: Multiaddr) -> ActorResult<()> {
```
olivereanderson, line 108: Consider explaining how this is different from `start_listening` in the documentation.

identity-actor/src/actor/system_builder.rs
```diff
@@ -0,0 +1,205 @@
...
+    REQ::Response: Send,
+  {
+    self.actors.insert(
+      REQ::endpoint(),
+      Box::new(ActorWrapper::new(actor)) as Box<dyn AbstractActor>,
+    );
+  }
+
+  /// Build the actor with a default transport which supports DNS, TCP and WebSocket capabilities.
+  #[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
```
olivereanderson, line 98: Can this be compiled unconditionally now?

identity-actor/src/didcomm/system_builder.rs
```diff
@@ -0,0 +1,167 @@
...
+    );
+  }
+
+  /// Attaches an [`Actor`] to this system.
+  ///
+  /// This means that when the system receives a request of type `REQ`, it will invoke this actor.
+  ///
+  /// Calling this method with a `REQ` type whose endpoint is already attached to an actor
+  /// will overwrite the previous attachment.
+  pub fn attach<REQ, ACT>(&mut self, actor: ACT)
```
olivereanderson, line 99: I think this method can be removed. As far as I can see the standard actors are discarded once the `DIDCommSystem` is built or am I wrong about that?

identity-actor/src/p2p/behaviour.rs
```diff
@@ -0,0 +1,75 @@
...
+use futures::AsyncWriteExt;
+use libp2p::core::upgrade;
+use libp2p::core::ProtocolName;
+use libp2p::request_response::RequestResponseCodec;
+
+use crate::p2p::RequestMessage;
+use crate::p2p::ResponseMessage;
+
+#[derive(Debug, Clone)]
+pub struct ActorProtocol();
```
olivereanderson, line 17: Document this public struct

identity-actor/src/p2p/behaviour.rs
```diff
@@ -0,0 +1,75 @@
...
+#[derive(Clone)]
+pub struct ActorRequestResponseCodec();
```
olivereanderson, line 19: Document this public struct

identity-actor/src/p2p/message.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::fmt::Debug;
+
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::actor::Endpoint;
+use crate::actor::RequestMode;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct RequestMessage {
```
olivereanderson, line 13: Document this public method

identity-actor/src/p2p/message.rs
```diff
@@ -0,0 +1,39 @@
...
+
+  pub fn from_bytes(bytes: &[u8]) -> std::io::Result<Self> {
+    serde_json::from_slice::<'_, Self>(bytes)
+      .map_err(|err| std::io::Error::new(std::io::ErrorKind::InvalidData, err.to_string()))
+  }
+
+  pub fn to_bytes(&self) -> std::io::Result<Vec<u8>> {
+    serde_json::to_vec(self).map_err(|err| std::io::Error::new(std::io::ErrorKind::InvalidData, err.to_string()))
+  }
+}
```
olivereanderson, line 36: Document these public methods

identity-actor/src/actor/system.rs
PhilippGackstatter: Thanks. Maybe `add_peer_address` is even more appropriate, given that it associates a `peer_id` with an `address`? Otherwise, I'm happy with your suggestion.

olivereanderson: `add_peer_address` is also nice. Use whichever you think is best.

identity-actor/src/didcomm/dcpm.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl<T> DidCommRequest for DidCommPlaintextMessage<T>
+where
+  T: DidCommRequest,
+{
+  fn endpoint() -> Endpoint {
+    T::endpoint()
+  }
+}
```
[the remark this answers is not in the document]
PhilippGackstatter: Can you elaborate what you mean here?

identity-actor/src/didcomm/system_builder.rs
PhilippGackstatter: No, this adds actors to the `inner` `SystemBuilder`, which is later used to build a `System` which is then part of `DidCommSystem`. The `test_didcomm_system_supports_actor_requests` tests for exactly that.

identity-actor/src/actor/endpoint.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct Endpoint {
+  name: Cow<'static, str>,
+}
+
+impl Endpoint {
+  /// Validates the endpoint.
+  fn validate(string: &str) -> ActorResult<()> {
+    let mut split: Split<'_, char> = string.split('/');
+
+    let namespace: &str = split.next().expect("split always returns at least one element");
```
olivereanderson, line 29: Could the use of `expect` here be an attack vector for malicious actors wanting to bring down a server?

identity-actor/benches/benchmark.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  #[async_trait::async_trait]
+  impl Actor<IdentityCreate> for RemoteAccount {
+    async fn handle(&self, request: RequestContext<IdentityCreate>) -> Result<IotaDocument, RemoteAccountError> {
+      let account: Account = self.builder.lock().await.create_identity(request.input.into()).await?;
+      let doc = account.document().to_owned();
+      self.accounts.insert(account.did().to_owned(), account);
+      Ok(doc)
```
olivereanderson, line 132: This is a fairly expensive operation hence I am concerned the benchmark ends up primarily benchmarking account creation which is an operation implemented in another crate. 

I do however agree with benching realistic use cases. Hence it would be nice if we could find another such example, but where less time is spent inside the handler. Perhaps something like benchmarking sending a presentation request/offer?

identity-actor/src/actor/endpoint.rs
PhilippGackstatter: This comment still applies, but apparently it's still confusing. Not sure how to make it less confusing.

https://github.com/iotaledger/identity.rs/pull/322#discussion_r889674514

olivereanderson: Thanks for elaborating. We could consider adding a comment explaining that once the never type `!` lands the call to expect can hopefully be replaced with a match where the `None` variant is mapped to `!`.

identity-actor/src/didcomm/dcpm.rs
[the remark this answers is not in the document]
olivereanderson: On second thought I think this is fine.

identity-actor/benches/benchmark.rs
PhilippGackstatter: Added a second benchmark that purely tests DIDComm operations.

PhilippGackstatter:
Thanks for the review! I addressed the comments.
- Added some documentation for didcomm tests.
- Removed one of the larger tests whose assertions are already tested in other tests, which should also make it easier to understand the test module. Let me know if it's better now, but feel free to request more docs if it's not yet sufficient.
- Renamed `System` to `Agent`, and `Actor` to `Handler`.
- Introduced a type alias `AgentId` for `PeerId` and use it everywhere, except for the `p2p` module (to keep modules loosely coupled).

Note that the `Cargo.toml` contains some `TODO`s which I don't think must be addressed in this PR. It has to do with some of our crates (`identity_iota_client` and `identity_account_storage`) not properly supporting `default-features = false`, which should be addressed in a separate PR.

identity_agent/benches/remote_account_benchmark.rs
```diff
@@ -0,0 +1,151 @@
...
+
+  let mut sender: Agent = AgentBuilder::new().build().await.unwrap();
+
+  sender.add_agent_address(receiver_agent_id, addr).await.unwrap();
+
+  (receiver, receiver_agent_id, sender)
+}
+
+fn bench_create_remote_account(c: &mut Criterion) {
+  static ITERATIONS: &[usize] = &[100, 10_000, 100_000];
```
olivereanderson, line 36: Same comment here as in the previous benchmark file. I don't see why these iterations are necessary.

identity_agent/src/agent/agent_state.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::agent::AgentConfig;
+use crate::agent::AgentId;
+use crate::agent::HandlerMap;
+
+/// The internal state of a `Agent`.
```
olivereanderson, line 8:
```suggestion
/// The internal state of an `Agent`
```

identity_agent/src/didcomm/agent.rs
```diff
@@ -0,0 +1,294 @@
...
+///
+/// Note: Currently an incomplete implementation.
+#[derive(Debug, Clone)]
+pub struct DidCommAgentIdentity {
+  // TODO: This type is meant to be used in a future update.
+  #[allow(dead_code)]
+  pub document: IotaDocument,
+}
+
+/// The internal state of a [`Agent`].
```
olivereanderson, line 41:
```suggestion
/// The internal state of a [`DidCommAgent`].
```

identity_agent/src/didcomm/agent.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    Ok(())
+  }
+
+  /// Wait for a message on a given `thread_id`. This can only be called successfully if
+  /// [`DidCommAgent::send_didcomm_request`] was called on the same `thread_id` previously.
+  ///
+  /// This will return a timeout error if no message is received within the duration passed
+  /// to [`DidCommAgentBuilder::timeout`](crate::didcomm::DidCommAgentBuilder::timeout).
+  pub async fn await_didcomm_request<T: DeserializeOwned + Send + 'static>(
```
olivereanderson, line 181: > This can only be called successfully if
  /// [`DidCommAgent::send_didcomm_request`] was called on the same `thread_id` previously. 

Should the docs also mention what to expect if `DidCommAgent::send_didcomm_request` was called multiple times on the same `ThreadId` previously?

identity_iota/Cargo.toml
```diff
@@ -59,6 +60,10 @@ revocation-bitmap = [
 # Breaking changes to types and functions behind this flag are not covered by semver.
 unstable-encryption = ["identity_account/encryption"]
 
+# Enables support for the unstable identity agent.
+# Breaking changes to types and functions behind this flag are not covered by semver.
+unstable-agent = ["identity_agent"]
```
olivereanderson, line 65: nit: 
```suggestion
unstable-agent = ["dep:identity_agent"]
```

identity_iota/Cargo.toml
```diff
@@ -16,6 +16,7 @@ description = "Framework for Self-Sovereign Identity with IOTA DID."
 # identity_comm = { version = "=0.5.0-dev.4", path = "../identity_comm", optional = true }
 identity_account = { version = "=0.6.0", path = "../identity_account", default-features = false, optional = true }
 identity_account_storage = { version = "=0.6.0", path = "../identity_account_storage", default-features = false, optional = true }
+identity_agent = { version = "=0.6.0", path = "../identity_agent", default-features = false, optional = true }
```
olivereanderson, line 19: nit: 
This implicitly creates a feature of the name `identity_agent` which we don't really want (at least not at this point).  By introducing my suggestion on line 65 (available since Rust 1.60) this can be avoided.

olivereanderson (approved):
This is great, 
Thank you for all the extra effort you put in to finding the best possible names, write great documentation and vastly improve the readability of the tests. 

As we discussed we may want to change the architecture somewhat in the feature in order to give users more control over performance, but that should be considered after we've merged this (unstable feature) and got some experience working with it (when implementing some DIDComm protocols). 

Finally,  I left a few comments you may wish to address before merging this.

identity_agent/src/agent/agent.rs
```diff
@@ -0,0 +1,260 @@
...
+impl Clone for Agent {
+  /// Produce a shallow copy of the agent, which uses the same event loop as the
+  /// agent that it was cloned from.
+  fn clone(&self) -> Self {
+    Self {
+      commander: self.commander.clone(),
+      state: self.state.clone(),
+    }
+  }
+}
```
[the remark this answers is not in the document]
PhilippGackstatter: I think it used to require a manual impl, but it no longer does. Still, the explicit docs are nice, so I'll keep it with the additional note.

identity_agent/src/didcomm/agent.rs
```diff
@@ -0,0 +1,294 @@
...
+/// user-attached handlers. Agents can be cloned without cloning the event loop, and doing so
+/// is a cheap operation.
+///
+/// Handlers are attached at agent build time, using the [`DidCommAgentBuilder`](crate::didcomm::DidCommAgentBuilder).
+///
+/// A `DidCommAgent` supports attachement of both [`Handler`](crate::agent::Handler)s and
+/// [`DidCommHandler`](crate::didcomm::DidCommHandler)s.
+///
+/// After shutting down the event loop of an agent using [`DidCommAgent::shutdown`], other clones of the
+/// agent will receive [`Error::Shutdown`] when attempting to interact with the event loop.
```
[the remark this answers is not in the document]
PhilippGackstatter: > While an [`Agent`] only supports attachements of synchronous [`Handler`](crate::agent::Handler)s, a [`DidCommAgent`] additionally supports asynchronous [`DidCommHandler`](crate::didcomm::DidCommHandler)s.

Would that be too concise?

identity_agent/src/didcomm/agent.rs
PhilippGackstatter: > Calling `send_didcomm_request` multiple times still only allows to await one message on the thread.

Added this. We'll see during DIDComm protocol implementations whether this is too restrictive and can change it then if we need it.

identity_agent/src/didcomm/agent.rs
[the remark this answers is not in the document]
olivereanderson: What about something like?  
> A [`DidCommAgent`] is an extension of an [`Agent`] with support for sending and awaiting [`DidCommRequests`].

Issue #321, opened 2021-07-19
[Task] Implement a serializable error


PhilippGackstatter:
## Motivation

The upcoming actor exchanges messages over the network, where errors can occur. The current error enums we use are not serializable, and thus cannot be exchanged between actors. Another motivation may be to improve the currently stringly-typed errors in the WebAssembly bindings. In order to pass a value to JavaScript, it needs to be serializable. Errors also need to be exposable to future bindings, which often involves (de)serialization.

## Description

Implement an error type that is serializable for over-the-network and bindings usage. WebAssembly bindings are a good case study. We currently use error enums in Rust, but any bindings that use a C FFI as the common interface, such as Wasm, can only use C-style enums. We have talked about this a number of times, and I want to summarize the approaches we've discussed, as it is not entirely clear yet, which is the best.

The general idea is to use an error code approach, and we've discussed variants of it. As an example we consider a part of the account error enum.

```rust
pub enum Error {
  #[error("Stronghold snapshot password not found")]
  StrongholdPasswordNotSet,
  #[error("Stronghold error: {0}")]
  StrongholdResult(String),
  #[error(transparent)]
  IoError(#[from] std::io::Error),
  #[error(transparent)]
  StrongholdError(#[from] iota_stronghold::Error),
  ...
}
```

The simplest approach would be to map each of the variants to some error code. For the variants that contain associated data, information will be lost (or preserved in stringified form). On the receiver side, the error cannot be reconstructed, since not all information is available. Furthermore, since no enum is defined, the receiver only has the bare number but has a hard time attaching a meaning to it.

Thus, to build on this approach, we define roughly this Rust code:

```rust
pub struct Error {
  code: ErrorCode,
  stringified: String,
}

pub enum ErrorCode {
  InternalError = 0,
  StrongholdPasswordNotSet = 1,
  StrongholdResult = 2,
  ...
}
```

To convert the example variants into this Error, we might do:

- the `StrongholdPasswordNotSet` variant is easily translated as some error code.
- the `StrongholdResult` has an associated `String`. We can translate the variant itself to some number, and additionally transmit the stringified error.
- The `IoError` is a fatal error from the other side, and the caller cannot do much about it, similar to a 5XX HTTP status. These errors can be mapped to a catch-all `InternalError` (or similar) and also stringified to provide more detailed human-readable information. Some errors also should not be exposed, such as `StrongholdMutexPoisoned` (not represented in the example above), which can similarly be mapped to the catch-all variant.
[...]

PR #328, opened 2021-07-21
Serializable errors proof-of-concept

cycraig:
# Description of change
Exploring various implementations of serializable errors for actor handlers and bindings.

Incomplete and just a proof-of-concept, **do no merge**.

Serializable errors #321 

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
   "identity-did",
   "identity-diff",
   "identity-iota",
+  "flat-error",
 
   "examples",
 ]
--- a/flat-error/Cargo.toml
+++ b/flat-error/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "flat_error"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+name = "flat_error"
+proc-macro = true
+
+[dependencies]
+proc-macro2 = { version = "1.0" }
+quote = { version = "1.0" }
+serde = { version = "1.0" }
+syn = { version = "1.0", features = ["extra-traits", "parsing", "derive"] }
--- a/flat-error/src/lib.rs
+++ b/flat-error/src/lib.rs
@@ -1,0 +1,29 @@
+extern crate proc_macro;
+#[macro_use]
+extern crate syn;
+#[macro_use]
+extern crate quote;
+
+use proc_macro::TokenStream;
+use syn::DeriveInput;
+//
+// #[proc_macro_derive(FlatError)]
+// pub fn flat_error(input: TokenStream) -> TokenStream {
+//   // See https://doc.servo.org/syn/derive/struct.DeriveInput.html
+//   let ast: DeriveInput = syn::parse_macro_input!(input as DeriveInput);
+//
+//   // Build the struct
+//   let gen = gen_flat_error_struct(&ast);
+//
+//   // Return the generated impl
+//   gen.parse().unwrap()
+// }
+//
+// fn gen_flat_error_struct(ast: &syn::DeriveInput) -> quote::Tokens {
+//   // get enum name
+//   let name = &ast.ident;
+//   let data = &ast.data;
+//   quote! {
+//       // TODO
+//     }
+// }
--- a/identity-account/Cargo.toml
+++ b/identity-account/Cargo.toml
@@ -66,4 +66,5 @@ stronghold = [
 ]
 wasm = ["identity-iota/wasm"]
 async = ["identity-iota/async"]
-default = ["stronghold", "async"]
+serde-errors = []
+default = ["stronghold", "serde-errors", "async"]
--- a/identity-account/src/error.rs
+++ b/identity-account/src/error.rs
@@ -102,3 +102,117 @@ pub trait PleaseDontMakeYourOwnResult<T> {
   #[allow(clippy::wrong_self_convention)]
   fn to_result(self) -> Result<T>;
 }
+
+#[cfg(feature = "serde-errors")]
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize)]
+pub struct SerdeError {
+  pub code: ErrorCode,
+  pub description: Option<String>,
+}
+
+#[cfg(feature = "serde-errors")]
+impl SerdeError {
+  pub fn new(code: ErrorCode, description: Option<String>) -> Self {
+    Self {
+      code,
+      description,
+    }
+  }
+}
+
+#[cfg(feature = "serde-errors")]
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize)]
+pub enum ErrorCode {
+  /// [Error::CryptoError]
+  CryptoError,
+  /// [Error::CoreError]
+  CoreError,
+  /// [Error::DIDError]
+  DIDError,
+  /// [Error::CredentialError]
+  CredentialError,
+  /// [Error::IotaError]
+  IotaError,
+  /// [Error::IoError]
+  IoError,
+  /// [Error::ActorSystemError]
+  ActorSystemError,
+  /// [Error::StrongholdError]
+  StrongholdError,
+  /// [Error::StrongholdResult]
+  StrongholdResult,
+  /// [Error::InvalidResourceIndex]
+  InvalidResourceIndex,
+  /// [Error::StrongholdPasswordNotSet]
+  StrongholdPasswordNotSet,
+  /// [Error::StrongholdProcedureFailure]
+  StrongholdProcedureFailure,
+  /// [Error::StrongholdMutexPoisoned]
+  StrongholdMutexPoisoned,
+  /// [Error::SharedReadPoisoned]
+  SharedReadPoisoned,
+  /// [Error::SharedWritePoisoned]
+  SharedWritePoisoned,
+  /// [Error::GenerationOverflow]
+  GenerationOverflow,
+  /// [Error::GenerationUnderflow]
+  GenerationUnderflow,
+  /// [Error::IdentityIdOverflow]
+  IdentityIdOverflow,
+  /// [Error::IdentityIdInvalid]
+  IdentityIdInvalid,
+  /// [Error::MissingDocumentId]
+  MissingDocumentId,
+  /// [Error::KeyVaultNotFound]
+  KeyVaultNotFound,
+  /// [Error::KeyPairNotFound]
+  KeyPairNotFound,
+  /// [Error::IdentityNotFound]
+  IdentityNotFound,
+  /// [Error::EventNotFound]
+  EventNotFound,
+  /// [Error::IdentityAlreadyExists]
+  IdentityAlreadyExists,
+  /// [Error::MethodNotFound]
+  MethodNotFound,
+  /// [Error::ServiceNotFound]
+  ServiceNotFound,
+  /// [Error::CommandError]
+  CommandError,
+}
+
+#[cfg(feature = "serde-errors")]
+impl From<Error> for SerdeError {
+  fn from(error: Error) -> Self {
+    match error {
+      Error::CryptoError(inner) => Self::new(ErrorCode::CryptoError, Some(inner.to_string())),
+      Error::CoreError(inner) => Self::new(ErrorCode::CoreError, Some(inner.to_string())),
+      Error::DIDError(inner) => Self::new(ErrorCode::CoreError, Some(inner.to_string())),
+      Error::CredentialError(inner) => Self::new(ErrorCode::CredentialError, Some(inner.to_string())),
+      Error::IotaError(inner) => Self::new(ErrorCode::IotaError, Some(inner.to_string())),
+      Error::IoError(inner) => Self::new(ErrorCode::IoError, Some(inner.to_string())),
+      Error::ActorSystemError(inner) => Self::new(ErrorCode::ActorSystemError, Some(inner.to_string())),
+      Error::StrongholdError(inner) => Self::new(ErrorCode::StrongholdError, Some(inner.to_string())),
+      Error::StrongholdResult(inner) => Self::new(ErrorCode::StrongholdResult, Some(inner.to_string())),
+      Error::InvalidResourceIndex => Self::new(ErrorCode::InvalidResourceIndex, None),
+      Error::StrongholdPasswordNotSet => Self::new(ErrorCode::StrongholdPasswordNotSet, None),
+      Error::StrongholdProcedureFailure => Self::new(ErrorCode::StrongholdProcedureFailure, None),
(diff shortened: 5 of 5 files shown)


cycraig:
I think this approach---deriving `FlatEnum` for errors exported to Wasm bindings---can be considered complete as a proof-of-concept now. It's not perfect but it's useful in that it highlights some fundamental issues and things to look out for going forward.

**Advantages**:
- **informative**: the derived `FlatEnum` exposes the enum variant in JS code, compared to just a string error message currently, potentially allowing certain errors to be handled differently in user code (e.g. retrying on a timeout).
- **easily maintainable**: using `#[derive(FlatEnum)`, any updates to the error enums are propagated, so the definitions only need to be updated once in the Rust code.
- **simplifies error handling**: by implementing `Into<JsValue>` (through the `From` trait), errors in Wasm bindings can be short-circuited using the `?` operator, without an intermediate `.map_err(<function that converts error to JsValue>)` step, as a minor convenience.

**Disadvantages**:
- **stringly-typed**: the enum variant `code` is serialized to `JsValue` as a string, and so cannot be exhaustively matched in a switch/case statement in e.g. Typescript. As mentioned in this PR and issue #321, this is a fundamental problem with `wasm-bindgen` [requiring all result errors to be converted to `JsValue`](https://rustwasm.github.io/wasm-bindgen/reference/types/result.html).
- **"clunky"**: the derive approach requires the `wasm-bindgen` dependency to be added to all crates which export errors to bindings. While this is feature-gated, it's still a disadvantage until it's decided whether or not to use the `#[wasm_bindgen]` attribute throughout the `identity.rs` crates, which would require the dependency regardless.
- **wasm-specific**: these flat errors are only really useful for the Wasm bindings, as other language bindings throw errors as exceptions automatically (e.g. [`uniffi-rs`](https://mozilla.github.io/uniffi-rs/udl/errors.html)) or have their own requirements for converting errors.

I'm not a massive fan of this approach but `wasm-bindgen` only allowing `JsValue` for result errors is really the core problem preventing an ideal, non-stringly-typed solution. I can't imagine any solution from our side would be perfect until that is addressed in `wasm-bindgen` (along with the ability to export rich enums automatically using `#[wasm_bindgen]`).

An alternative approach would be to define the C-style enum errors in the wasm bindings to avoid polluting the `identity.rs` crates. That would use a declarative macro though, which might have problems enumerating the enum variants without extra dependencies.

PhilippGackstatter:
Very nice PoC!

I am primarily wondering whether JS developers would actually match on the code property. I'm very unfamiliar with JS error handling, but at least according to [this](https://stackoverflow.com/a/2826050) SO answer, the accepted default is both messy and inaccurate, i.e. catching all or nothing. So, essentially, is the trouble with the `proc-macro` and generating all those errors worth it? If our error handling approach with the codes is actually unidiomatic for JS, then it might not end up being used and most of the work might just go to waste :confused:.


It would be interesting to have a JS developer's view. @abdulmth do you think having

```json
{"code":"DecodeBitmap","description":"Failed to decode roaring bitmap: something went wrong!"}
```

thrown instead of just

```javascript
"Failed to decode roaring bitmap: something went wrong!"
```

would be a worthwhile improvement? Would JS devs actually match on the `code ` if they have the possibility to do so? Or are they generally used to match on strings if they want to catch exception `A`, but not exception `B`?

@cycraig as you mentioned in #321, the C-style enums aren't actually a fundamental binding requirement, contrary to my initial belief. If we end up going with the current approach anyway, because it has value for the JS devs, I think I would still prefer regular Rust enums (option 2) for the actor,  since that would be more idiomatic for matching. Do you think we could allow deriving both C-style enums for bindings, and Rust enums for the actor?

abdulmth:
Exceptions are usually done by throwing an `Error` object (see [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)). Expected are the `name` and `message` properties. But as Phillip said, you can throw anything in JavaScript (a string or an arbitrary object). But usually I expect an Error object to be thrown that has `name` and  `message` properties.

cycraig:
> @cycraig as you mentioned in #321, the C-style enums aren't actually a fundamental binding requirement, contrary to my initial belief. If we end up going with the current approach anyway, because it has value for the JS devs, I think I would still prefer regular Rust enums (option 2) for the actor, since that would be more idiomatic for matching. Do you think we could allow deriving both C-style enums for bindings, and Rust enums for the actor?

Indeed, the intermediate C-style are basically useless for the wasm-bindings because the enum variants can't be matched on in JavaScript except as a string and the Rust code exported to Wasm can just use the rich error enums directly. In my opinion the actor should use Rust enums that wrap any errors it needs as a string, unless it is a useful subvariant e.g.

```
#[derive(serde::Serialize, serde::Deserialize)]
enum SomeActorError {
  StrongholdPasswordNotSet,
  AccountError(string),
  ClientError(string),
  ...
}
```

Then implement `From` manually for each of the errors as-needed:
```
impl From<AccountError> for SomeActorError {
  fn from(error: AccountError) -> SomeActorError {
    match error {
      StrongholdPasswordNotSet => SomeActorError::StrongholdPasswordNotSet,
      err => SomeActorError::AccountError(err.to_string()),
    }
  }
}
```

If we need to send things like backtrace information (which seems to be a somewhat complex subject), then we could use an intermediate struct (that is still serializable) to hold it, e.g.

```
#[derive(serde::Serialize, serde::Deserialize)]
struct SerdeError {
  error: String,
  backtrace: Backtrace/String/something serializable,
}


#[derive(serde::Serialize, serde::Deserialize)]
enum SomeActorError {
  StrongholdPasswordNotSet(SerdeError),
  AccountError(SerdeError ),
  ClientError(SerdeError),
  ...
}
```

The above examples are just for demonstration. One could derive those `From` implementations to embed the variant code using a proc-macro, as done in this PR but in my opinion the actor errors and wasm binding errors should be considered completely separately.

> Exceptions are usually done by throwing an `Error` object (see [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error))

This changes things a bit. The current implementation, only returns arbitrary objects/strings, never a proper JavaScript `Error`. I.e. when using try-catch for errors in the following code example, `e instanceof Error` will always return `false` for anything we return, even with the current changes in this PR:

```
let messageId;
try {
    messageId = await client.publishDocument(doc.toJSON());
} catch (e) {
    if (e instanceof Error) {
        console.log(`e is an Error! ${e}`)
    } else {
        console.log(`not an Error... ${e}`)
    }
}
```
This will always print "not an Error..."

We can convert the error we return into a `js_sys::Error`, as suggested by https://github.com/rustwasm/wasm-bindgen/issues/1735#issuecomment-524348781. This would make `e instanceof Error` return `true`. However, this approach may not be ideal since the only two fields of `Error` (`name` and `message`) are both simple strings, so while we could pass the serialised JSON error as the `message` I think it may need to be parsed as JSON inside JavaScript to access custom fields like `code` and `description`?

Based on the points raised already, I'll work on another proof-of-concept. Essentially it will convert the Rust errors directly to some `WasmError` struct without C-style enums as an intermediary (since those are useless and unused).

PhilippGackstatter:
> I think it may need to be parsed as JSON inside JavaScript to access custom fields like `code` and `description`?

I'm not sure I've understood yet why we can't map the `code` into `name` and `description` into `message`? If `code` is the name of the enum variant, you can match on that even if it's a string.

I've taken a quick look at [aws-sdk-js](https://github.com/aws/aws-sdk-js) to see their approach, and it seems they frequently use `throw new Error("some error description")`. If we overwrite `error.name` with the name of the variant, then devs can use that to implement branching logic. I understand that we have not the same level of expressiveness as in Rust, because of nesting, but I thought we had accepted this as a trade-off. This seems like the best and simplest thing we can do with the JS `Error` type. If someone wants branching logic within a variant, they can still do some (ugly of course) `error.name.contains("error string i want to match")`, which is probably what most JS devs need to do anyway with most exceptions.

cycraig:
> I'm not sure I've understood yet why we can't map the code into name and description into message? If code is the name of the enum variant, you can match on that even if it's a string.

We absolutely can do that and that will likely be the final implementation. I'm just checking whether we can do any better than that, by e.g. adding more fields to the returned error apart from the description, such as stacktraces to aid in debugging.

On a related note: I have to correct my previous statement that we can't use `#[wasm_bindgen]` in the proc-macro, we absolutely can (as long as it's placed above/before the `#[derive(...)]` attributes).

This enables use to generate something like the following:

```rust
#[wasm_bindgen]
#[derive(Clone, Debug)]
pub struct WasmError {
  code: WasmErrorCode,
  description: String,
}

#[wasm_bindgen]
impl WasmError {
  #[wasm_bindgen(constructor)]
  pub fn new(code: WasmErrorCode, description: String) -> Self {
    Self {
      code,
      description,
    }
  }

  #[wasm_bindgen(getter)]
  pub fn code(&self) -> SerdeErrorCode {
    self.code
  }

  #[wasm_bindgen(getter)]
  pub fn description(&self) -> String {
    self.description.clone()
  }
}

#[wasm_bindgen]
#[derive(Clone, Copy, Debug)]
pub enum WasmErrorCode {
  CryptoError,
  ...
}
```

In Javascript, this would enable the user to check the type of code too:

```js
let messageId;
try {
    messageId = await client.publishDocument(doc.toJSON());
} catch (e) {
    if (e instanceof Error) {
        console.log(`e is an Error! ${e}`)          // WasmError is still NOT a Javascript Error
    } else if (e instanceof WasmError) {            // WasmError is exported by identity-wasm
        if (e.code === WasmErrorCode.CryptoError) { // WasmErrorCode enum variants can be compared directly
            console.log(`CryptoError: ${e}`)
        } else {
            console.log(`Unknown WasmError: ${e}`)
        }
    } else {
        console.log(`not an Error: ${e}`)
    }
}
```

**However when I actually tried doing that for all the different error enums identity.rs exports, the linker fails due to "duplicate symbols"**:

```
 = note: rust-lld: error: duplicate symbol: flaterror_new
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_credential-61734c946a494da3.rlib(identity_credential-61734c946a494da3.identity_credential.23b4pm7p-cgu.1.rcgu.o)
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_did-9b8e16a2407ef1de.rlib(identity_did-9b8e16a2407ef1de.identity_did.9s0pxrfz-cgu.0.rcgu.o)

          rust-lld: error: duplicate symbol: __wbindgen_describe_flaterror_new
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_credential-61734c946a494da3.rlib(identity_credential-61734c946a494da3.identity_credential.23b4pm7p-cgu.1.rcgu.o)
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_did-9b8e16a2407ef1de.rlib(identity_did-9b8e16a2407ef1de.identity_did.9s0pxrfz-cgu.0.rcgu.o)

          rust-lld: error: duplicate symbol: flaterror_code
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_credential-61734c946a494da3.rlib(identity_credential-61734c946a494da3.identity_credential.23b4pm7p-cgu.1.rcgu.o)
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_did-9b8e16a2407ef1de.rlib(identity_did-9b8e16a2407ef1de.identity_did.9s0pxrfz-cgu.0.rcgu.o)

          rust-lld: error: duplicate symbol: __wbindgen_describe_flaterror_code
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_credential-61734c946a494da3.rlib(identity_credential-61734c946a494da3.identity_credential.23b4pm7p-cgu.1.rcgu.o)
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_did-9b8e16a2407ef1de.rlib(identity_did-9b8e16a2407ef1de.identity_did.9s0pxrfz-cgu.0.rcgu.o)

          rust-lld: error: duplicate symbol: flaterror_description
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_credential-61734c946a494da3.rlib(identity_credential-61734c946a494da3.identity_credential.23b4pm7p-cgu.1.rcgu.o)
          >>> defined in C:\Work\iota\identity.rs\bindings\wasm\target\wasm32-unknown-unknown\release\deps\libidentity_did-9b8e16a2407ef1de.rlib(identity_did-9b8e16a2407ef1de.identity_did.9s0pxrfz-cgu.0.rcgu.o)

[and so on...]
```

So while we could try and make the derived struct names hygienic (it's really a problem with the code generated by wasm_bindgen that's unhygienic and conflicting in the same namespace), it doesn't seem worth it.

> I've taken a quick look at aws-sdk-js to see their approach, and it seems they frequently use throw new Error("some error description").

With the norm being generic error strings in Javascript, I think simply returning a Javascript Error with `name = code` and `message = description` is "fine". I've been spoiled by the enums and error handling in Rust but the reality is probably that it's not especially useful nor idiomatic in Javascript.

cycraig:
I've updated the code with proper usage of `#[wasm_bindgen]` which renames the exported generated types from e.g. `WasmIotaError` back to `IotaError` in Javascript code, so it elides the conversion on their side.

This means proper type comparisons may now be used as per the original proposal in the issue (instead of string comparisons). 
E.g.

```js
let messageId;
try {
    messageId = await client.publishDocument(doc.toJSON());
} catch (e) {
    if (e instanceof Error) {                 // exported errors are still NOT subclasses of Javascript Errors
        console.log(`e is an Error! ${e}`)
    } else if (e instanceof CoreError) {      // CoreError is exported by identity-wasm
        switch (e.code) {                     // CoreErrorCode enum variants can be compared directly
            case CoreErrorCode.Crypto: {  
                console.log(`CryptoError: ${e}`); break;
            }
            case CoreErrorCode.InvalidUrl: {
                console.log(`InvalidUrl: ${e}`); break;
            }
            default: {
                console.log(`Unhandled CoreError: ${e}`)
            }
        }
    } else {
        console.log(`not an Error: ${e}`)
    }
}
```

Most of the original disadvantages of this approach have been addressed now (at the cost of re-naming the errors to avoid namespace conflicts and including `wasm-bindgen` as a feature-gated dependency in most of the identity crates) so it's not too bad anymore. It's perhaps overkill and unlikely to be used though.

I'll still push the alternative implementation based on `js_sys::Error` soon.

cycraig:
Closed in favour of https://github.com/iotaledger/identity.rs/pull/344

PR #329, opened 2021-07-22
Impl type that mimics a thread-safe `Cell`

PhilippGackstatter:
# Description of change

We received an error report:

```
`std::cell::Cell<bool>` cannot be shared between threads safely
...
= note: required because it appears within the type `identity::crypto::Signature`
```
which should be fixed by this PR. It essentially implements a type that mimics a thread-safe `Cell`.

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/identity-core/src/crypto/signature/signature.rs
+++ b/identity-core/src/crypto/signature/signature.rs
@@ -1,7 +1,6 @@
 // Copyright 2020-2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use core::cell::Cell;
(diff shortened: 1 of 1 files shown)


cycraig (approved):
All checks pass so looks good. 

I have some concerns regarding using `Ordering::Relaxed` in general, since we might only discover any problems with it on architectures with weaker ordering guarantees like ARM. The usage here as just a flag _seems_ fine internally though. 

In my opinion we should default to `Ordering::SeqCst` everywhere and only change it to a weaker ordering in performance-critical code after careful review.

PR #344, opened 2021-08-02
Improve Wasm error handling

cycraig:
# Description of change
Adds `WasmError` to enable idiomatic Javascript error handling with `js_sys::Error`, which includes enum variant names for internal errors. E.g.

```
let receipt;
try {
    receipt= await client.publishDocument(doc.toJSON());
} catch (e) {
    if (e instanceof Error) {    // <--- exported errors are now instances of Javascript Error
        console.log(`e is an Error! Name: ${e.name}, message: ${e.message}`)
    } else {
        console.log(`not an Error: ${e}`)
    }
}
```

This is different from the proof-of-concept in PR #328, which uses a procedural macro to derive a new struct and enum from internal types. This PR instead defines a single `WasmError` struct as an intermediate representation that can be converted to `js_sys::Error`. 

The advantage of this approach is that we can convert errors from dependencies into `WasmError` too, which is not possible with a `derive` proc-macro. It is also self-contained in the wasm-bindings, so no need to add `wasm-bindgen` to all the projects - though the internal errors do require a new dependency,  [`strum`](https://github.com/Peternator7/strum), to convert variants to strings. The disadvantage is that the error variant is represented as a string, so typed matching is not possible in Javascript/Typescript, but that doesn't seem necessary as per previous discussions in #328.

See issue #321 

## Type of change

- [ ] Documentation Fix

Added a new `wasm_bindgen_test` for `WasmError`, run with `wasm-pack test --node`
Wasm examples all run successfully locally.

Manually checked that throwing an error from `identity.rs` matches the `js_sys::Error` type in Node with the name and message fields.

## Change checklist


cycraig:
The CI errors are in `libjose` (not any files touched by this PR) due to new `clippy` lints and `rustfmt` updates in the latest stable release (2021-07-29), https://github.com/rust-lang/rust/blob/master/RELEASES.md

PhilippGackstatter (approved):
I really like this approach. It's very elegant in how minimal it is, but still produces an idiomatic JavaScript error, which can be used to implement branching logic.

It has the previously discussed trade-off that

```rust
return Err(wasm_error(identity::comm::Error::CoreError(identity::core::Error::InvalidSigLength(4, 5))));
```

would produce:

```
name: CoreError
message: Invalid Signature Length. Received 4, Expected 5
```

Again, I think it's fine and I still can't think of a more fine-grained way to do it given the JS `Error` limitations. If someone wants to specifically catch an `Invalid Signature Length` they can still do `message.contains("Invalid Signature Length")`.

cycraig:
> Again, I think it's fine and I still can't think of a more fine-grained way to do it given the JS `Error` limitations. If someone wants to specifically catch an `Invalid Signature Length` they can still do `message.contains("Invalid Signature Length")`.

The loss of context in the error chain is unfortunate, PR #328 suffers from the same problem. Perhaps the chain of nested enums can be serialized in the name string? E.g. 

```rust
return Err(wasm_error(identity::comm::Error::CoreError(identity::core::Error::InvalidSigLength(4, 5))));
```
would yield something like:
```json
name: "CommError::CoreError::InvalidSigLength"
```

Unfortunately not all error enums support converting the variant name to a string---particularly those thrown from dependencies---so a lot of manual conversions would likely be required, increasing the maintenance cost.

Another alternative would be to use debug format rather than the `Display/ToString` implementation? E.g.

```rust
format!("{:?}", identity::comm::Error::CoreError(identity::core::Error::InvalidSigLength(4, 5)));
```
yields:
```json
"CoreError(InvalidSigLength(4, 5))"
```

Just `"CoreError"` is probably sufficient for this PR as it can be changed later but any suggestions are welcome.

l1h3r (approved):
Nice approach 💯 

I think this is a pretty reasonable start, backtrace support in `wasm-bindgen` will further improve this.

I do like the idea of the `"Identity::"` prefix to add some context, although I'm not sure if this is a good JS paradigm 🤔

PhilippGackstatter:
> I do like the idea of the `"Identity::"` prefix to add some context, although I'm not sure if this is a good JS paradigm thinking

I totally agree with that. I'd probably leave that out for now.

PhilippGackstatter:
> Another alternative would be to use debug format rather than `Display/ToString`? E.g.

Interesting idea, but the main problem is going to be the [instability](https://doc.rust-lang.org/std/fmt/trait.Debug.html#stability) of this output.

Overall, I would merge this as is and see how it goes / what the feedback is from bindings users.

Issue #325, opened 2021-07-20
[Task] Improve Account API by internalizing commands


JelleMillenaar:
## Description

Small refactor on the account API in order to make updating DID Documents from a two step into a one step process. Currently the Account API allows developers to generate update commands that they have to manage and pass on into the `update_identity` function. These commands can be stored inside the account and can be either published immediately or queued to be published in a single call of `update_identity`. If more commands are queued they are published in the order they are added. Later, we can add additional optimization code to combine multiple commands into a single DID update. 

## Motivation

Reduce the complexity of the account even further.

## Resources

[Example](https://github.com/iotaledger/identity.rs/blob/dev/examples/account/methods.rs#L29)

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Internalize Commands
- [ ] Update `update_identity` function to no longer take a command parameter.
- [ ] Update all account examples
- [ ] Rename `document` in account examples to `DID`
- [ ] Enable queuing multiple updates before publishing - PR #377.
- [x] Add the ability to create an identity with an existing keypair to account - PR #352

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

JelleMillenaar:
Added 'Add the ability to create an identity with an existing keypair to account' to the task as it is a minor account API change.

l1h3r:
@JelleMillenaar what is the intended benefit of reusing an existing keypair here? is this just for convenience or space-saving? 

Stronghold doesn't support copying keys (maybe this *could* be added) and sharing keys across multiple identities would require a fair amount of internal changes. If you wanted to create a "child" identity owned by a "parent" identity wouldn't the `controller` field be a better fit once we have proper support for it?

PhilippGackstatter:
In regards to the simplification & optimization of identity updates, I think we have (at least) these three variants to consider.

[...]

PR #352, opened 2021-08-10
Enable the creation of an identity from a `KeyPair` in the `Account`

PhilippGackstatter:
# Description of change

The IOTA Streams team requested the ability to create an identity from an existing `KeyPair`. Due to the pre-existing `KeyType` field on the `IdentityCreate` type, only a `SecretKey` was added to that type, to avoid adding the `KeyType` field a second time (in the `KeyPair` itself) and the possibly resulting confusion. This means an identity can only be created from an `Ed25519` key, not a `MerkleKeyCollection`. 

As an example, developers can create an identity from an `ed25519` `KeyPair`, by doing:

```rust
let keypair: KeyPair = KeyPair::new_ed25519()?;

let id_create = IdentityCreate::new()
    .key_type(KeyType::Ed25519)
    .method_secret(MethodSecret::Ed25519(keypair.secret().clone()));

account.create_identity(id_create).await?;
```

Alternatively, it's also possible to add a new method with a custom secret key. Here both `Ed25519` keys as well as `MerkleKeyCollection`s are possible, which can be set through the `MethodSecret` wrapper enum.

```rust
let keypair: KeyPair = KeyPair::new_ed25519()?;

let command: Command = Command::create_method()
    .type_(MethodType::Ed25519VerificationKey2018)
    .fragment("key-1")
    .method_secret(MethodSecret::Ed25519(keypair.secret().clone()))
    .finish()?;

account.update_identity(did, command).await?;
```

Note: A stronghold upgrade was necessary, to include a [bugfix](https://github.com/iotaledger/stronghold.rs/pull/253).

Part of #325 

## Type of change

- [ ] Documentation Fix

Various tests were added, including:
- A stronghold test, `test_store_secret_key` to make sure that inserting a secret key into stronghold works
- Two account tests
    - `test_create_identity_from_secret_key`
    - `test_create_method_from_secret_key`

I modified the `account_stronghold` example with the above code to test the change, but no automated test was added. It seems like we don't have high-level account-stronghold tests in the first place.

## Change checklist


DyrellC:
In addition to what you've added already here, is it possible to easily add the ability to use an existing keypair not as the base for the account (`CreateIdentity` command), but as the base for a method (`CreateMethod`)? This is important for allowing existing account owners to add a new verification method to longer lasting identities. For example I made this for testing purposes on my end, i'm sure there's a better way to go about it though: 
https://github.com/DyrellC/identity.rs/commit/9427a528196c8e3b6fd5e86b168644ae3253da2a

PhilippGackstatter:
> is it possible to easily add the ability to use an existing keypair as the base for a method (`CreateMethod`)?

Sounds like a reasonable addition to me - we can add this. And thanks for providing the code!

PhilippGackstatter:
@DyrellC The PR is done now, feel free to take another look and confirm that this is what you expected.

identity-account/src/identity/identity_create.rs
```diff
@@ -47,6 +49,16 @@ impl IdentityCreate {
...
+
+  /// Sets the [SecretKey] for the Identity creation.
+  /// This needs to be the correct number of bytes for the respective [KeyType].
+  ///
+  /// For [KeyType::Ed25519] the key needs to be 32 bytes in size.
+  #[must_use]
+  pub fn secret_key(mut self, value: SecretKey) -> Self {
+    self.secret_key = Some(value);
+    self
+  }
```
cycraig, line 61: `SecretKey` seems specific to `KeyType::Ed25519`, which will be a problem since other methods may use different representations, e.g. Merkle keys use `KeyCollection` / `Box<[SecretKey]>`. We could use something like an enum with optional associated data as the `key_type` instead.

Since the `Account` lacks support for Merkle key collections in general this is fine for now, just noting that it will need a breaking change in the future.

identity-account/src/storage/memstore.rs
```diff
@@ -109,6 +113,32 @@ impl Storage for MemStore {
     }
   }
 
+  async fn key_insert(&self, id: IdentityId, location: &KeyLocation, secret_key: SecretKey) -> Result<PublicKey> {
+    let mut vaults: RwLockWriteGuard<'_, _> = self.vaults.write()?;
+    let vault: &mut MemVault = vaults.entry(id).or_default();
+
+    match location.method() {
+      MethodType::Ed25519VerificationKey2018 => {
+        let mut secret_key_bytes: [u8; 32] = secret_key.as_ref().try_into().expect("expected a slice of 32 bytes");
```
cycraig, line 122: Nit: return a proper `Error` instead of panicking.

E.g.
```
    let mut secret_key_bytes: [u8; 32] = <[u8; 32]>::try_from(secret_key.as_ref())
      .map_err(|err| Error::InvalidSecretKey(format!("expected a slice of 32 bytes - {}", err)))?;
```
This would require a new error type or one could wrap the `TryFromSliceError` directly. Replacing `try_into` with `try_from` is optional, I just prefer the latter to lookup which implementation it's using more easily.

identity-account/src/storage/stronghold.rs
```diff
@@ -98,6 +99,19 @@ impl Storage for Stronghold {
     Ok(public)
   }
 
+  async fn key_insert(&self, id: IdentityId, location: &KeyLocation, secret_key: SecretKey) -> Result<PublicKey> {
+    let vault = self.vault(id);
+
+    vault
+      .insert(location_skey(location), secret_key.as_ref(), default_hint(), &[])
```
cycraig, line 106: Do we always want a blank `RecordHint`? Would setting it to the key type , e.g. `"Ed25519"`, be useful at all or does it introduce a security risk, or is that information essentially in the location anyway? 

Just asking out of interest, not suggesting a change.

identity-account/tests/commands.rs
```diff
@@ -114,6 +117,29 @@ async fn test_create_identity_already_exists() -> Result<()> {
   Ok(())
 }
 
+#[tokio::test]
+async fn test_create_identity_from_secret_key() -> Result<()> {
```
cycraig, line 121: We may want to add equivalent tests to ensure we error on secret keys with invalid lengths, e.g. `let secret_key = [0_u8; 33].to_vec().into();`

identity-account/tests/commands.rs
```diff
@@ -114,6 +117,29 @@ async fn test_create_identity_already_exists() -> Result<()> {
...
+
+  account.create_identity(id_create.clone()).await?;
+  account2.create_identity(id_create).await?;
+
+  let ident = account.find_identity(identity).await.unwrap().unwrap();
+  let ident2 = account.find_identity(identity).await.unwrap().unwrap();
+
+  // The same secret key should result in the same did
+  assert_eq!(ident.identity().did(), ident2.identity().did());
+
```
cycraig, line 139: This should also test that the generated authentication methods match:

E.g.
```
  assert_eq!(ident.identity().authentication()?, ident2.identity().authentication()?);
```

cycraig (approved):
Looks good to me. I don't see anything obviously wrong but could have overlooked something as I have little experience with Stronghold.

See minor comments below.

DyrellC:
> @DyrellC The PR is done now, feel free to take another look and confirm that this is what you expected.

Yup looks good to me! Thank you!

identity-account/tests/commands.rs
PhilippGackstatter: It's not going to error on that. `Vault::insert` takes an arbitrary vector of bytes. I'll add a check in the account.

identity-account/src/events/command.rs
```diff
@@ -21,20 +23,23 @@ use crate::identity::TinyService;
...
 
 // Supported authentication method types.
 const AUTH_TYPES: &[MethodType] = &[MethodType::Ed25519VerificationKey2018];
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 pub enum Command {
   CreateIdentity {
     network: Option<String>,
+    secret_key: Option<SecretKey>,
```
cycraig, line 35: Can we also make this a `MethodSecret`?

PhilippGackstatter: No, we don't have to. Creating an identity from a KeyCollection does not need to be allowed according to @JelleMillenaar. It would also be difficult to add this, since the KeyCollection takes a parameter to create it. So when the user does not give a KeyCollection, it's unclear what the parameter should be.

cycraig: Right, we're unlikely to ever use `KeyCollection` for identity creation, I was just thinking in terms of forward compatibility with any other methods we might add and for consistency, since extending the `MethodSecret` enum would be non-breaking.

PR #360, opened 2021-08-18
Improve `Network` type to support private tangles

### LINKED ISSUE #355: Improve Support for Private Tangle
## Description
Creating a configuration for Private Tangles should be improved and follow similar steps as configuring Mainnet or Testnet.

## Motivation
A `Network` object can't be configured for networks other than Mainnet or Testnet. 

## To-do list
- [ ] Improve the `Network` class to be configured for private tangles.
- [ ] Add an example for a private tangle configuration.

## Change checklist

PhilippGackstatter:
# Description of change

Currently, the `Network` enum consists only of two types: `test` and `main`. To support private tangle networks, another variant was added, called `Other`, which contains the name of the network.

An unfortunate side-effect of this change is that the `Network` type can no longer be `Copy`, which introduces a bunch of `.clone()` invocations in various places.

One thing I'd like to discuss is the change in `Network::from_name`. We said we wanted to prevent empty strings from being valid network names. That changes the signature of that method to be a `Result`, which obviously means callers need to handle that every single time they create a network. I'm no longer sure that tiny check justifies this downside in usability, and I would actually like to remove it again. If we actually wanted to introduce more "safety" for network names, we would also need to check for all kinds of unicode white space in the name, and disallow that, too. That's not a very cheap check and not sure if it's worth it overall.

Another change that results from the added `Other` variant, is that there are no longer an obvious `default_node_url` or `explorer_url`, which means those methods also return `Result`s.

fixes #355 

## Type of change

    - Signatures of public methods changed, in particular of `Network` methods.
- [ ] Documentation Fix

I ran a [one-click-tangle](https://github.com/iotaledger/one-click-tangle/tree/chrysalis/hornet-private-net) and then ran the new examples for Rust and JS. These then produce a document with a DID such as:

```
did:iota:private-tangle:4NcFXKH6B3PvDyZrFQz2Fowpxmw7QN8e2Ky14Kk7Qwan
```

I don't think there's an easy way to add an automated test for these changes.

## Change checklist


bindings/wasm/examples/browser/private_tangle.js
```diff
@@ -0,0 +1,48 @@
...
+    // Sign the DID Document with the generated key.
+    doc.sign(key);
+
+    // Create a client configuration and set the custom network.
+    const config = new Config();
+    config.setNetwork(network);
+
+    // This URL points to the REST API of the locally running hornet node.
+    config.setNode("http://127.0.0.1:14265/");
+
```
abdulmth, line 36: permanode should be set too?

PhilippGackstatter: But for the example it's not required and I don't think there's a one-click permanode available, right?

abdulmth: afaik resolving a DID requires a permanode. 
You can run your own permanode separately [chronicle](https://github.com/iotaledger/chronicle.rs). And there's an issue for adding it to one-click-tangle [here](https://github.com/iotaledger/one-click-tangle/issues/46). We still didn't test that yet. Eike @eike-hass had some progress regarding that.

identity-iota/src/error.rs
```diff
@@ -51,4 +51,8 @@ pub enum Error {
   CannotRemoveAuthMethod,
   #[error("Cannot Revoke Verification Method")]
   CannotRevokeMethod,
+  #[error("No Client Nodes Provided")]
+  NoClientNodesProvided,
+  #[error("No Explorer for Private Tangles")]
+  NoExplorerForPrivateTangles,
```
PhilippGackstatter, line 57: It's actually possible to deploy an [explorer for private tangles](https://github.com/iotaledger/one-click-tangle/blob/chrysalis/explorer/README.md), so I can add the option to set an explicit explorer url on the `Network`.

identity-iota/src/tangle/network.rs
```diff
@@ -16,22 +19,29 @@ lazy_static! {
...
+  pub fn from_name(string: &str) -> Result<Self> {
     match string {
-      TEST_NETWORK_NAME => Self::Testnet,
-      _ => Self::Mainnet,
+      "" => Err(Error::InvalidDIDNetwork),
+      TEST_NETWORK_NAME => Ok(Self::Testnet),
+      MAIN_NETWORK_NAME => Ok(Self::Mainnet),
+      other => Ok(Self::Other(other.to_owned())),
     }
   }
```
cycraig, line 46: We should probably enforce that the network name adheres to the IOTA DID Spec:
https://github.com/iotaledger/identity.rs/blob/99049c0e4a64a7babf224298d5d38bc243421965/documentation/docs/specs/iota_did_method_spec.md

That is, completely lowercase, only chars in `{0-9,a-b}`, and only up to 6 chars (is that arbitrary or a hard limit?). We can maybe transform it in `from_name` to lowercase, strip out spaces etc. but some basic validation to prevent DIDs being generated with invalid network tags seems necessary at the very least.

identity-iota/src/tangle/network.rs
```diff
@@ -46,35 +56,37 @@ impl Network {
...
-  pub const fn as_str(self) -> &'static str {
+  pub fn as_str(&self) -> Cow<'static, str> {
     match self {
-      Self::Mainnet => MAIN_NETWORK_NAME,
-      Self::Testnet => TEST_NETWORK_NAME,
+      Self::Mainnet => Cow::Borrowed(MAIN_NETWORK_NAME),
+      Self::Testnet => Cow::Borrowed(TEST_NETWORK_NAME),
+      Self::Other(network) => Cow::Owned(network.clone()),
     }
   }
```
cycraig, line 91: `as_str()` should probably be renamed `.name()` or `.network_name()` to be more clear, there are a lot of properties on a `Network`.

identity-iota/src/did/url/iota_did.rs
```diff
@@ -248,7 +248,8 @@ impl IotaDID {
 
   /// Returns the Tangle `network` of the `DID`.
   pub fn network(&self) -> Network {
-    Network::from_name(self.network_str())
+    // SAFETY: `network_str` never returns an empty string.
+    Network::from_name(self.network_str()).unwrap()
   }
```
cycraig, line 253: This should probably return a ` Result` too, not all non-empty network strings will be valid networks (according to our spec).

bindings/wasm/examples/browser/main.js
```diff
@@ -44,3 +45,8 @@ document
 document
     .querySelector("#merkle_key_btn")
     .addEventListener("click", () => merkleKey(clientConfig));
+
+//handle private tangle DID creation on click event
+document
+    .querySelector("#private_tangle_btn")
+    .addEventListener("click", () => createIdentityPrivateTangle());
```
cycraig, line 52: What do we actually expect to happen when a user runs a private tangle example? There will always be an error in the browser console but it's unclear and not printed to the text box on the screen.

Maybe the browser example can have a text input for the private network name and node address similar to the resolve DID example?

bindings/wasm/examples/browser/private_tangle.js
```diff
@@ -0,0 +1,48 @@
...
+    // Create a client instance from the configuration to publish messages to the Tangle.
+    const client = Client.fromConfig(config);
+
+    // Publish the Identity to the IOTA Network, this may take a few seconds to complete Proof-of-Work.
+    const receipt = await client.publishDocument(doc.toJSON());
+
```
cycraig, line 42: Maybe throw a custom error message here that's testable and can be used with `logToScreen` too?

bindings/wasm/examples/browser/private_tangle.js
```diff
@@ -0,0 +1,48 @@
...
+    logToScreen,
+} from "./utils.js";
+
+/**
+    This example shows how a DID document can be created on a private tangle.
+    It can be run together with a local hornet node.
+    Refer to https://github.com/iotaledger/one-click-tangle/tree/chrysalis/hornet-private-net
+    for setup instructions.
+**/
+export async function createIdentityPrivateTangle(log = true) {
```
cycraig, line 16: Can we add the new examples to the examples test suite (`wasm/bindings/examples/node/test.js` for Nodejs and `wasm/bindings/cypress/integration/browser_test.js`)? I don't expect to start up a private tangle in CI (although that's possible) but we could at least throw a predefined error when `publish_document` fails and test that it is thrown in the tests (so no other e.g. syntax errors creep in).

bindings/wasm/examples/browser/private_tangle.js
cycraig: Since the example is just publishing sure but it's probably necessary if you're going to use a private tangle. Maybe it's worth at least mentioning it in a comment that it is required for resolution when using a custom network and "this is how you would set it".

cycraig (changes requested):
I think the overall approach is fine. 

The alternative would be to make the `Network` a proper struct so developers can set the `default_node_url` and `explorer_url` manually, but that requires a larger refactor to avoid cloning the strings around every time `Network` is used somewhere. It would also overlap a bit too much with the `Config` and might be confusing to new users where one should set the `node_url` for example.

The `private_tangle` examples not being testable is not ideal. Either we can spin up a private Tangle in GitHub Actions (not impossible) or we throw a predefined error message when `publish_document` fails and catch it, to ensure there are no e.g. syntax errors. I do think the new examples should be added to the tests either way.

I believe the custom network names need to be validated according to the IOTA DID Spec as they form part of generated DID strings (when not published to "main", which is the default and excluded). E.g. "did:iota:Private-tangle:H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV" is not valid. Unless I'm mistaken this is the main thing that needs to change, the rest is debatable.

examples/account/private_tangle.rs
```diff
@@ -0,0 +1,64 @@
...
+use identity::iota::Network;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+  pretty_env_logger::init();
+
+  let network_name = "private-tangle";
+
+  // Unwrap is fine since we provided a non-empty string.
+  let network = Network::from_name(network_name).unwrap();
```
cycraig, line 27: Can we just use `?` instead of `unwrap()` to return the error for brevity?

examples/account/private_tangle.rs
```diff
@@ -0,0 +1,64 @@
...
+
+  // Create a new Account with the default configuration
+  let account: Account = Account::builder()
+    // Configure a client for the private network.
+    // Also set the URL that points to the REST API
+    // of the locally running hornet node.
+    .client(network, |builder| {
+      // unwrap is safe, we provided a valid node URL
+      builder.node("http://127.0.0.1:14265/").unwrap()
```
cycraig, line 36: Can we just use `?` instead of `unwrap()` to return the error for brevity?

examples/low-level-api/private_tangle.rs
```diff
@@ -0,0 +1,55 @@
...
+use identity::iota::Receipt;
+use identity::prelude::*;
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+  // This is the `networkID` defined in hornet's node configuration.
+  let network_name = "private-tangle";
+
+  // Unwrap is fine since we provided a non-empty string.
+  let network = Network::from_name(network_name).unwrap();
```
cycraig, line 24: Can we just use `?` instead of `unwrap()` to return the error for brevity?

cycraig (commented):
Nit: I prefer returning errors with `?` in examples over `unwrap()`. It avoids the comments too.

identity-iota/src/tangle/network.rs
PhilippGackstatter: You raise a good point!

The name of the `Network`, however, isn't necessarily the same as the iota-network-id . It's possible to publish a `did:iota:arbitrary_string:...` to the [testnet](https://explorer.iota.org/testnet/message/f290987610dbbb251193b103cbf125befb5006eda89688bdc89a1fd7dff94683). This is because this method takes an arbitrary string, not a `Network` type: `IotaDocument::from_keypair_with_network(keypair, "arbitrary_string");` So the spec isn't currently enforced in the low-level API. In current `dev`, using the Account API with `IdentityCreate::new().network("arbitrary")` also lets users publish a `did:iota:arbitrary:...` to the [mainnet](https://explorer.iota.org/mainnet/message/941f463dea07dcb7b145ee0229dca5f6ac091a8fca0398413ea4d23a946a93ab). 

Which raises the question, why isn't this enforced currently, and do we want to change that? I guess so. In that case, we shouldn't limit the name of the `Network` to 6 characters, or even just `{0-9,a-b}`. If someone has a `private-tangle` running, they probably don't want to change the name of the network just to publish their did. Then the reasonable thing to do, is to create a separate type that enforces the spec.

bindings/wasm/examples/browser/private_tangle.js
JelleMillenaar: Permanodes are not needed, but they guarantee that an existing identity is found.

identity-iota/src/tangle/network.rs
JelleMillenaar: What @cycraig said is correct. It should enforce the method spec with 6 chars, lowercase and `{0-9,a-z}`. The "name of the network" for the identity framework indeed should mean the iota-network-id, any other names are arbitrary and meaningless to the framework. Btw, the 6-char limit is arbitrary, but something had to be chosen.

examples/account/private_tangle.rs
PhilippGackstatter: I guess that's fine, too. Since we can assert the error never occurs, I figured using `?` sends a weird message to the reader. Overall, it does make examples nicer to read though, so I'll change it.

examples/account/private_tangle.rs
PhilippGackstatter: I changed the other occurrences of `unwrap` to use `?`. Since this one is inside a closure, we can't use `?`.

PhilippGackstatter:
Since the last review, these were the significant changes:

- Added the examples to the jest and cypress tests
- Since it's possible to have private explorers, I've added the ability to set the explorer url. When retrieving the url, I've made it an error if no explorer url was set prior to that call. Most of the times we'll be working with the mainnet and testnet, where explorer urls are always available. Thus making `network.explorer_url()` return an `Option` would be cumbersome to work with, when most of the time it'd be a `Some` anyway. 

Other than that I've incorporated the requested changes. Please take another look @cycraig, thanks!

bindings/wasm/examples/node/private_tangle.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+    This example shows how a DID document can be created on a private tangle.
+    It can be run together with a local hornet node.
+    Refer to https://github.com/iotaledger/one-click-tangle/tree/chrysalis/hornet-private-net
+    for setup instructions.
+**/
+async function createIdentityPrivateTangle() {
+    // This is an arbitrarily defined network name
+    const network = Network.from_name("atoi");
```
cycraig, line 14: Standardise example network names to `"custom"`. Even though `"atoi"` is an anagram of `"iota"`, it could be confusing since it's also the name of a well-known function.

```suggestion
    // This is an arbitrarily defined network name
    const network = Network.from_name("custom");
```

examples/account/private_tangle.rs
```diff
@@ -0,0 +1,64 @@
...
+use identity::iota::Network;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+  pretty_env_logger::init();
+
+  let network_name = "private-tangle";
+
+  // Unwrap is fine since we provided a non-empty string.
+  let network = Network::from_name(network_name)?;
```
cycraig: This example fails network name validation. Standardise example network names to `"custom"`.

```suggestion
  // This is an arbitrarily defined network name
  let network_name = "custom";
  let network = Network::from_name(network_name )?;
```

cycraig: This fails network name validation. Standardise example network names to `"custom"`.

```suggestion
  // This is an arbitrarily defined network name
  let network_name = "custom";
  let network = Network::from_name(network_name)?;
```

examples/low-level-api/private_tangle.rs
```diff
@@ -0,0 +1,55 @@
...
+use identity::iota::Receipt;
+use identity::prelude::*;
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+  // This is the `networkID` defined in hornet's node configuration.
+  let network_name = "atoi";
+
+  // Unwrap is fine since we provided a non-empty string.
+  let network = Network::from_name(network_name)?;
```
cycraig: Standardise example network names to `"custom"`.

```suggestion
  // This is an arbitrarily defined network name
  let network_name = "custom";
  let network = Network::from_name(network_name)?;
```

bindings/wasm/examples/browser/private_tangle.js
```diff
@@ -0,0 +1,64 @@
...
+    let restURL
+    let networkName
+
+    if (inBrowser) {
+        // Get the required parameters from the input fields
+        restURL = document.querySelector("#create-private-rest-url").value;
+        networkName = document.querySelector("#create-private-network-name").value;
+    } else {
+        restURL = "http://127.0.0.1:14265/";
+        networkName = "priv";
```
cycraig, line 29: Standardise example network names to `"custom"`.

```suggestion
        networkName = "custom";
```

examples/account/private_tangle.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[tokio::main]
+async fn main() -> Result<()> {
+  pretty_env_logger::init();
+
+  // This is an arbitrarily defined network name
+  let network_name = "atoi";
+
+  // Unwrap is fine since we provided a non-empty string.
+  let network = Network::from_name(network_name)?;
```
cycraig, line 28: Standardise example network names to `"custom"`.

```suggestion
  // This is an arbitrarily defined network name
  let network_name = "custom";
  let network = Network::from_name(network_name)?;
```

examples/low-level-api/private_tangle.rs
```diff
@@ -0,0 +1,55 @@
...
+use identity::iota::Receipt;
+use identity::prelude::*;
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+  // This is an arbitrarily defined network name
+  let network_name = "atoi";
+
+  // Unwrap is fine since we provided a non-empty string.
+  let network = Network::from_name(network_name)?;
```
cycraig, line 24: Standardise example network names to `"custom"`.

```suggestion
  // This is an arbitrarily defined network name
  let network_name = "custom";
  let network = Network::from_name(network_name)?;
```

cycraig (approved):
Looks good to me! Minor comments.

Overall I think this is a compromise between keeping a lightweight enum as the `Network` was previously and a full struct, the latter of which might need a builder pattern etc. The string handling could be optimised a bit to avoid some clones by using `Cow` but that muddies the interfaces and might require adding a lifetime to the `Network`, so I think it's fine for now.

PR #374, opened 2021-08-30
Make sccache installation a bit faster on macOS

rajivshah3:
# Description of change
Makes `sccache` installation a bit faster on macOS by adding the `--preinstall` flag to the `brew update` command, which skips some checks that are redundant when running `brew update` just before `brew install`

N/A

## Type of change

- [x] Chore

CI ran successfully

## Change checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -115,7 +115,7 @@ jobs:
       - name: Install sccache (macos-latest)
         if: matrix.os == 'macos-latest'
         run: |
-          brew update
+          brew update --preinstall
           brew install sccache
 
       - name: Install sccache (windows-latest)


cycraig:
I didn't realise `sccache` takes 7-8 minutes to install on MacOS, we should consider removing it from that target entirely since that eclipses any compile time gains.

rajivshah3:
I think that should be resolved in the near future (when GitHub releases a new version of the [macOS image](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md)). It seems like `sccache` depends on OpenSSL, and a new version of OpenSSL was just released. So Homebrew updates OpenSSL, but then it wants to [update anything that depends on OpenSSL](https://github.com/iotaledger/identity.rs/runs/3466300241#step:10:161).

cycraig (approved):
Looks good to me then. Thanks for the explanation and PR!

PR #377, opened 2021-08-31
Enable lazy tangle updates in the `Account`

PhilippGackstatter:
# Description of change

Give users of the `Account` more control over when to publish to the tangle. For example, if an identity is created and a service and a method are added subsequently, this would currently result in three messages being published to the tangle. With this PR, all changes can be applied to a locally stored document, and then published in a single message. That increases the overall speed, since there are fewer network interactions and less Proof-of-Work, and reduces the used storage space on nodes.

## Implementation

The `Account` already organizes the updates to a document in `Commit`s. Each `Commit` is identified by a `Generation` (an incrementing `u32`). The basic idea is to store which generation was last published to the tangle, and when updating, publishing all commits that have been added since.

One potential use-case might require setting the `autopublish` configuration not just when building the `Account`, but also on the instance itself. Users could then switch between the modes very flexibly. I would probably not add this for now, though, and see if the requirement actually comes up in practice. This is mainly to be in line with the current immutability of the account config, but would be open to discuss this. It would be fairly simple to add this through an `AtomicBool` in the `Config`.

## Testing

Testing this really only makes sense if there is a tangle. As far as I'm aware, no tangle interaction is currently tested, except "passively" in the examples (they do run in CI now, but only once per day, not per PR). So we could do a couple of things:

1. We could skip testing the network aspects of this feature, too, which is basically everything in that case.
2. Another idea is to use a private tangle in CI and test it that way. That means we're not reliant on the mainnet or testnet, and tests should be faster. This would probably run per PR, and would also test the private tangle support as a side bonus. Would add a barrier to local testing though, as we need to run a private tangle locally, or skip these tests - neither is ideal.
3. We could actually start using the main- or testnet, increasing testing time, but requiring no special private tangle configuration and being always available, from both CI and locally.
4. We could add an example that includes assertions. This is quite common in Rust in general, but we don't have any assertions in our examples right now, so it would break our style a bit.

I've chosen the third option and added an example that tests directly on the testnet. The test runs in 5 seconds; not adding significantly to the CI execution time.

part of #325 

The other tasks of the issue will be completed in separate PRs.

## Type of change

- [ ] Documentation Fix

The stronghold implementation was tested by modifying the stronghold example. The added test uses the `MemStore`.

## Change checklist

### DIFF
--- a/identity-account/src/account/account.rs
+++ b/identity-account/src/account/account.rs
@@ -428,9 +428,7 @@ impl Account {
   }
 
   /// Push all unpublished changes for the given identity to the tangle in a single message.
-  ///
-  /// Returns the latest snapshot which is equivalent to the state on-tangle.
-  pub async fn publish_changes<K: IdentityKey>(&self, key: K) -> Result<IdentitySnapshot> {
+  pub async fn publish_changes<K: IdentityKey>(&self, key: K) -> Result<()> {
     let identity: IdentityId = self.try_resolve_id(key).await?;
 
     // Get the last commit generation that was published to the tangle.
@@ -443,12 +441,16 @@ impl Account {
     // Get the commits that need to be published.
     let commits = self.store.collect(identity, last_published).await?;
 
+    if commits.len() == 0 {
+      return Ok(());
+    }
+
     // Load the snapshot that represents the state on the tangle.
     let snapshot = self.load_snapshot_at(identity, last_published).await?;
 
     self.publish(snapshot, commits, true).await?;
 
-    self.load_snapshot(identity).await
+    Ok(())
   }
 
   /// Publishes according to the autopublish configuration.
--- a/identity-account/src/account/account.rs
+++ b/identity-account/src/account/account.rs
@@ -458,23 +458,25 @@ impl Account {
 
   /// Publishes according to the autopublish configuration.
   async fn publish(&self, snapshot: IdentitySnapshot, commits: Vec<Commit>, force: bool) -> Result<()> {
-    if force || self.config.autopublish {
-      let id = snapshot.id();
+    if !force && !self.config.autopublish {
+      return Ok(());
+    }
 
-      match Publish::new(&commits) {
-        Publish::Auth => self.process_auth_change(snapshot).await?,
-        Publish::Diff => self.process_diff_change(snapshot).await?,
-        Publish::None => {}
-      }
+    let id = snapshot.id();
 
-      if !commits.is_empty() {
-        let last_commit_generation = commits.last().unwrap().sequence();
-        // Publishing adds an AuthMessage or DiffMessage event, that contains the message id
-        // which is required to be set for subsequent updates.
-        // The next snapshot that loads the tangle state will require this message id to be set.
-        let generation = Generation::from_u32(last_commit_generation.to_u32() + 1);
-        self.store.set_published_generation(id, generation).await?;
-      }
+    match Publish::new(&commits) {
+      Publish::Auth => self.process_auth_change(snapshot).await?,
+      Publish::Diff => self.process_diff_change(snapshot).await?,
+      Publish::None => {}
+    }
+
+    if !commits.is_empty() {
+      let last_commit_generation = commits.last().unwrap().sequence();
+      // Publishing adds an AuthMessage or DiffMessage event, that contains the message id
+      // which is required to be set for subsequent updates.
+      // The next snapshot that loads the tangle state will require this message id to be set.
+      let generation = Generation::from_u32(last_commit_generation.to_u32() + 1);
+      self.store.set_published_generation(id, generation).await?;
     }
 
     Ok(())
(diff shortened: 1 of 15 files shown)


identity-account/src/account/account.rs
```diff
@@ -418,6 +428,52 @@ impl Account {
...
+    let snapshot = self.load_snapshot_at(identity, last_published).await?;
+
+    self.publish(snapshot, commits, true).await?;
+
+    Ok(())
+  }
+
+  /// Publishes according to the autopublish configuration.
+  async fn publish(&self, snapshot: IdentitySnapshot, commits: Vec<Commit>, force: bool) -> Result<()> {
+    if force || self.config.autopublish {
```
cycraig, line 455: nit: return early if this check fails.

E.g.
```rust
if !force && !self.config.autopublish {
  return Ok(());
}
```

identity-account/src/account/account.rs
```diff
@@ -418,6 +428,52 @@ impl Account {
...
+
+      if !commits.is_empty() {
+        let last_commit_generation = commits.last().unwrap().sequence();
+        // Publishing adds an AuthMessage or DiffMessage event, that contains the message id
+        // which is required to be set for subsequent updates.
+        // The next snapshot that loads the tangle state will require this message id to be set.
+        let generation = Generation::from_u32(last_commit_generation.to_u32() + 1);
+        self.store.set_published_generation(id, generation).await?;
+      }
+    }
```
cycraig, line 472: I think there's a potential race condition with `set_published_generation` here but also updates and `publish_changes` in general? The potential effects of a race condition don't seem to be fatal, e.g. double publishing a commit, but we should discuss it, unless I'm wrong? 

Fixing it would probably require fine-grained `RwLocks` on each of the local `IdentityId` entries for most operations.

identity-account/src/storage/memstore.rs
```diff
@@ -51,8 +53,9 @@ pub struct MemStore {
 impl MemStore {
   pub fn new() -> Self {
     Self {
-      expand: false,
+      expand: true,
```
cycraig, line 56: This defaults to printing private keys in debug strings, was changing this intentional? The `MemStore` is likely to be used only for testing/debugging during development and not in production, but defaulting to not potentially printing private keys in logs seems safer.

identity-account/src/account/account.rs
```diff
@@ -466,6 +524,12 @@ impl Config {
     self
   }
 
+  /// Sets the account auto-publish behaviour.
+  pub fn autopublish(mut self, value: bool) -> Self {
```
cycraig, line 528: Nit: this comment should also explain what auto-publish does and the default value? This probably applies to all config options really.

E.g.
```
  /// Sets the account auto-publish behaviour.
  /// - `true` => publish to the Tangle on every change
  /// - `false` => never publish automatically
  /// 
  /// Default: `true`
  pub fn autopublish(mut self, value: bool) -> Self {
```

identity-account/src/account/builder.rs
```diff
@@ -55,6 +55,12 @@ impl AccountBuilder {
     self
   }
 
+  /// Sets the account auto-save behaviour.
+  pub fn autopublish(mut self, value: bool) -> Self {
```
cycraig, line 59: Nit: same comment as `Config::autopublish`.

identity-account/src/account/account.rs
```diff
@@ -418,6 +434,54 @@ impl Account {
     Ok(())
   }
 
+  /// Push all unpublished changes for the given identity to the tangle in a single message.
+  pub async fn publish_changes<K: IdentityKey>(&self, key: K) -> Result<()> {
+    let identity: IdentityId = self.try_resolve_id(key).await?;
```
cycraig, line 439: I think `publish_changes` also needs a write lock to prevent the potential race condition in `publish`?

identity-account/src/identity/identity_index.rs
```diff
@@ -79,6 +100,46 @@ impl Default for IdentityIndex {
...
+
+        for value in data.values() {
+          locks.insert(*value, Arc::new(RwLock::new(*value)));
+        }
+
+        IdentityIndex { data, locks }
+      },
+    )
+  }
+}
```
cycraig, line 141: Can we change `get_lock` to lazily create locks as-needed instead of creating them all at once here, or on `insert`? That would be a performance benefit too since one would not need locking at all if not using `update_identity` / `publish_changes`, and then only for those identities being updated.

I am a bit concerned with how the `IdentityIndex` is used by the `Storage` where it is cloned (in `MemStore`) or serialized then de-serialized (`Stronghold`) that one could circumvent the locks by using different copies of them. It may be better to store the lock map in the `Account` struct proper to avoid these issues altogether?

identity-account/src/identity/identity_index.rs
```diff
@@ -64,12 +82,15 @@ impl IdentityIndex {
...
       Entry::Vacant(entry) => {
         entry.insert(id);
-        Ok(())
       }
     }
+
+    self.locks.insert(id, Arc::new(RwLock::new(id)));
+
+    Ok(())
   }
```
cycraig, line 94: IdentityIndex::del() should also remove locks when the identity is removed.

identity-account/src/identity/identity_index.rs
PhilippGackstatter: You're right, this is better. I was previously worried about contention on the `RwLock<IdentityIndex>` in the account, since `get_lock` would then modify the index and require the lock to be acquired as a write lock rather than just a read one. But most likely the performance is better this way then adding locks for a large number of identities.

Yes theoretically possible, but prevent by the design of the account. The index is only cloned when returned by storage, and added when the account is created, and never modified as a whole. So for any given account no cloning happens, and thus this should be fine.

Having the locks in the account requires more locks, as you would have `RwLock<IdentityIndex>` and also a separate `RwLock<HashMap<IdentityId, Arc<RwLock<IdentityId>>>>`, which I wanted to avoid. One possibility is to remove the custom `Clone` impl and add a derive, in which case the `Arc` around the lock would be cloned, and thus your concern of getting around the locks when cloning would be alleviated.

cycraig: I think using something like `Dashmap<IdentityId, Arc<RwLock<IdentityId>>>` in the account wouldn't be too terrible.

`Clone` is only used by `MemStore`, the index is (de)serialized by the Stronghold storage so it would still suffer from that. I think in general this may not be a problem since, as you indicated, the `Account` itself is not cloned. I also do not see any cases where an `Account` re-loads the index from storage after the initial configuration but may be incorrect.

Overall the locking seems fine. I'll take another look as there may be places where a read lock is appropriate.

identity-account/tests/lazy.rs
```diff
@@ -0,0 +1,151 @@
...
+      ok @ Ok(_) => {
+        return ok;
+      }
+      Err(AccountError::IotaError(IotaError::ClientError(client_err))) => {
+        eprintln!("test run {} errored with {:?}", test_run, client_err);
+
+        if test_run == test_runs - 1 {
+          return Err(AccountError::IotaError(IotaError::ClientError(client_err)));
+        }
+      }
```
cycraig, line 143: Nit: this branch can be simplified a bit, it changes what's printed but it doesn't matter that much either way.

```suggestion
      error @ Err(AccountError::IotaError(IotaError::ClientError(_))) => {
        eprintln!("test run {} errored with {:?}", test_run, error);

        if test_run == test_runs - 1 {
          return error;
        }
      }
```

cycraig (approved):
Looks good to me! Minor comments.

I would still prefer the lock to be in the `Account` itself rather than the index to avoid all any potential index cloning, storage issues and workarounds for e.g. `PartialEq` but this still solves the race condition.

PR #382, opened 2021-09-03
Extend Network doc comment for other networks

Thoralf-M:
# Description of change
Extend Network doc comment for other networks, because without it it's not clear why it might not work

## Type of change

- [x] Documentation Fix

### DIFF
--- a/identity-iota/src/tangle/network.rs
+++ b/identity-iota/src/tangle/network.rs
@@ -39,7 +39,8 @@ impl Network {
   ///
   /// The names `"test"` and `"main"` will be mapped to the well-known [`Testnet`][Network::Testnet]
   /// and [`Mainnet`][Network::Mainnet] networks respectively. Other inputs will return an instance
-  /// of [`Other`][Network::Other] if the name is valid.
+  /// of [`Other`][Network::Other] if the name is valid. It must match a part or all of the networkId returned in the
+  /// nodeinfo from a node. For exampe, if the networkId is `"private-tangle"`, `"tangle"` can be used.
   ///
   /// Network names must comply with the IOTA DID Method spec, that is: be non-empty, at most
   /// 6 characters long, and only include alphanumeric characters `0-9` and `a-z`.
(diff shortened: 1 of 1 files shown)


Thoralf-M:
If you don't use .with_network() for the ClientBuilder then it's not needed to match and the only difference would be that also nodes from another network, for example if you provide "mainnet" and a single testnet node, then it would just use this testnet node and now it would return an error that no node is available or ` Failed to get an answer from all nodes`

cycraig:
I would argue that we still want that behaviour with `.with_network()` to prevent mismatches in where DIDs are published/resolved, e.g. we don't want a `"did:iota:test:xyz"` DID published on the mainnet, nor ones with a custom network name? I acknowledge that it might be confusing for new developers when it doesn't work with only a cryptic error message, as per Discord, so the restrictions on DID network names should be relaxed a bit to match those of private Tangle nodes (at least "private-tangle" for starters).

identity-iota/src/tangle/network.rs
```diff
@@ -39,7 +39,8 @@ impl Network {
   ///
   /// The names `"test"` and `"main"` will be mapped to the well-known [`Testnet`][Network::Testnet]
   /// and [`Mainnet`][Network::Mainnet] networks respectively. Other inputs will return an instance
-  /// of [`Other`][Network::Other] if the name is valid.
+  /// of [`Other`][Network::Other] if the name is valid. It must match a part or all of the networkId returned in the
+  /// nodeinfo from a node. For exampe, if the networkId is `"private-tangle"`, `"tangle"` can be used.
```
cycraig, line 43:
```suggestion
  /// of [`Other`][Network::Other] if the name is valid. It must match a part or all of the `networkId` returned in the
  /// nodeinfo from a node. For example, if the networkId is `"private-tangle"`, `"tangle"` can be used.
```

cycraig (approved):
Spelling fix otherwise looks good to me. Thanks for the clarification.

I'll open another PR for any changes to the network name pending discussions.

PR #367, opened 2021-08-24
Remove target caches in GitHub Actions

eike-hass:
# Description of change
Remove cargo target caches to decrease cache entries and size, hopefully increasing cache hits for the other caches.

## Type of change
Add an `x` to the boxes that are relevant to your changes.

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [X] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

## How the change has been tested
Describe the tests that you ran to verify your changes.
Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist
Add an `x` to the boxes that are relevant to your changes.

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

PR #386, opened 2021-09-06
Change build workflow step to avoid recompilation

cycraig:
# Description of change
Change the Rust build step in the build-and-test GitHub Actions workflow to avoid recompilation and improve CI speed.

The command changes from:
`cargo build --all --release`
to: 
`cargo test --all --release --no-run`

This is done because the compilation performed in the "Run Tests" step typically takes a significant amount of time (roughly 13 minutes) while the tests themselves complete in a few seconds to one minute. The change avoids some recompilation by compiling the tests and examples at the same time as the rest of the library, rather than separately. 

The disadvantages are that the build step is slightly different than it would be for a regular release as a library, and the tests/examples may fail compilation while a regular build may have passed. This would make debugging why a CI build failed slightly more difficult, as one would not be able to see that the build step passed but the test build step failed, but that can be resolved easily by reading the logs. I would also argue that we always want the tests and examples to compile anyway.

See the following discussion for more insight: https://github.com/iotaledger/identity.rs/pull/367#issuecomment-913808030

This PR also removes the redundant "Print sccache stats" step, as `sccache` prints statistics when it is stopped anyway.

## Type of change

- [x] Bug fix/chore (a non-breaking change which fixes an issue)
- [ ] Documentation Fix

Tested the command locally, pending workflow run against this PR.

## Change checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -133,17 +133,15 @@ jobs:
       - name: Build
         uses: actions-rs/cargo@v1
         with:
-          command: build
-          args: --all --release
+          # Build the library, tests, and examples without running them to avoid recompilation in the run tests step
+          command: test
+          args: --all --release --no-run
 
       - name: Run tests
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --all --release
-
-      - name: Print sccache stats
-        run: sccache --show-stats
 
       - name: Stop sccache server
         run: sccache --stop-server || true
@@ -243,9 +241,6 @@ jobs:
         with:
           command: test
           args: --manifest-path ./libjose/Cargo.toml --release
-
-      - name: Print sccache stats
-        run: sccache --show-stats
 
       - name: Stop sccache server
         run: sccache --stop-server || true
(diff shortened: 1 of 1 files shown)


cycraig:
The results without any caching are a bit disappointing: https://github.com/iotaledger/identity.rs/actions/runs/1207022847
The timings are very similar to before the change also without caching: https://github.com/iotaledger/identity.rs/runs/3525111153

Similarly with caching: https://github.com/iotaledger/identity.rs/runs/3527664317
compared to the best prior run with caching: https://github.com/iotaledger/identity.rs/runs/3506689025

| | ubuntu-latest | macos-latest | windows-latest |
| ------------- | ------------- | ------------- | ------------- |
| before (no cache) | 21m 23s | 28m 9s | 27m 17s |
| after (no cache) | 18m 43s | 26m 37s | 29m 0s |
| before (cache) | 14m 18s | 13m 52s | 14m 16s |
| after (cache) | 12m 4s | 10m 48s | 13m 56s |

For Ubuntu and MacOS, the times are lower by around one standard deviation (established in https://github.com/iotaledger/identity.rs/pull/367#issuecomment-913808030), but for Windows it makes little difference.

We would need better averages to make a definite conclusions but it _seems_ faster, just not dramatically.

PhilippGackstatter (approved):

PR #390, opened 2021-09-08
use Cargo.lock for caches, remove libjose caches

eike-hass:
# Description of change
Remove caches from libjose to reduce global cache size. Rework remaining caches to use Corgo.lock hash. 

## Type of change

- [ ] Documentation Fix

Ran action, all jobs pass.

## Change checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -227,10 +227,12 @@ jobs:
             ${{ matrix.os }}-cargo-${{ github.job }}-
             ${{ matrix.os }}-cargo-
 
+      # Generate Cargo.lock files for build, sccache cache keys.
+      # Allows dependencies updated on crates.io between runs to trigger storing an updated cache, 
+      # which hashing Cargo.toml files alone does not.
       - name: Cargo update
         uses: actions-rs/cargo@v1
         with:
-          # Create Cargo.lock file for more accurate caches
           command: update
 
       - name: Cache wasm target
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -74,10 +74,12 @@ jobs:
             ${{ matrix.os }}-cargo-${{ github.job }}-
             ${{ matrix.os }}-cargo-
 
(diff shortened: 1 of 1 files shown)


cycraig:
The `build-and-test-libjose` task is failing due to the workflow-global `RUSTC_WRAPPER` environment variable which tells it to still use `sccache` during compilation, but the change removed it.

The solution would be to move the `RUSTC_WRAPPER` variable out of the global environment into job-specific environments (excluding `build-and-test-libjose`).
```
env:
  RUST_BACKTRACE: full
  RUSTC_WRAPPER: sccache
  SCCACHE_DOWNLOAD_LINK: https://github.com/mozilla/sccache/releases/download
  SCCACHE_VERSION: v0.2.15
  SCCACHE_CACHE_SIZE: 2G
  SCCACHE_IDLE_TIMEOUT: 0
  # SCCACHE_RECACHE: 1 # uncomment to clear sccache cache, then re-comment
```

.github/workflows/build-and-test.yml
```diff
@@ -289,15 +225,20 @@ jobs:
             ${{ matrix.os }}-cargo-${{ github.job }}-
             ${{ matrix.os }}-cargo-
 
+      - name: Cargo update
+        uses: actions-rs/cargo@v1
+        with:
+          # Create Cargo.lock file for more accurate caches
+          command: update
```
cycraig, line 232: Same as above:

```suggestion
      # Generate Cargo.lock files for build, sccache cache keys.
      # Allows dependencies updated on crates.io between runs to trigger storing an updated cache, 
      # which hashing Cargo.toml files alone does not.
      - name: Cargo update
        uses: actions-rs/cargo@v1
        with:
          command: update
```

cycraig (approved):
Looks good to me. 

As discussed, this may result in more cache churn (and therefore cache misses) based on how often out dependencies push updates but maybe the removal of the `build-and-test-libjose` caches offsets that. It's worth noting for any observers that from previous runs of `build-and-test-libjose` without caching typically only take 2-3 minutes, hence the removal of caching there is acceptable.

PR #391, opened 2021-09-08
Internalize `Account` commands

### LINKED ISSUE #325: [Task] Improve Account API by internalizing commands

PhilippGackstatter:
# Description of change

Following the discussion from #325, this implements the "internalization" of commands in the `Account`. What this means, is that what was previously created through associated functions on the `Command`, such as `Command::create_method`, is now created through the account itself. As an example, the current approach looked like this:

```rust
 let command = Command::create_method()
  .fragment("my-auth-key")
  .scope(MethodScope::Authentication)
  .finish()?;

account.update_identity(document, command).await?;
```

Now the same is achieved in one statement:

```rust
account
  .update_identity(did)
  .create_method()
  .fragment("my-auth-key")
  .scope(MethodScope::Authentication)
  .apply()
  .await?;
```

closes #325

## Type of change

- [ ] Documentation Fix

Since this is a non-functional change, no tests were added. The examples show the two different styles this can be used. Either inline, as above, or by first creating an `let identity_updater = account.update_identity(did)` and then applying multiple updates on that `identity_updater`. These successfully compile.

## Change checklist


identity-account/src/account/account.rs
```diff
@@ -153,13 +154,12 @@ impl Account {
...
-    let identity: IdentityId = self.try_resolve_id(key).await?;
-
-    self.process(identity, command, true).await?;
-
-    Ok(())
+  /// Return the `IdentityUpdater` for the given `key`.
+  ///
+  /// On this type, various operations can be executed
+  /// that modify an identity, such as creating services or methods.
+  pub fn update_identity<K: IdentityKey + Clone>(&self, key: K) -> IdentityUpdater<'_, K> {
```
cycraig, line 161: Nit: comment suggestion. Ignore if we want to exclude `Command` from the public docs entirely but I think we need some way to reference the allowed operations.

```suggestion
  /// Return an `IdentityUpdater` for the given `key`.
  ///
  /// Allows executing various commands that modify an identity, 
  /// such as adding or removing services and methods.
  ///
  /// See [`Command`] for more details.
  pub fn update_identity<K: IdentityKey + Clone>(&self, key: K) -> IdentityUpdater<'_, K> {
```

identity-account/src/identity/identity_updater.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::account::Account;
+
+use super::IdentityKey;
+
+/// A struct created by the [`Account::update_identity`] method, that
+/// allows executing various updates on the identity it was created on.
+#[derive(Debug, Clone)]
+pub struct IdentityUpdater<'account, K: IdentityKey + Clone> {
+  account: &'account Account,
```
cycraig, line 12: I really like the named lifetime!

identity-account/src/events/macros.rs
```diff
@@ -33,46 +33,47 @@ macro_rules! impl_command_builder {
...
+          let account = self.account;
+          let update = $crate::events::Command::$ident {
             $(
               $field: impl_command_builder!(@finish self $requirement $field $ty $(= $value)?),
             )*
-          })
+          };
+          account.apply_command(self.key, update).await?;
+          Ok(())
         }
```
cycraig, line 71: Did we decide for or against chaining consecutive updates? We could return a new `IdentityUpdater` on `apply()` to enable chaining, but I think it was decided not to be particularly useful.

identity-account/src/events/macros.rs
```diff
@@ -33,46 +33,47 @@ macro_rules! impl_command_builder {
...
-          Self::new()
-        }
-      }
-
-      impl $crate::events::Command {
-        pub fn [<$ident:snake>]() -> [<$ident Builder>] {
-          [<$ident Builder>]::new()
+      impl<'account, K: $crate::identity::IdentityKey + Clone> $crate::identity::IdentityUpdater<'account, K> {
+        pub fn [<$ident:snake>](&self) -> [<$ident Builder>]<'account, K> {
+          [<$ident Builder>]::new(self.account(), self.key().clone())
```
cycraig, line 76: We could alter the builder to take the `self` instead of `&self` then destruct it to get the `account` and `key` references without cloning. This would prevent re-using an `IdentityUpdater` for more than one command but I'm not convinced multiple commands per updater is necessary given how lightweight creating a new one is.

cycraig (approved):
Looks great! 

Only a few questions below. I'm also a bit wary of allowing `create_identity` to be called from the `IdentityUpdater`. Since it does not operate on an existing `did`, `account.update_identity(did).create_identity()` seems a bit weird and might imply that you're creating an identity using that `did` as a controller? (Reaching a bit I know). 

I think `create_identity` is the only exception so if there's an easy way to exclude it in the macro that's great, otherwise leaving it is also acceptable for now.

identity-account/src/events/macros.rs
PhilippGackstatter: I'm not sure we reached a proper consensus on it, but in my opinion running one update is complex enough, that I wouldn't personally do chaining. I think with what is shown in the examples, we have a nice middle ground between the previous "two statements for one update" and "we could run multiple updates in one statement".

identity-account/src/account/account.rs
PhilippGackstatter: Agreed. Perhaps we should simply pull the methods that create the builders out of the macro and implement them by hand. We could then document each operation on these methods.

Then we can also exclude the `create_identity` builder method, although it can also be removed by removing the macro invocation for it.

Another approach for the documentation would be to summarize it on the `IdentityUpdater` struct documentation.

identity-account/src/events/macros.rs
PhilippGackstatter: I've got rid of the `Clone` requirement and the `clone()` itself by letting the updater take a reference to the `key` as well. It can still be used as before in the methods example, although I've updated that one.

examples/account/methods.rs
```diff
@@ -22,62 +22,57 @@ async fn main() -> Result<()> {
   let snapshot: IdentitySnapshot = account.create_identity(IdentityCreate::default()).await?;
 
   // Retrieve the DID from the newly created Identity state.
-  let document: &IotaDID = snapshot.identity().try_did()?;
+  let did: &IotaDID = snapshot.identity().try_did()?;
 
-  // Add a new Ed25519 (defualt) verification method to the identity - the
+  // Get the updater for the given `did`, so we can run multiple updates on it.
+  let did_updater: IdentityUpdater<'_, _> = account.update_identity(did);
```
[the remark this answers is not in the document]
PhilippGackstatter: I've updated it. I figured the service example showed one approach for using it, while the services example was a good way to show another approach, if one actually has multiple updates. But I'm fine with either one, we can also hint at it in the advanced documentation or something.

PhilippGackstatter:
I've made some updates:

- Use reference in `IdentityUpdater` to make it cloneable without requiring `Clone` on the `IdentityKey`.
- Document the builder types.
- Got rid of `Command` in the public interface, renamed `CommandError` to `UpdateError`
  - Moved tests inside the `account` crate, so `Command` can be `pub(crate)` rather than requiring the `#[doc(hidden)]` hack.
- In the examples, use `update_identity` on every invocation rather than creating an extra variable.

identity-account/src/account/account.rs
cycraig: I concur. Manually implementing the functions for the `IdentityUpdater` would be good since developers currently lack e.g. line completion or even a complete list of builder functions without looking at the docs. Hopefully there's a way to still use macros internally to avoid too much duplicate code though.

Enumerating the list of builder methods available for the `IdentityUpdater` in its doc-comment may be useful to reference but not ideal.

identity-account/src/error.rs
```diff
@@ -90,7 +90,7 @@ pub enum Error {
   ServiceNotFound,
   /// Caused by attempting to perform a command in an invalid context.
   #[error("Command Error: {0}")]
-  CommandError(#[from] crate::events::CommandError),
+  CommandError(#[from] crate::events::UpdateError),
```
cycraig, line 93: This should change to `UpdateError` too.

PR #407, opened 2021-09-19
Initial content for VC Wiki Overview page

m-renaud:
# Description of change

Initial prose introducing VCs in IOTA.

N/A

## Type of change

- [x] Documentation Fix

Built docs locally.

## Change checklist

### DIFF
--- a/documentation/docs/getting-started/verifiable_credentials/overview.md
+++ b/documentation/docs/getting-started/verifiable_credentials/overview.md
@@ -9,7 +9,38 @@ keywords:
 - person
 ---
 
+Credentials are statements about an entity, such as properties that the entity posseses or capabilities that they have (like drivers licences, passports, or a person's age). Verifiable Credentials (VCs) are statements about the holder of the credential that can be cryptographically verified by a third party, either online or in person. Additionally, the holder decides what is shared and who it is shared.
 
-Verifiable Credentials are statements about the holder. They can be verified online or in person, and the holder decides who to share them with.
+There are several types of actors that play different roles in a verifiable credential system. We'll start with a common example from the world today and outline the roles that various entities play.
 
-[Verifiable Credentials Data Model 1.0 Specification](https://w3c.github.io/vc-data-model/)
+:::tip Example - Passport Issuance
+
+A government (the _Issuer_) issues a passport asserting citizenship (the _Verifiable Credential_) to Alice (the _Subject_ and _Holder_), and writes the information to a database (the _Verifiable Data Registry_). When crossing the border, Alice (the _Holder_) presents her passport to a border agent (the _Verifier_) who can verify that Alice (the _Subject_) is indeed a citizen.
+
+:::
+
+**Subject:** An entity about which claims are made – Alice (the _Subject_) is a citizen of this country.
+
+**Holder:** An entity which posseses verifiable credentials – Alice (the _Holder_) posseses the passport (the _VC_).
+
+**Issuer:** An entity which asserts claims about a subject – The governing body (the _Issuer_), which is trusted, issues Alice a passport.
+
+**Verifier:** An entity which check's if the VC a holder presents is legitimate – The border agent (the _Verifier_) trusts the government (the _Issuer_) which issued Alice her passport, and validates that Alice (the _Subject_) is a citizen.
+
+:::note
+
+See the [Verifiable Credentials Data Model 1.0 Specification](https://w3c.github.io/vc-data-model/) for more information.
+
+:::
+
+### Verifiable Credentials in IOTA
+
+In the IOTA Identity framework, instead of a physical passport being given to Alice with the passport information being written into a centralized database owned by the government, Alice receives a digital verifiable credential, and the information required for verification in the future is written to the Tangle.
+
+At a high level, the creation and verification of a VC on IOTA works as follows:
+
+The first step is to create a verifiable credential which requires the subject (Alice) and issuer (the government) to have DIDs published to the tangle, and a set of statements being asserted (that Alice has a passport). The issuer then publishes a credential to the Tangle which contains the `credentialSubject` (Alice's DID in our example), the set of statements, and a proof that can be used to assert the authenticity of the credentials by verifiers.
+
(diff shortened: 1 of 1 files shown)


documentation/docs/getting-started/verifiable_credentials/overview.md
```diff
@@ -9,7 +9,38 @@ keywords:
 - person
 ---
 
+Credentials are statements about an entity, such as properties that the entity posseses or capabilities that they have (like drivers licences, passports, or a person's age). Verifiable Credentials (VCs) are statements about the holder of the credential that can be cryptographically verified by a third party, either online or in person. Additionally, the holder decides what is shared and who it is shared.
```
JelleMillenaar, line 12: Perhaps avoid "Holder" until introduced. Can be replaced with "Subject" or something more clear.

documentation/docs/getting-started/verifiable_credentials/overview.md
```diff
@@ -9,7 +9,38 @@ keywords:
...
 
-Verifiable Credentials are statements about the holder. They can be verified online or in person, and the holder decides who to share them with.
+There are several types of actors that play different roles in a verifiable credential system. We'll start with a common example from the world today and outline the roles that various entities play.
 
-[Verifiable Credentials Data Model 1.0 Specification](https://w3c.github.io/vc-data-model/)
\ No newline at end of file
+:::tip Example - Passport Issuance
+
+A government (the _Issuer_) issues a passport asserting citizenship (the _Verifiable Credential_) to Alice (the _Subject_ and _Holder_), and writes the information to a database (the _Verifiable Data Registry_). When crossing the border, Alice (the _Holder_) presents her passport to a border agent (the _Verifier_) who can verify that Alice (the _Subject_) is indeed a citizen.
```
JelleMillenaar, line 18: > and writes the information to a database (the _Verifiable Data Registry_)

We don't do this. The government signs the VC with a public key, stored on a Tangle. The only copy of the passport that needs to exist is in Alice's control (in an App or something).

JelleMillenaar: Ah sorry, is this text representing the current physical / digital centralized format?

documentation/docs/getting-started/verifiable_credentials/overview.md
```diff
@@ -9,7 +9,38 @@ keywords:
...
+
+:::
+
+### Verifiable Credentials in IOTA
+
+In the IOTA Identity framework, instead of a physical passport being given to Alice with the passport information being written into a centralized database owned by the government, Alice receives a digital verifiable credential, and the information required for verification in the future is written to the Tangle.
+
+At a high level, the creation and verification of a VC on IOTA works as follows:
+
+The first step is to create a verifiable credential which requires the subject (Alice) and issuer (the government) to have DIDs published to the tangle, and a set of statements being asserted (that Alice has a passport). The issuer then publishes a credential to the Tangle which contains the `credentialSubject` (Alice's DID in our example), the set of statements, and a proof that can be used to assert the authenticity of the credentials by verifiers.
```
JelleMillenaar, line 42: The Issuer does NOT send the credential to a Tangle, as that would break GDPR's "right-to-be-forgotten". Instead, Alice received the only copy of the Verifiable Credential. The only thing that is uploaded to a Tangle is the public key used by the Issuer in order to verify the Verifiable Credential.

JelleMillenaar (changes requested):
Excellent start, but needs a few changes. Using this PR as an educational moment =D 

I will add a section to this PR about GDPR compliance.

documentation/docs/getting-started/verifiable_credentials/overview.md
m-renaud: Yes sorry, my plan here was to illustrate a real world example (which as you mention is currently done in physical / centralized format) and use that to explain the parts of the VC spec and how IOTA implements them.

I was thinking more generally that having a consistent end-to-end narrative throughout that ties all of the IOTA client libraries would help folks reading it to build up a proper mental model.

Thoughts on the best way to clarify this? I could re-word the last sentence of the paragraph before to explicitly call out that this is "an example of how things work in the world today" (or something like that). wdyt?

documentation/docs/getting-started/verifiable_credentials/overview.md
m-renaud: Haha I initially had "subject" for both, but then was re-reading the spec and thought it would be best to use the same terminology. Will change to "(VCs) are statements about the subject (eg. Alice has a drivers licence) that can be cryptographically verified".

This way it doesn't mention "subject", which would technically be misaligned with the spec, or "holder" at all.

documentation/docs/getting-started/verifiable_credentials/overview.md
m-renaud: You're absolutely right, not sure what I was thinking when I wrote that sentence 😅 Updated this paragraph. There's a lot more details about verifiable presentations and proofs that need to be added here as well.

documentation/docs/getting-started/verifiable_credentials/overview.md
JelleMillenaar: Yea, the explicit call-out would work :)

documentation/docs/getting-started/verifiable_credentials/overview.md
m-renaud: Going to close this out. Please open a new thread if you'd like to see any other changes.

documentation/docs/getting-started/verifiable_credentials/overview.md
m-renaud: The sentence before the example now reads:

"We'll start with a common example of how things work in the world today using physical credentials and centralized databases, and outline the roles that various entities play in the Verifiable Credential system.""

documentation/docs/getting-started/verifiable_credentials/overview.md
m-renaud: PTAL and resolve this conversation if you're good with the new wording.

JelleMillenaar (approved):
All good now, thanks for the fixes :)

PR #415, opened 2021-09-26
Add M1 Mac instructions for building WASM bindings

m-renaud:
# Description of change

The npm version of wasm-pack does not support the new Darwin arm64 architecture. For some reason installing through cargo works though, so suggest that as a workaround.

N/A

## Type of change

- [x] Documentation Fix

N/A - README.md change only.

## Change checklist

### DIFF
--- a/bindings/wasm/README.md
+++ b/bindings/wasm/README.md
@@ -19,6 +19,8 @@ npm install @iota/identity-wasm@dev
 
 ## Build
 
+**NOTE:** _If you're on an M1 Mac the `npm install` command below will fail due to missing support for the architecture in the `npm` version of `wasm-pack`. To work around this use `cargo install wasm-pack` and then go directly to the `npm run` step._
+
 Alternatively, you can build the bindings if you have Rust installed. If not, refer to [rustup.rs](https://rustup.rs) for the installation. Then install the necessary dependencies using:
 ```bash
 npm install


m-renaud:
@cycraig can you review?

cycraig:
Unfortunately I am unable to verify the workaround as I do not have access to a Mac with an M1 chip; I can only offer general comments.

>  **NOTE:** _If you're on an M1 Mac the `npm install` command below will fail due to missing support for the architecture in the `npm` version of `wasm-pack`. To work around this use `cargo install wasm-pack` and then go directly to the `npm run` step._

To be clear, the developer would still have to run `npm install` (and have it fail) to get the rest of the packages right? It's a tiny bit ambiguous whether they can just skip the step altogether. 

The message alone lacks some context or directions for more information. This is a long-standing issue with `wasm-pack` we should link to: https://github.com/rustwasm/wasm-pack/issues/952 . I wonder if opening up an issue here (or leaving this PR open) instead of updating the README.md file would offer better visibility and give more context?

Is there any way to easier for new developers by checking if they're using an M1 Mac when they run `npm install`? E.g. if `node -p process.arch` reports `arm64` (or whatever architecture is reported for an M1 chip) we automatically run `cargo install wasm-pack` for them or display an error message with instructions?

Does emulating an x86 architecture also work around this issue? E.g. https://www.jurnalanas.com/node-js-mac-m1/

PR #421, opened 2021-09-28
Add default signing key as a generic verification method

### LINKED ISSUE #195: [Task] Rename Authentication Key and Chain
## Description
Describe the task, this may be scoped beyond a single PR.

Instead of using the authentication field and key for DID messages, utilize a general verification method. As this removes the logic behind the authentication chain name, I request a rename to the Integration chain (Also in the code). The name of the key we use for signing DID messages may be anything, but we can default to #_sign-x, where x is the number of the key, incremented every time it is updated. This should not be enforced though. 

## Motivation
Describe the motivation for the feature to be developed.

We currently always use an authentication key for signing and updating DID messages. This doesn't fall in line with the standardized way of using the authentication verifiable relationship according to the DID standard. Instead, we should utilize a general verification method.

## Resources

## To-do list

- [x] Rename AuthChain and other references to IntChain (Integration Chain) in the code
- [ ] Remove enforcement of signing with "auth key"
- [ ] Edit examples to add verification method with a #_sign-1 key.
- [ ] Edit account to standardize the #_sign-x naming instead of auth keys.
- [x] Edit documentation to rename all references to AuthChain

## Change checklist

PhilippGackstatter:
# Description of change

- Rename the last remaining occurrences of auth chain to int chain. This is a breaking change because multiple method names have changed.
- Add the default verification method we use to sign document update messages as a generic verification method, instead of only as an [authentication](https://www.w3.org/TR/did-core/#authentication) method. However, the `authentication` verification relationship is attached to the method. An example of a newly generated did document thus looks like:

```
{
    "id": "did:iota:dev:7uUb2ES6qnuzsKjWG7Ft8iVz5Q7S3ZgXrcJGTiKskBi8",
    "verificationMethod": [
        {
            "id": "did:iota:dev:7uUb2ES6qnuzsKjWG7Ft8iVz5Q7S3ZgXrcJGTiKskBi8#_sign-0",
            "controller": "did:iota:dev:7uUb2ES6qnuzsKjWG7Ft8iVz5Q7S3ZgXrcJGTiKskBi8",
            "type": "Ed25519VerificationKey2018",
            "publicKeyBase58": "4GeSr3ZmRGDQiSmDni2Su5MZXm4gzgTmhCrxNybWAquy"
        }
    ],
    "authentication": [
        "did:iota:dev:7uUb2ES6qnuzsKjWG7Ft8iVz5Q7S3ZgXrcJGTiKskBi8#_sign-0"
    ],
    "created": "2021-09-28T11:53:47Z",
    "updated": "2021-09-28T11:53:47Z"
}
```

fixes #195

## Type of change

- [ ] Documentation Fix

Change existing tests appropriately to test the changed functionality.

## Change checklist

### DIFF
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -1074,4 +1074,19 @@ mod tests {
     let diff_index = IotaDocument::diff_index(&message_id).expect("failed to generate diff_index");
     assert_eq!(diff_index, "2g45GsCAmkvQfcrHGUgqwQJLbYY3Gic8f23wf71sGGGP");
   }
+
+  #[test]
+  fn test_new_document_has_verification_method_with_authentication_relationship() {
+    let keypair: KeyPair = generate_testkey();
+    let document: IotaDocument = IotaDocument::from_keypair(&keypair).unwrap();
+
+    let verif_method = document.resolve("#authentication").unwrap();
+    let auth_method = document.authentication();
+
+    // `methods` returns all embedded verification methods, so only one is expected.
+    assert_eq!(document.methods().count(), 1);
+
+    // Assert that the verification method and the authentication method are the same
+    assert_eq!(verif_method, auth_method);
+  }
 }
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -1088,5 +1088,11 @@ mod tests {
 
     // Assert that the verification method and the authentication method are the same
     assert_eq!(verif_method, auth_method);
+
+    // Assert that the fragment of the authentication method reference is `authentication`
+    match document.document.authentication().first().unwrap().clone().into_inner() {
+      MethodRef::Refer(did) => assert_eq!(did.fragment().unwrap_or_default(), "authentication"),
+      MethodRef::Embed(_) => panic!("authentication method should be a reference"),
+    }
   }
 }
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -8,6 +8,7 @@ use core::fmt::Display;
 use core::fmt::Formatter;
 use core::fmt::Result as FmtResult;
 
+use identity_did::did::DID;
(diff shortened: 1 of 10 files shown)


identity-iota/src/did/doc/iota_document.rs
```diff
@@ -113,9 +114,12 @@ impl IotaDocument {
   ///
   /// This must be guaranteed safe by the caller.
   pub unsafe fn from_authentication_unchecked(method: IotaVerificationMethod) -> Self {
+    let verification_method_did: DID = method.id().as_ref().clone();
+
     CoreDocument::builder(Default::default())
       .id(method.controller().clone().into())
-      .authentication(method)
+      .verification_method(method.into())
+      .authentication(MethodRef::Refer(verification_method_did))
```
cycraig, line 122: Should we add a unit test to check the structure of the generated document, i.e. that it now has a single verification method which the authentication section references? The current tests use `authentication()` which resolves the reference and doesn't check for a verification method.

cycraig (approved):
I'm not a big fan of using `int` instead of `integration` since the latter is used throughout the rest of the projects, but I'm also not strongly opposed.

Adding one or two unit tests for the new default document structure may be appropriate. 
Otherwise looks good to me!

identity-iota/src/did/doc/iota_document.rs
PhilippGackstatter: I've added a test that checks if the verification method with the default fragment `#authentication` is the same as what `document.authentication()` resolves to. `IotaDocument` doesn't expose access to the set of `authentication` methods, and I didn't find a way to get a reference to the `CoreDocument` inside. But I think that test should be good enough.

cycraig: I think one can check that the authentication is a reference as follows:

```rust
    match document.document.authentication().first().unwrap().clone().into_inner() {
      MethodRef::Refer(did) => assert_eq!(did.fragment().unwrap_or_default(), "#authentication"),
      MethodRef::Embed(_) => panic!("authentication method should be a reference"),
    }
```

PR #423, opened 2021-09-28
Fix Document constructor Typescript bindings

cycraig:
# Description of change
This replaces the constructor for `WasmDocument` to fix a type mismatch in the generated bindings. Several `Document` constructors are also replaced with a single `new()` function in both Rust and Wasm. This is for consistency and to clarify the canonical method of constructing a new `Document`.

~~The downside is that this removes the ability for a developer to pass in the desired fragment name for the authentication verification method on a newly generated `Document` in Wasm (e.g. instead of "#authentication" one could specify "#customkey123"). This was not possible in Rust, if that behaviour is still desired we should add it in this PR to both languages.~~ 
Edit: added the ability to set the default verification method fragment in both Rust and Wasm.

**Breaking changes**:
- replace IotaDocument `from_keypair()`, `from_keypair_with_network()` with `new()`, `new_with_options()`.
- the default authentication verification method fragment for a new `WasmDocument` is changed from `"#key"` to `"#authentication"`
  - this discrepancy seemed unintentional and now matches the Rust behaviour
- `new Document()` in the Wasm bindings now take a `KeyPair` instead of a `KeyType` and returns just a `Document` instead of a `NewDocument { Document, KeyPair }` struct.
  - BEFORE: `pub fn new(type_: KeyType, network: Option<String>, tag: Option<String>) -> Result<NewDocument>`
  - AFTER:  `pub fn new(key: KeyPair, network: Option<String>, fragment: Option<String>) -> Result<Document>`

Creating a new DID document in Wasm is also now a two-step process, which should be acceptable since this only affects the low-level API and matches the Rust experience.

BEFORE:
```ts
const { doc, key } = new Document(KeyType.Ed25519);
```

AFTER:
```ts
const key = new KeyPair(KeyType.Ed25519);
const doc = new Document(key);
```

**NOTE: this will require updating the repl.it examples in the documentation.**

# Motivation
This is currently a mismatch in the expected type returned by the `WasmDocument` constructor compared to the generated Typescript definition:

```rust
  /// Creates a new DID Document from the given KeyPair.
  #[wasm_bindgen(constructor)]
  #[allow(clippy::new_ret_no_self)]
  pub fn new(type_: KeyType, network: Option<String>, tag: Option<String>) -> Result<NewDocument> {
```

where `NewDocument` is a struct holding the new `WasmDocument` and `KeyPair`:

```rust
#[wasm_bindgen(inspectable)]
pub struct NewDocument {
  key: KeyPair,
  doc: WasmDocument,
}
```

However, the generated Typescript constructor implicitly expects it to return `WasmDocument`, not `NewDocument`:

```ts
export class Document {
  free(): void;
/**
* Creates a new DID Document from the given KeyPair.
* @param {number} type_
* @param {string | undefined} network
* @param {string | undefined} tag
*/
  constructor(type_: number, network?: string, tag?: string);
```

Unfortunately there is no easy way to catch these types of bugs without creating dedicated Typescript tests or converting our examples to Typescript. I tried using `tsc --allowJs --checkJs --noEmit` on the examples but it doesn't link up to the Typescript definitions to catch the type-mismatch.

## Minimal Reproducible Example

The following code snippet fails to compile when using Typescript:

example.ts:
```ts
import { Document, KeyType } from "@iota/identity-wasm/node";
const { doc, key } = new Document(KeyType.Ed25519);
```

Run `tsc example.ts`

Error:
```ts
example.ts:2:13 - error TS2339: Property 'doc' does not exist on type 'Document'.

7     const { doc, key } = new Document(KeyType.Ed25519);
              ~~~

example.ts:2:18 - error TS2339: Property 'key' does not exist on type 'Document'.

7     const { doc, key } = new Document(KeyType.Ed25519);
                   ~~~


Found 2 errors.
```

Note that this example runs fine in Javascript because it doesn't validate the expected type.

## Type of change

- [ ] Documentation Fix

Tests and examples were updated and pass locally. New constructor compiles to Typescript.

## Change checklist

### DIFF
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -72,9 +72,10 @@ impl IotaDocument {
   /// Creates a new DID Document from the given [`KeyPair`] and network, defaulting to
   /// [`Network::Mainnet`](crate::tangle::Network::Mainnet) if unspecified.
   ///
-  /// The DID Document will be pre-populated with a single authentication verification method
-  /// derived from the provided [`KeyPair`]. This method will have the DID URL fragment
-  /// `#authentication` and can be easily retrieved with [`Document::authentication`].
+  /// The DID Document will be pre-populated with a single verification method
+  /// derived from the provided [`KeyPair`], with an attached authentication relationship.
+  /// This method will have the DID URL fragment `#authentication` and can be easily
+  /// retrieved with [`Document::authentication`].
   ///
   /// NOTE: the generated document is unsigned, see [`Document::sign`].
   pub fn new(keypair: &KeyPair, network: Option<&str>) -> Result<Self> {
@@ -516,7 +517,7 @@ impl IotaDocument {
 
   /// Returns the Tangle index of the integration chain for this DID.
   ///
-  /// This corresponds with the tag segment of the [`IotaDID`].
+  /// This is equivalent to the tag segment of the [`IotaDID`].
   ///
   /// E.g.
   /// For an [`IotaDocument`] `doc` with `"did:iota:1234567890abcdefghijklmnopqrstuvxyzABCDEFGHI"`,
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -42,9 +42,9 @@ use crate::did::IotaVerificationMethod;
 use crate::did::Properties as BaseProperties;
 use crate::error::Error;
 use crate::error::Result;
-use crate::tangle::MessageId;
 use crate::tangle::MessageIdExt;
 use crate::tangle::TangleRef;
+use crate::tangle::{MessageId, NetworkName};
 
 type Properties = VerifiableProperties<BaseProperties>;
 type BaseDocument = CoreDocument<Properties, Object, Object>;
@@ -87,7 +87,7 @@ impl IotaDocument {
   /// #
   /// // Create a DID Document from a new Ed25519 keypair.
   /// let keypair = KeyPair::new_ed25519().unwrap();
-  /// let mut document = IotaDocument::new(&keypair).unwrap();
+  /// let document = IotaDocument::new(&keypair).unwrap();
   /// ```
   pub fn new(keypair: &KeyPair) -> Result<Self> {
     Self::new_with_options(keypair, None, None)
@@ -109,18 +109,19 @@ impl IotaDocument {
   /// ```
   /// # use identity_core::crypto::KeyPair;
   /// # use identity_iota::did::IotaDocument;
+  /// # use identity_iota::tangle::Network;
   /// #
   /// // Create a new DID Document for the devnet from a new Ed25519 keypair.
   /// let keypair = KeyPair::new_ed25519().unwrap();
-  /// let mut document = IotaDocument::new_with_options(&keypair, Some("dev"), Some("auth-key")).unwrap();
+  /// let document = IotaDocument::new_with_options(&keypair, Some(Network::Devnet.name()), Some("auth-key")).unwrap();
   /// assert_eq!(document.id().network_str(), "dev");
   /// assert_eq!(document.authentication().try_into_fragment().unwrap(), "#auth-key");
   /// ```
-  pub fn new_with_options(keypair: &KeyPair, network: Option<&str>, fragment: Option<&str>) -> Result<Self> {
+  pub fn new_with_options(keypair: &KeyPair, network: Option<NetworkName>, fragment: Option<&str>) -> Result<Self> {
     let public_key: &PublicKey = keypair.public();
 
-    let did: IotaDID = if let Some(network) = network {
-      IotaDID::new_with_network(public_key.as_ref(), network)?
+    let did: IotaDID = if let Some(network_name) = network {
+      IotaDID::new_with_network(public_key.as_ref(), network_name)?
     } else {
       IotaDID::new(public_key.as_ref())?
     };
@@ -959,8 +960,7 @@ mod tests {
   #[test]
   fn test_new_with_options_network() {
     let keypair: KeyPair = generate_testkey();
-    let document: IotaDocument =
-      IotaDocument::new_with_options(&keypair, Some(Network::Devnet.name_str()), None).unwrap();
+    let document: IotaDocument = IotaDocument::new_with_options(&keypair, Some(Network::Devnet.name()), None).unwrap();
     compare_document_devnet(&document);
   }
 
--- a/identity-iota/src/did/doc/iota_document.rs
+++ b/identity-iota/src/did/doc/iota_document.rs
@@ -8,19 +8,19 @@ use core::fmt::Display;
 use core::fmt::Formatter;
 use core::fmt::Result as FmtResult;
 
-use identity_did::did::DID;
 use serde::Serialize;
 
 use identity_core::common::Object;
 use identity_core::common::Timestamp;
 use identity_core::common::Url;
 use identity_core::convert::SerdeInto;
-use identity_core::crypto::KeyPair;
 use identity_core::crypto::PrivateKey;
 use identity_core::crypto::SetSignature;
 use identity_core::crypto::Signature;
 use identity_core::crypto::TrySignature;
 use identity_core::crypto::TrySignatureMut;
+use identity_core::crypto::{KeyPair, PublicKey};
+use identity_did::did::DID;
 use identity_did::document::CoreDocument;
 use identity_did::service::Service;
 use identity_did::utils::DIDKey;
@@ -70,8 +70,7 @@ impl TryMethod for IotaDocument {
 impl IotaDocument {
   const DEFAULT_METHOD_FRAGMENT: &'static str = "authentication";
 
-  /// Creates a new DID Document from the given [`KeyPair`] and network, defaulting to
-  /// [`Network::Mainnet`](crate::tangle::Network::Mainnet) if unspecified.
+  /// Creates a new DID Document from the given [`KeyPair`].
   ///
   /// The DID Document will be pre-populated with a single verification method
   /// derived from the provided [`KeyPair`], with an attached authentication relationship.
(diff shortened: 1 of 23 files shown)


identity-iota/src/did/doc/iota_document.rs
```diff
@@ -67,38 +67,30 @@ impl TryMethod for IotaDocument {
...
-  /// The authentication method will have the DID URL fragment `#authentication`
-  /// and can be easily retrieved with [Document::authentication].
-  pub fn from_keypair(keypair: &KeyPair) -> Result<Self> {
-    let method: IotaVerificationMethod = IotaVerificationMethod::from_keypair(keypair, "authentication")?;
-
-    // SAFETY: We don't create invalid Methods.  Method::from_keypair() uses the MethodBuilder
-    // internally which verifies correctness on construction.
-    Ok(unsafe { Self::from_authentication_unchecked(method) })
-  }
+  const DEFAULT_METHOD_FRAGMENT: &'static str = "authentication";
```
PhilippGackstatter, line 70: I'm wondering a bit about the discrepancy between what the default account-generated document and the default `IotaDocument` look like. The low-level API uses this `"authentication"` fragment, while the account uses `"_sign-0"` as the default fragment.

https://github.com/iotaledger/identity.rs/blob/31e073e3694e59ac29333568ac8e8049e8eac784/identity-account/src/types/key_location.rs#L23

I don't see why they should not be consistent, and if we want to make them so, I would rather change the account to use something a little easier to type. But with the changes in #421 calling the default verification method `"authentication"` isn't quite right either, because it's a general verification method at that point. Since its main purpose is signing, I would suggest to use `"sign"` as the default in the low-level API, and `"sign-0"` in the account. Any thoughts on this? (This might be out of scope for this PR.)

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -524,10 +516,11 @@ impl IotaDocument {
 
   /// Returns the Tangle index of the integration chain for this DID.
   ///
-  /// This is simply the tag segment of the [`IotaDID`].
+  /// This corresponds with the tag segment of the [`IotaDID`].
```
PhilippGackstatter, line 519:
```suggestion
  /// This is equivalent to the tag segment of the [`IotaDID`].
```

I feel like this would be a bit clearer.

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -67,38 +67,30 @@ impl TryMethod for IotaDocument {
...
 
-  /// Creates a new DID Document from the given KeyPair and network.
+  /// Creates a new DID Document from the given [`KeyPair`] and network, defaulting to
+  /// [`Network::Mainnet`](crate::tangle::Network::Mainnet) if unspecified.
   ///
-  /// The DID Document will be pre-populated with a single authentication
-  /// method based on the provided [KeyPair].
+  /// The DID Document will be pre-populated with a single authentication verification method
+  /// derived from the provided [`KeyPair`]. This method will have the DID URL fragment
+  /// `#authentication` and can be easily retrieved with [`Document::authentication`].
```
PhilippGackstatter, line 77: This doc comment will be outdated after #421 is merged. So something along the lines of this would be better:

```suggestion
  /// The DID Document will be pre-populated with a single verification method
  /// derived from the provided [`KeyPair`], with an attached authentication relationship.
  /// This method will have the DID URL fragment `#authentication` and can be easily
  /// retrieved with [`Document::authentication`].
```

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -67,38 +67,30 @@ impl TryMethod for IotaDocument {
...
-  /// and can be easily retrieved with [Document::authentication].
-  pub fn from_keypair_with_network(keypair: &KeyPair, network: &str) -> Result<Self> {
-    let method: IotaVerificationMethod =
-      IotaVerificationMethod::from_keypair_with_network(keypair, "authentication", network)?;
-
-    // SAFETY: We don't create invalid Methods.  Method::from_keypair() uses the MethodBuilder
-    // internally which verifies correctness on construction.
-    Ok(unsafe { Self::from_authentication_unchecked(method) })
+  /// NOTE: the generated document is unsigned, see [`Document::sign`].
+  pub fn new(keypair: &KeyPair, network: Option<&str>) -> Result<Self> {
```
PhilippGackstatter, line 80: I'm not a huge fan of the optional argument. I think having two methods `new` and `new_with_network` would be preferable. Seems to be more in line with the rest of the library and more idiomatic for Rust. What do you think?

PhilippGackstatter (changes requested):
I really like the decoupling of key generation and document creation. My only concern is the optional argument in `IotaDocument::new` that we should discuss.

identity-iota/src/did/doc/iota_document.rs
cycraig: There's also the default verification method fragment `"key"`:

https://github.com/iotaledger/identity.rs/blob/73f8aa754bc22e7e7a7107f9a622c18c21d1c403/identity-iota/src/did/doc/iota_verification_method.rs#L38

I agree that the `Account` should default to the same name as the low-level API. I particularly dislike how the fragment is used to determine whether or not a verification method is used for authentication in [`KeyLocation`](https://github.com/iotaledger/identity.rs/blob/31e073e3694e59ac29333568ac8e8049e8eac784/identity-account/src/types/key_location.rs#L57):

```
  /// Returns true if the key location points to an authentication method.
  pub fn is_authentication(&self) -> bool {
    Self::is_authentication_fragment(&self.fragment)
  }

  /// Returns true if the fragment points to an authentication method.
  pub fn is_authentication_fragment(fragment: &Fragment) -> bool {
    fragment.name().starts_with(Self::AUTH)         // <============= "_sign-"
  }
```

This could lead to confusion if someone happens to name their key `"_sign-xyx"`? Probably out-of-scope for this PR.

While the default key is used for signing, it's a bit more special than a regular verification method since it can sign document updates too. The two choices are either: choose a special name like `"auth"`/`"default"`/`"bikeshedding"`, or treat it as a regular verification method, like `"key-0"` or `"sign-0"`. Probably worth getting some more opinions on this.

identity-iota/src/did/doc/iota_document.rs
cycraig: I'm glad you brought attention to this, I originally chose a single `new` constructor for a few reasons:

1. to move away from the `from_x` naming scheme to `new`, which is more idiomatic for constructors,
2. for clarity by reducing the number of options a developer has when constructing a `Document`,
3. to make it clear that `network` is important to specify for a new `Document`.

Previously we had `from_keypair`, `from_keypair_with_network` and `from_authentication` (also `from_authentication_unchecked`, `try_from_core`, and `try_from_base` but those are more clearly not the intended constructors). It's far more clear with just `new` and `from_authentication` which the intended method is.

My arguments in favour of `new` and `new_with_network`:

- it is similarly more idiomatic than `from_x` 
- adding new constructor options would be a non-breaking change (e.g. `new_with_network_and_tag` would be a separate function)
- it fits the convention in the code base (or at least aligns with the current `from_x`, `from_x_with_y` scheme)
- creating a new `Document` on the `mainnet` is simpler (`new(key)` vs `new(key, None)`)

My arguments against using `new` and `new_with_network` are:

- many-constructors problem: with just two fields `key` and `network` it's fine, but if we want to add the default `tag`/`fragment` we could introduce `new_with_tag` and `new_with_network_and_tag` too.
- longer constructor names: `new_with_network_and_tag` is probably the upper-limit of what I would tolerate
- nit: brevity in the external API for the low-level API is not as important as the `Account`, particularly if we're intending on pushing towards the `Account` as the default.

For the sake of completeness, the alternatives would be either the `Builder` pattern, or a sort of `Config` pattern where one creates a single configuration struct as input to the `Document` constructor, and that can be extended at-will (but this just "moves" the construction problem to the `Config` struct). I consider the `Builder` pattern overkill for 2-3 fields and a `DocumentConfig` an anti-pattern in this case.

**tl;dr** Overall I am slightly in favour of `new` and `new_with_network`. However, I also think developers should be able to specify the fragment of the initial verification method somehow. So the only real question is how do we handle that?

**Option 1**:

- `new(keypair, network, fragment)`

**Option 2**:

- `new(keypair)`
- `new_with_network(keypair, network, fragment)`

**Option 3**:

- `new(keypair)`
- `new_with_network(keypair, network)`
- `new_with_network_and_fragment(keypair, network, fragment)`

**Option 4**:

Don't allow setting the default verification method fragment on creation. Developers will have to change the fragment for both the verification method and authentication reference, which is a chore but not impossible.

identity-iota/src/did/doc/iota_document.rs
cycraig: The IOTA DID specification recommends the `"_sign-x"` name for verification method used to update the document. Just a reminder to apply whatever the final decision is to the documentation too.

> It is RECOMMENDED to name this public key #_sign-x, where x is the index of the signing key, which is incremented every time the signing key is updated, starting at index 1. 


https://github.com/iotaledger/identity.rs/blob/73f8aa754bc22e7e7a7107f9a622c18c21d1c403/documentation/docs/specs/iota_did_method_spec.md#L102

identity-iota/src/did/doc/iota_document.rs
PhilippGackstatter: Thanks a lot for the write-up! I like the overall change to move away from `from_*` to the more idiomatic `new*`. I think the fragment should be specifiable, since it's the low-level API which should give control over something like that. I also agree that builder and `DocumentConfig` would be overkill.

I feel like option 2 would be a good middle ground. I'm not sure what signature you had in mind for `new_with_network`, but we could make this the constructor with optional arguments. So keypair is required, but both `network` and `fragment` are `Option<&str>`. To reflect in its name that it offers two optional fields, we could go with `new_with_options`, or something like that - can't think of anything better right now. If one wants to create a document with the default network but a custom fragment, then `IotaDocument::new_with_options(keypair, None, Some("fragment"));` seems acceptable.

JelleMillenaar: I like `new_with_options`!

identity-iota/src/did/doc/iota_document.rs
cycraig: Looks like we're going with `"sign-x"`.

identity-iota/src/did/doc/iota_document.rs
cycraig: `new_with_options` is a good suggestion and probably the convention I would use going forward: `new()` with required fields for an easy-to-use default constructor, `new_with_options()` to cover everything else except where we use the `Builder` pattern.

identity-iota/src/did/doc/iota_document.rs
cycraig: Unfortunately verification method names are tightly coupled with the logic in the Account, so it's not as simple as choosing a different name and using any `authentication` verification method. It needs its own issue/PR.

PhilippGackstatter (approved):
Looks great now - thanks for addressing the comments. One thought while reading over `new_with_options` was that we _could_ ask for an `Option<NetworkName>` as the network name. But I suppose the reasoning is that this is the low-level API which shouldn't impose this restriction, and I would agree with that. We can think about changing `IdentityCreate` in the account though, which takes an arbitrary `String` as the network name.

cycraig:
> One thought while reading over `new_with_options` was that we _could_ ask for an `Option<NetworkName>` as the network name. But I suppose the reasoning is that this is the low-level API which shouldn't impose this restriction, and I would agree with that. We can think about changing `IdentityCreate` in the account though, which takes an arbitrary `String` as the network name.

The reason for using `Opton<&str>` was more to do with ergonomics. What I really wanted to use was `Option<impl TryInto<NetworkName>>` as the parameter. The problem I encountered was that using any generic as an optional parameter caused issues with `None` required a type annotation. 

### Example:

```rust
pub fn new_with_options(keypair: &KeyPair, network: Option<impl TryInto<NetworkName>>, fragment: Option<&str>)
```

Trying to use `new_with_options` with `None` as the network parameter causes annoying issues since it requires a concrete type for monomorphisation:

```rust 
let document = IotaDocument::new_with_options(&keypair, None, Some("fragment"))?;
```

```rust
error[E0283]: type annotations needed
   --> identity-iota\src\did\doc\iota_document.rs:977:47
    |
119 |   pub fn new_with_options(keypair: &KeyPair, network: Option<impl TryInto<NetworkName>>, fragment: Option<&str>) -> Result<Self> {
    |                                                                   -------------------- required by this bound in `IotaDocument::new_with_options`
...
977 |     let result: Result<IotaDocument, Error> = IotaDocument::new_with_options(&keypair, None, Some("fragment"));
    |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `impl TryInto<NetworkName>` declared on the associated function `new_with_options`
    |
    = note: cannot satisfy `_: TryInto<network::NetworkName>`
help: consider specifying the type argument in the function call
    |
977 |     let result: Result<IotaDocument, Error> = IotaDocument::new_with_options::<impl TryInto<NetworkName>>(&keypair, None, Some(""));
    |                                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```


Using just a `NetworkName` as the parameter is fine, and it is better than just a `&str` in that it communicates the restriction to the caller at compile time. It is, however, slightly more cumbersome:

E.g. with `Option<NetworkName>`:
```rust
pub fn new_with_options(keypair: &KeyPair, network: Option<NetworkName>, fragment: Option<&str>) -> Result<Self>;

// BEFORE:
// let document = IotaDocument::new_with_options(&keypair, Some("dev"), Some("auth-key"))?;

// AFTER:
let document = IotaDocument::new_with_options(&keypair, Some(Network::Devnet.name()), Some("auth-key"))?;
// OR
let network_name = NetworkName::try_from("dev");
let document = IotaDocument::new_with_options(&keypair, Some(network_name), Some("auth-key"))?;
```

One can also make the same argument to use `NetworkName` for `IotaDID::new_with_network`, which currently uses `&str` too:
```
pub fn new_with_network(public: &[u8], network: &str) -> Result<Self>
```

On a final note, we don't seem to expose `NetworkName` in the Wasm bindings yet, we probably should if only for developers to validate that arbitrary strings are valid IOTA DID network names using our API.

PhilippGackstatter:
I mainly thought of it because you used

```rust
let document = IotaDocument::new_with_options(&keypair, Some(Network::Devnet.as_name_str()), Some("auth-key"))?;
```

before, which isn't less cumbersome than

```rust
let document = IotaDocument::new_with_options(&keypair, Some(Network::Devnet.name()), Some("auth-key"))?;
```

But the main question is if the low-level API should restrict that, or in other words, should the low-level API _enforce_ the IOTA DID method spec? In this particular case, (after thinking more about this) I think it's actually okay to do so (contrary to what I said above).

cycraig:
True, then it really only affects private Tangle network names:

```rust
let document = IotaDocument::new_with_options(&keypair, Some("private"), Some("auth-key"))?;

// vs

let network_name = NetworkName::try_from("private")?;
let document = IotaDocument::new_with_options(&keypair, Some(network_name), Some("auth-key"))?;
```

cycraig:
@PhilippGackstatter I changed the parameters for `IotaDocument::new_with_options()`, `IotaDID::new_with_network()` and `IdentityCreate::network()` to use `NetworkName` instead of `&str`. However, I used `impl TryInto<NetworkName>` so developers can still pass (static/owned) strings if they so choose. During the refactor I started to feel like `NetworkName` should be an internal construct, since it's so similar to `Network` and is a bit confusing to deal with in multiple places. I'm not entirely happy with this last change since it's a bit less clean and doesn't translate well to Wasm, but the one benefit is that `IdentityCreate::network` is now fallible, so invalid `NetworkNames` are caught earlier.

PhilippGackstatter (approved):
Excellent improvements!

Issue #408, opened 2021-09-20
[Task] Separate `DID` from `DIDUrl`


cycraig:
## Description
Split `CoreDID` and `IotaDID` into separate `DID` and `DIDUrl` structs.

E.g.:

- `DID`: `"did:iota:main:9rK6DPF46MCEzgfLD8AHFsaTuMqvmRo6kbXfjqQJPJmC"`
- `DIDUrl`: `"did:iota:main:9rK6DPF46MCEzgfLD8AHFsaTuMqvmRo6kbXfjqQJPJmC#keys?query/path`

![image](https://user-images.githubusercontent.com/5772887/133995369-4e51b598-32fd-4849-9a98-94039d9a4703.png)

## Motivation
We currently have two `DID` structs: `CoreDID` and `IotaDID`:

- `CoreDID` is a re-export of the [`did-url` crate](https://crates.io/crates/did_url), which is optimised to  represent a full `DIDUrl`, potentially including a fragment, query, and/or path.
- `IotaDID` just wraps `CoreDID` to enforce the [IOTA DID Method specification](https://identity.docs.iota.org/docs/specs/iota_did_method_spec).

The discussion in #387 makes apparent that we need some way to distinguish a `DID` from a `DIDUrl`. Specifically to enforce invariants in structs and improve clarity in function parameters.

For example:

- a `CoreDocument` uses `CoreDID` (`DIDUrl`) for the `id` and `controller` fields, which should [conform to the base `DID` syntax according to the did-core specification](https://www.w3.org/TR/did-core/#did-document-properties), but currently theoretically allows any `DIDUrl`.
- `Client::resolve` takes an `IotaDID` (`DIDUrl`) parameter but ignores any URL segments and just uses the base `DID`.
- the `Account` defines an `IdentityTag` which represents just the network and tag of a `DID` for mapping internally.

Note that any solution should translate well to the external Wasm bindings API, which does not support generics or interfaces properly.

NOTE: any invariants should be enforced when deserialising these structures.

## Resources
`DID`: https://www.w3.org/TR/did-core/#did-syntax
`DIDUrl`: https://www.w3.org/TR/did-core/#did-url-syntax

## To-do list
Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Implement `CoreDID` and `CoreDIDUrl`
- [ ] Implement `IotaDID` and `IotaDIDUrl`
[...]

PR #428, opened 2021-10-05
Separate `DID`, `DIDUrl` constructs

cycraig:
# WORK-IN-PROGRESS

# Description of change
This PR replaces `CoreDID` and `IotaDID` (both of which represented a [DID Url](https://www.w3.org/TR/did-core/#did-url-syntax) rather than a base [DID](https://www.w3.org/TR/did-core/#did-syntax)) with several constructs:

- `DID` trait: defines shared behaviour for all DID structs,
- `CoreDID` struct: validates and represents a general base DID from any method,
- `IotaDID` struct: wraps `CoreDID` and enforces invariants from the [IOTA DID Method Specification](https://identity.docs.iota.org/docs/specs/iota_did_method_spec/),
- `RelativeDIDUrl` struct: holds the path, query, and fragment of a DID Url,
- `DIDUrl<T: DID>` struct: generic DID Url which holds some `DID`,
- `CoreDIDUrl`: type alias for `DIDUrl<CoreDID>`,
- `IotaDIDUrl`: type alias for `DIDUrl<IotaDID>`,

The overall intention is to:
- enforce invariants for struct fields that should only ever contain a DID rather than a DID Url, and
- improve our API by clarifying when function parameters should only be a DID.

There are some disadvantages of this approach:
- TODO (performance, more string allocations, lack of `as_str()` on `DIDUrl`, choice of generic struct over trait for `DIDUrl` causing problems with traits `from`, `try_from` due to lack of specialisation/negative trait bounds)

The following structs have been updated to use `CoreDID`/`IotaDID` rather than a `DIDUrl` as before:
- TODO

Fixes issue #408

## Type of change
- [ ] Documentation Fix

TODO

## Change checklist


cycraig:
> General question: Is there a reason we're not following the naming convention with DIDs, and use IotaDID instead of IotaDid? The latter would be my preference.

It just matches the specification and the current convention in the codebase: `DIDKey`, `WasmDID`, `DIDResult` etc. I slightly prefer `DID` over `Did` but wouldn't strongly oppose changing it.

If we choose to change the convention can we do that in a separate PR from this?

Issue #433, opened 2021-10-08
[Task] Refactor the Account state


JelleMillenaar:
Note: This issue used to be about low-level API access, but this task has moved to #475.

## Description

Refactor the `Account` state to enable lower-level access (#475) to the document stored inside. This should refactor the account snapshots with its event system to use `IotaDocument`s to represent the state of an identity. The stored document represent the last resolved tangle state of the document. When an account is created, this document is loaded and any updates to the identity are applied to that in-memory document.

Whether these updates are written into storage or stay temporary until published needs to be figured out. Doing so would enable updates to be applied to the identity locally, and publish them at a later point. The downside is the higher complexity involved, since the low-level access to the document isn't explicit to the account, so it cannot write them into storage immediately. Some caveats may apply in that case.

Finally, the refactor should introduce an abstraction that allows for easier implementation of future did methods.

## Motivation

Simplify the internal state representation and take advantage of the account only having to manage one identity now.

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Remove `IdentitySnapshot` and the event/commit system
- [ ] Upgrade `IotaDocument` with any better APIs from `IdentityState`
- [ ] Replace the internal state with an `IotaDocument` and apply all updates to that document
- [ ] Abstract identity management into an interface

PR #436, opened 2021-10-12
Refactor `Account` to handle just one identity

### LINKED ISSUE #431: [Task] Convert Account to handle one Identity
## Description
Describe the task, this may be scoped beyond a single PR.

The `Account` has been used to handle multiple identities and therefore has some complex logic. This is more application level behavior and therefore is not the responsibility of this framework. The challenge for this PR will be around cleaning up the  account configurations as these can be shared between multiple accounts. Similarly the `Storage` mechanism needs to be sharable between multiple accounts by using something like `Arc<dyn Storage>`, meaning the account doesn't control a `Storage` itself, but rather a reference to it. 

In addition, the PR should tackle and question the need for a `Config`, `State` (which has the `ClientMap`) and a `store` (which is the `Storage`) and combine them into something more clear and easier to handle across multiple accounts. Lastly, it should also be checked that two different accounts can use different configurations, but share a storage mechanism. For example Account A uses Config A, Account B copies Config A into Config B, and changes the `ClientMap`, in this case the Storage would be the same. 

This Issue does NOT have to take care of any `factory`, `builder` or whatever pattern we decide for to create multiple accounts, but needs to prepare for the future existence of such a feature by having these configurations comply to the above requirements. 

## Motivation

Cleaner Account API and better maintainable code. 

## Resources

- [Github Discussion](https://github.com/iotaledger/identity.rs/discussions/396).

## To-do list

- [x] Remove IdentityIndex
- [x] Improve API and implementation to utilize that the `Account` now only has one identity
- [x] Update `IdentityUpdater` to no longer require an identifier in its updating process.
- [x] Clean up `configurations`
- [x] Make sure multiple `Accounts` can share the same `Storage`.
- [x] Remove the `find_identity`, `list_identities` and `resolve_identity` methods.

PhilippGackstatter:
# Description of change

This PR implements #431 and the builder pattern to create accounts. 

The account was converted to handle only one identity instead of multiple ones. The new API looks like this:

```rust
#[tokio::main]
  let password: String = "my-password".into();
    
  let mut builder: AccountBuilder = Account::builder()
    .storage(AccountStorage::Stronghold(stronghold_path, Some(password)))
    .dropsave(true)
    .await?;

  let account: Account = builder
    .create_identity(IdentitySetup::default().network("dev")?)
    .build()
    .await?;

  // To show how existing identities are loaded into accounts
  let mut account = builder.load_identity(account.did().to_owned()).await?;
    
  // Requires &mut self to disallow concurrent updates on the same identity
  account
    .update_identity()
    .create_method()
    .fragment("new-method")
    .apply()
    .await?;

  // Consumes the account
  account.delete_identity().await?;

  Ok(())
}
```

A constructor API on the account is also available. This was required anyway, for the builder to call, and we can decide if we want to make it public or not. To make that ergonomic, a wrapper around `Config` and `Storage` was created, which is called `AccountSetup` (name debatable). The idea of simply merging the storage into the config seemed like coupling config and storage too much together. Now it's still possible to create the two types separately and the setup is just a thin wrapper. They can also continue to be two different fields in the account which seems cleaner. An example of the constructor API:

```rust
// Arc required, so users can decide explicitly which setups share a storage
// I found it to be okay if this API is more powerful in exchange for slightly more complexity
// since the ergonomic builder API exists, too.
let setup = AccountSetup::new_with_options(Arc::new(MemStore::new()), Some(Config::new().autopublish(false)), None);

let account = Account::create_identity(setup, IdentityCreate::new()).await?;

let did: IotaDID = "did:iota:...".parse().unwrap();
let account = Account::load_identity(setup, did).await?;
```

The definition of `AccountSetup` is:

```rust
pub struct AccountSetup {
  pub(crate) config: Config,
  pub(crate) storage: Arc<dyn Storage>,
  pub(crate) client_map: Arc<ClientMap>,
}
```

Since a builder is used to potentially create numerous accounts, it seemed reasonable to let accounts share `ClientMap`s, so that ten accounts don't have to instantiate ten `Client`s. The builder clones the setup before passing it to the account. That means `Config` is not shared between accounts, so developers can reconfigure the builder and produce another account instance with a different, non-shared configuration.

Couple of other notes:

- The current state representation still basically lives in the storage and is fetched and written back every time an update is executed. The account only needs to hold the did, and not the `IotaDocument` or the `IdentityState` directly. That (potential) change is considered out of scope since this PR is already of medium size, and we have not come to a conclusion on how to proceed with the [state representation](https://github.com/iotaledger/identity.rs/issues/433#issuecomment-942194667).
- The `Identity{Index, Id, Name, Key, Tag}` were removed entirely.
- Changed `update_identity` to require `&mut self` rather than `&self`. This ensures at compile-time that concurrent updates to the same identity are not possible, except by creating multiple accounts with the same did. Whether we should prevent the latter is debatable. Concurrent updates can potentially overwrite each other, which is why we added the locks to the `IdentityIndex` in the past. With the index gone, the proper solution seems to be to let the update method take `&mut self`.
- "Flattened" the `Account` struct fields to make it easier to see at a glance where things are stored (i.e. `State` was removed, and `ClientMap` and `actions` live directly in the `Account`).
- Renamed `store` to `storage` to be more consistent with the `Storage` trait naming.
- Pulled `CreateIdentity` out of the `Command` enum, since commands operate on an existing identity, whereas create identity obviously doesn't. This also means the `Commands` are now exclusively updates to an identity and thus the type was renamed to `Update`, in line with the earlier change of renaming `CommandError` to `UpdateError`.
  - Due to this change, `commit_events` had to be split into two methods, so it can be used without an account instance during instantiation.
- The initial `KeyPair` for the creation of an identity is no longer generated in stronghold but in our library instead. This is hopefully only a temporary solution until stronghold gains the ability to move a secret. The reason for this change was that we no longer identify things in storage by `IdentityId` but by the did itself. Since the did cannot be generated until the keypair is generated, but the did is the location of the keypair, this solution was a way to break the circular dependency.

Final note: The `identity-account/src/tests/commands.rs` file is supposed to be named `updates`, but git treated it as a new file, so I renamed it back to make it easier to review. Will rename it before merging.

Final note, for real: Note that this PR will be merged into an epic branch, which will see additional PRs merged into it, before the epic itself is merged into dev.

fixes #431

## Type of change

- [x] (Very) Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

Refactored `identity-account/src/tests/commands.rs` to use the new API, and added another test in `identity-account/src/tests/account.rs`.

## Change checklist

### DIFF
--- a/identity-account/src/account/account.rs
+++ b/identity-account/src/account/account.rs
@@ -131,7 +131,7 @@ impl Account {
     Ok(self.load_snapshot().await?.into_identity())
   }
 
-  /// Resolves the DID Document associated with the specified `key`.
+  /// Resolves the DID Document associated with this `Account` from the Tangle.
   pub async fn resolve_identity(&self) -> Result<IotaDocument> {
     let snapshot: IdentitySnapshot = self.load_snapshot().await?;
     let did: &IotaDID = snapshot.identity().try_did()?;


PhilippGackstatter:
I've implemented the `IdentityBuilder` now, but want to point out that this makes the constructor route a bit less clean, since we need to duplicate the same builder pattern across `IdentityBuilder` and `IdentityCreate`. However, as requested, I've made the constructors private, so this only matters a bit for our own internal sanity.

I've also implemented the requested mechanism to prevent two accounts from managing the same did, which is relatively lightweight.

I'm basically done and think I've responded to every comment made. I still need to check if the `Drop` impl of `Account`'s deadlocks, which I fear it does (uses a similar setup as #441) and might present a big blocker for this PR, if we want to keep `dropsave`, which we probably do. The stuff mentioned above can be reviewed, though, so I'll request another review.

PR #455, opened 2021-10-25
Add publish with retry method

### LINKED ISSUE #426: [Task] Add Tangle Promotion / Reattach logic for IOTA Messages
## Description

When any message is published to the Tangle, the function returns success if the message has successfully been pushed into the network. This does however not provide any guarantees that messages will be confirmed. Without a confirmation a message is not immutably stored on the Tangle as we have no guarantee that all nodes have observed the message, nor will it be stored in the permanodes. Therefore messages are either `promoted` or `reattached` in order to make sure they get confirmed. `Promotion` adds a new message to the Tangle with no `payload` (and therefore little PoW) that references our original DID message, increasing the odds it will be picked up by consensus. `Reattaching` publishes the message with the original `payload` again to the Tangle, this is done if the original message has become to old to be considered for consensus (also known as "below max-depth"). The Client library already has a very helpful `retry_until_included` function that we can utilize. 

Since Chrysalis Phase 2, consensus has been VERY reliable and most messages get confirmed without promotion or reattachment, however for 1.0 we need to provide guarantees. In addition, we need to update the API in order for an application to know if a message fails to be included even after retries, such as during a time that the coordinator is offline. This will probably be triggered after several MINUTES! So we need to be careful about that. 

## Motivation

For a 1.0 release, we need to provide guarantees that a message is included in the Tangle.

## Resources

* `Retry_until_included` from IOTA.rs

## To-do list

- [ ] Implement `retry_until_included` into our publishing logic
- [ ] Provide a way for library consumers to be notified if confirmation fails

HenriqueNogara:
# Description of change
Adds a method for promoting/reattaching messages sent to the tangle.

closes #426 

## Type of change

- [ ] Documentation Fix

## Change checklist


olivereanderson (approved):

PhilippGackstatter (changes requested):

identity-iota/src/tangle/client.rs
```diff
@@ -97,6 +101,41 @@ impl Client {
...
+    index: &str,
+    data: &T,
+    interval: Option<u64>,
+    max_attempts: Option<u64>,
+  ) -> Result<Receipt> {
+    let receipt = self.publish_json(index, data).await?;
+    let reattached_messages = self
+      .retry_until_included(receipt.message_id(), interval, max_attempts)
+      .await?;
+    match reattached_messages.last() {
```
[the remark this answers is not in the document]
HenriqueNogara: As we discussed, I'll put this on hold until this is changed in iota.rs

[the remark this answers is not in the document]
cycraig: Looks like it was merged? https://github.com/iotaledger/iota.rs/pull/730

bindings/wasm/Cargo.toml
```diff
@@ -18,7 +18,7 @@ crate-type = ["cdylib", "rlib"]
 console_error_panic_hook = { version = "0.1" }
 futures = { version = "0.3" }
 js-sys = { version = "0.3" }
-reqwest = { version = "=0.11.4" } # pin version to 0.11.4: https://github.com/iotaledger/identity.rs/issues/438
+reqwest = { version = "0.11.6" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", default-features = false }
 tokio = { version = "*", default-features = false, features = ["rt"] } # enable "rt" feature to fix build: https://github.com/iotaledger/identity.rs/issues/403
```
cycraig, line 24: I think this can also be removed with the latest `iota-client` version.

bindings/wasm/Cargo.toml
```diff
@@ -37,6 +36,7 @@ wasm-bindgen-test = { version = "0.3" }
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 chrono = { version = "0.4", features = ["wasmbind"] }
 getrandom = { version = "0.2", features = ["js"] }
+parking_lot = { version = "0.11.2", features = ["wasm-bindgen"] }
```
cycraig, line 39: nit: no need to specify the patch, will be overridden anyway. I wonder if these versions can be specified as `"*"` with `default-features = false`, similar to the former tokio workaround?

```suggestion
parking_lot = { version = "0.11", features = ["wasm-bindgen"] }
```

identity-iota/src/tangle/client.rs
```diff
@@ -97,6 +105,27 @@ impl Client {
...
+    index: &str,
+    data: &T,
+    interval: Option<u64>,
+    max_attempts: Option<u64>,
+  ) -> Result<Receipt> {
+    let receipt = self.publish_json(index, data).await?;
+    let reattached_messages = self
+      .client
+      .retry_until_included(receipt.message_id(), interval, max_attempts)
+      .await?;
```
cycraig, line 122: nit: very minor but the codebase convention is usually to annotate the type for new variables. Not sure what everyone's opinion on this is but it's the current convention.

E.g.
```rust
    let receipt: Receipt = self.publish_json(index, data).await?;
    let reattached_messages: Vec<MessageWrap (or whatever this returns)> = self
      .client
      .retry_until_included(receipt.message_id(), interval, max_attempts)
      .await?;
```

cycraig (approved):
Looks good to me!

identity-iota/src/tangle/client.rs
```diff
@@ -97,6 +105,27 @@ impl Client {
...
+    match reattached_messages.first() {
+      Some((_, message)) => Ok(Receipt::new(self.network.clone(), message.clone())),
+      None => Ok(receipt),
+    }
```
cycraig, line 126: Missed this initially: do we really need to clone the entire message which includes the payload, can we extract it from the list instead?

cycraig: E.g.

```suggestion
    match reattached_messages.into_iter().next() {
      Some((_, message)) => Ok(Receipt::new(self.network.clone(), message)),
      None => Ok(receipt),
    }
```

identity-iota/src/tangle/client.rs
HenriqueNogara: I was not aware of this convention, but I kind of like it to be honest. I guess it makes it easier to figure things out when looking at code directly from GitHub.

identity-iota/src/tangle/client.rs
```diff
@@ -97,6 +105,27 @@ impl Client {
...
+  ) -> Result<Receipt> {
+    let receipt: Receipt = self.publish_json(index, data).await?;
+    let reattached_messages: Vec<(MessageId, Message)> = self
+      .client
+      .retry_until_included(receipt.message_id(), interval, max_attempts)
+      .await?;
+    match reattached_messages.into_iter().next() {
+      Some((_, message)) => Ok(Receipt::new(self.network.clone(), message)),
+      None => Ok(receipt),
+    }
```
PhilippGackstatter, line 126: Do we not have to handle the `Error::TangleInclusionError` that can be returned from `retry_until_included`? I'm not entirely sure if that method can still return this error, even if `publish_json` successfully included our message.

If that error is returned, we may have to do another call to check the `LedgerInclusionStateDto` of the message, as it is done [here](https://github.com/iotaledger/iota.rs/tree/b9dbaa735963e88a061f44bc5fdce6b0fe8463e9/iota-client/src/client.rs#L1341-L1344), but I haven't looked too deep into the logic of `retry_until_included`.

Also, if it is guaranteed that `retry_until_included` returns the included message on `Ok`, then we technically don't have to handle the case where `reattached_messages.into_iter().next()` returns `None`, although we also shouldn't panic, so mapping it to some error seems like the best we can do.

PhilippGackstatter: Talked to Thoralf about this.

- It's practically guaranteed that the returned vector is non-empty, but we should handle this anyway and map it to some reasonable error. 
- And the retry function can return a `TangleInclusionError`, but the message might have been included with `publish_json` already, so I think it'd be good to add the additional inclusion check.

identity-iota/src/tangle/client.rs
```diff
@@ -101,6 +111,39 @@ impl Client {
...
+          }
+          return Err(Error::from(e));
+        }
+        _ => return Err(Error::from(e)),
+      },
+    };
+    match reattached_messages.into_iter().next() {
+      Some((_, message)) => Ok(Receipt::new(self.network.clone(), message)),
+      None => Err(Error::MessageInclusionNotChecked),
+    }
```
PhilippGackstatter, line 144:
```suggestion
    let receipt: Receipt = self.publish_json(index, data).await?;
    let retry_result = self
      .client
      .retry_until_included(receipt.message_id(), interval, max_attempts)
      .await;

    let reattached_messages: Vec<(MessageId, Message)> = match retry_result {
      Ok(reattached_messages) => reattached_messages,
      Err(inclusion_error @ IotaClientError::TangleInclusionError(_)) => {
        if self.is_message_included(receipt.message_id()).await? {
          return Ok(receipt);
        } else {
          return Err(Error::from(inclusion_error));
        }
      }
      Err(error) => {
        return Err(Error::from(error));
      }
    };

    match reattached_messages.into_iter().next() {
      Some((_, message)) => Ok(Receipt::new(self.network.clone(), message)),
      None => Err(Error::from(IotaClientError::TangleInclusionError(
        receipt.message_id().to_string(),
      ))),
    }
```

Maybe restructure the section a bit:
- Untangle the match statement as much as possible, since that is fairly complex
- I don't think we need a custom error, we can return `TangleInclusionError`, since `MessageInclusionNotChecked` ultimately means the same to the user, i.e. they cannot be sure the message was included.

PhilippGackstatter (approved):
Thanks for the changes. Just some restructuring and not sure whether we need the new error. Other than that, looks good now!

cycraig:
The timeout for the Wasm examples/tests should be increased to account for the new confirmation delay on publishing.

https://github.com/iotaledger/identity.rs/blob/0d332152c4c7db0baad7657dd6e395082d6ee64a/bindings/wasm/cypress/integration/browser_test.js#L9

https://github.com/iotaledger/identity.rs/blob/0d332152c4c7db0baad7657dd6e395082d6ee64a/bindings/wasm/examples/src/test.js#L13

Issue #434, opened 2021-10-11
[Task] Research (and impl) DID Message Compression to reduce Message size


JelleMillenaar:
## Description

As of right now, DID Messages are published as a JSON formatted document containing (parts of a) DID Document and metadata. As the size of messages on the Tangle, directly influences the Proof-of-Work requirement, it is important to reduce the amount of data on the Tangle, while not losing any information. You might call something like that lossless compression 😄 

Two things need to be considered for compression and may be combined if it achieves the best outcome:
* Lossless Compression Algorithms
* Handmade Compression 

A handmade compression could be a "conversion map" (Not sure if this is a thing), where we map frequently used fields to smaller, often single, characters. DID Messages contain valid JSON, but technically speaking we could map `"verificationMethod":` to `v:`. This may already be caught by the lossless compression algorithm, but I expect a combination to net the best results. 

Any compression will remove the human-readability from the Tangle messages. It will be important to decompress messages in an early stage of the processing, such that no tools are broken (Identity Resolver). 

It was already the idea to add version to the DID messages in order to have the ability to support backwards compatibility. This would need to be a field outside of the compression as a new version could also indicate a new compression / decompression algorithm in the future. 

## Motivation

Reduce DID Messages size, therefore reducing the Proof-of-Work requirement for IOTA Identity, increasing the speed of the framework's writing ability and reduce the carbon footprint. 

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Compare Lossless Compression Algorithms
- [ ] Implement Handmade Compression
- [ ] Compare in combination with Handmade Compression
- [ ] Introduce DID Message version outside of compressed region

abdulmth:
### Comparing Lossless Compression Algorithms

For compressing a DID message, the following algorithms has been tested: 

* Bzip2: uses file extension `.bz2`, initial release in 1996, usually more effective than deflate.
* zlib: abstraction of the DEFLATE compression algorithm, initial release in 1995.
* Snappy, aims for speed more than maximum compression, developed by Google, initial release 2011.
* Brotli: works best for text compression, developed by Google, initial release 2013.

[...]

PR #466, opened 2021-10-30
Add message compression and versioning

abdulmth:
# Description of change

- Messages are compressed with brotli before saving them to the tangle, and decompressed before getting parsed. 
- Introducing message version and encoding version. 

The format of a message on the tangle is as follows:
`$message_version $encoding_version $data`

where
- `$message_version` is a one byte flag indicating the message version (1 for the current, not-modified message).
- `$encoding_version` is a one byte flag indicating the encoding version (1 for json + brotli).
- `$data` is the compressed data.

`$message_version` can be used in the future to indicate the format that follows the first byte.

implements issue #434 

## Type of change

- [ ] Documentation Fix

Unit tests, manual tests, wasm bindings tests.  

## Change checklist


abdulmth:
we might need to check if the license of Brotli is compatible with the library
https://github.com/dropbox/rust-brotli/blob/master/LICENSE

identity-iota/src/error.rs
```diff
@@ -57,4 +57,8 @@ pub enum Error {
   NoExplorerURLSet,
   #[error("Invalid Explorer Url")]
   InvalidExplorerURL,
+  #[error("compression Error")]
+  CompressionError,
+  #[error("Invalid Message Flags")]
```
olivereanderson, line 62: Nit: Error messages should be lower case (yes I know we are currently doing this wrong). Hence it should be `#[error("invalid message flags")]`, same goes for `#[error("compression error")]` above.

identity-iota/src/tangle/message_version.rs
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::Error;
+
+#[derive(Copy, Clone)]
+pub(crate) enum MessageVersion {
+  V1 = 1,
+}
+static CURRENT_MESSAGE_VERSION: MessageVersion = MessageVersion::V1;
```
olivereanderson, line 10: Consider using `const` instead of `static`.

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::tangle::compression_brotli::compress_brotli;
+use crate::tangle::compression_brotli::decompress_brotli;
+use crate::Error;
+
+#[derive(Copy, Clone)]
+pub(crate) enum MessageEncodingVersion {
+  JsonBrotli = 1,
+}
+
+static CURRENT_ENCODING_VERSION: MessageEncodingVersion = MessageEncodingVersion::JsonBrotli;
```
olivereanderson, line 13: consider using `const` instead of `static`

olivereanderson:
> we might need to check if the license of Brotli is compatible with the library https://github.com/dropbox/rust-brotli/blob/master/LICENSE

It looks like it is licensed under the BSD 3-Clause. According to the accepted answer here: https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache
it should be compatible with the Apache 2.0 License, but we might have to double check.

identity-iota/Cargo.toml
```diff
@@ -12,6 +12,7 @@ description = "An IOTA Tangle intergration for the identity-rs library."
 
 [dependencies]
 async-trait = { version = "0.1", default-features = false }
+brotli = "3.3.2"
```
cycraig, line 15: nit: disable unused features (we don't need `ffi-api`).

```suggestion
brotli = { version = "3.3", default-features = false, features = ["std"] }
```

identity-iota/src/tangle/compression_brotli.rs
```diff
@@ -0,0 +1,44 @@
...
+#[cfg(test)]
+mod test {
+  use crate::did::IotaDocument;
+  use crate::tangle::compression_brotli::compress_brotli;
+  use crate::tangle::compression_brotli::decompress_brotli;
+  use identity_core::convert::ToJson;
+  use identity_core::crypto::KeyPair;
+
+  #[test]
+  fn test_compression_algorithm() {
```
cycraig, line 33:
```suggestion
  fn test_brotli() {
```

olivereanderson:
Would it maybe make more sense to move `compression_brotli.rs` into the `identity-core` crate?

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::tangle::compression_brotli::compress_brotli;
+use crate::tangle::compression_brotli::decompress_brotli;
+use crate::Error;
+
+#[derive(Copy, Clone)]
+pub(crate) enum MessageEncodingVersion {
+  JsonBrotli = 1,
+}
```
cycraig, line 11: Can we add an uncompressed/JSON-only encoding for debugging too?

```suggestion
pub(crate) enum MessageEncodingVersion {
  Json = 0,
  JsonBrotli = 1,
}
```

identity-iota/src/tangle/compression_brotli.rs
```diff
@@ -0,0 +1,44 @@
...
+  let mut result = Vec::new();
+  let mut compressor = brotli::CompressorReader::new(input.as_ref(), 4096, 5, 22);
+  compressor
+    .read_to_end(&mut result)
+    .map_err(|_| Error::CompressionError)?;
+  Ok(result)
+}
+
+pub(crate) fn decompress_brotli<T: AsRef<[u8]> + ?Sized>(input: &T) -> Result<String, Error> {
+  let mut decompressor = brotli::Decompressor::new(input.as_ref(), 4096 /* buffer size */);
```
cycraig, line 18: Please declare constants for magic numbers like buffer sizes at the top of the file.

E.g.
```rust
const BUFFER_SIZE: usize = 4096;
```

cycraig:
> Would it maybe make more sense to move compression_brotli.rs into the identity-core crate?

It's specific to IOTA messages.

identity-iota/src/tangle/compression_brotli.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::error::Error;
+use crate::error::Error::CompressionError;
+use std::io::prelude::*;
+
+pub(crate) fn compress_brotli<T: AsRef<[u8]>>(input: T) -> Result<Vec<u8>, Error> {
+  let mut result = Vec::new();
+  let mut compressor = brotli::CompressorReader::new(input.as_ref(), 4096, 5, 22);
```
cycraig, line 10: Same suggestion: please declare constants for magic numbers at the top of the file to know what they are.

E.g.
```rust
const BUFFER_SIZE: usize = 4096;
const QUALITY: u32 = 5; // compression level
const WINDOWS_SIZE: u32 = 22; // LZ77 window size
```

identity-iota/src/tangle/message_version.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/// Checks if flag matches message_version.
+pub(crate) fn check_version_flag(flag: &u8, message_version: MessageVersion) -> Result<(), Error> {
+  if message_version as u8 == *flag {
+    Ok(())
+  } else {
+    Err(Error::InvalidMessageFlags)
+  }
+}
```
cycraig, line 26: nit: my personal preference is to declare associated functions rather than naked functions.

E.g.
```rust
impl MessageVersion {
  /// Adds the current message version flag at the beginning of arbitrary data.
  pub(crate) fn add_version_flag(mut data: Vec<u8>) -> Vec<u8> {
    let version_flag = CURRENT_MESSAGE_VERSION as u8;
    data.splice(0..0, [version_flag].iter().cloned());
    data
  }
  
  /// Checks if flag matches message_version.
  pub(crate) fn check_version_flag(flag: &u8, message_version: MessageVersion) -> Result<(), Error> {
    if message_version as u8 == *flag {
      Ok(())
    } else {
      Err(Error::InvalidMessageFlags)
    }
  }
}
```

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+    Err(Error::InvalidMessageFlags)
+  };
+}
+
+/// Compresses a message depending on current encoding version.
+pub(crate) fn compress_message<T: AsRef<[u8]>>(message: T) -> Result<Vec<u8>, Error> {
+  match CURRENT_ENCODING_VERSION {
+    MessageEncodingVersion::JsonBrotli => compress_brotli(message),
+  }
+}
```
cycraig, line 37: nit: my personal preference is to declare associated functions rather than naked functions.

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::tangle::compression_brotli::compress_brotli;
+use crate::tangle::compression_brotli::decompress_brotli;
+use crate::Error;
+
+#[derive(Copy, Clone)]
+pub(crate) enum MessageEncodingVersion {
```
cycraig, line 9: nit: can we rename this `MessageEncoding` instead?

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+  return if *encoding_flag == MessageEncodingVersion::JsonBrotli as u8 {
+    decompress_brotli(data.as_ref())
+  } else {
+    Err(Error::InvalidMessageFlags)
+  };
+}
+
+/// Compresses a message depending on current encoding version.
+pub(crate) fn compress_message<T: AsRef<[u8]>>(message: T) -> Result<Vec<u8>, Error> {
+  match CURRENT_ENCODING_VERSION {
```
cycraig, line 34: This should take the `MessageEncoding` as an input parameter rather than defaulting to the current one; the caller of the function should decide.

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+#[derive(Copy, Clone)]
+pub(crate) enum MessageEncodingVersion {
+  JsonBrotli = 1,
+}
+
+const CURRENT_ENCODING_VERSION: MessageEncodingVersion = MessageEncodingVersion::JsonBrotli;
+
+/// Adds the current encoding flag at the beginning of arbitrary data.
+pub(crate) fn add_encoding_version_flag(mut data: Vec<u8>) -> Vec<u8> {
+  let encoding_version = CURRENT_ENCODING_VERSION as u8;
```
cycraig, line 17: Same comment: this should take the `MessageEncoding` as an input parameter rather than defaulting to the current one; the caller of the function should decide.

cycraig (changes requested):
Looks good overall. 

Mostly nits and questions from my side. We need to decide whether or not to support uncompressed messages at all, for instance. We might want to move around some of the files since the `tangle` folder is getting a bit crowded.

identity-iota/src/tangle/message_version.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[test]
+fn test_add_version_flag() {
+  let message: Vec<u8> = vec![10, 4, 5, 5];
+  let message_with_flag = add_version_flag(message);
+  assert_eq!(message_with_flag, [CURRENT_MESSAGE_VERSION as u8, 10, 4, 5, 5])
+}
```
cycraig, line 33:
```rust
#[cfg(test)]
mod tests {
  #[test]
  fn test_add_version_flag() {
    let message: Vec<u8> = vec![10, 4, 5, 5];
    let message_with_flag = add_version_flag(message);
    assert_eq!(message_with_flag, [CURRENT_MESSAGE_VERSION as u8, 10, 4, 5, 5])
  }
}
```

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+    MessageEncodingVersion::JsonBrotli => compress_brotli(message),
+  }
+}
+
+#[test]
+fn test_add_version_flag() {
+  let message: Vec<u8> = vec![10, 4, 5, 5];
+  let message_with_flag = add_encoding_version_flag(message);
+  assert_eq!(message_with_flag, [CURRENT_ENCODING_VERSION as u8, 10, 4, 5, 5])
+}
```
cycraig, line 44:
```rust
#[cfg(test)]
mod tests {
  #[test]
  fn test_add_version_flag() {
    let message: Vec<u8> = vec![10, 4, 5, 5];
    let message_with_flag = add_encoding_version_flag(message);
    assert_eq!(message_with_flag, [CURRENT_ENCODING_VERSION as u8, 10, 4, 5, 5])
  }
}
```

cycraig (commented):
Oh, forgot to add these comments: tests should be wrapped in a tests module to prevent compiling them by default.

cycraig:
>> we might need to check if the license of Brotli is compatible with the library https://github.com/dropbox/rust-brotli/blob/master/LICENSE
>
> It looks like it is licensed under the BSD 3-Clause. According to the accepted answer here: https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache
it should be compatible with the Apache 2.0 License, but we might have to double check.

crypto.rs allows dependencies with a BSD 3-Clause licence so I think we should be good: https://github.com/iotaledger/crypto.rs/blob/dev/.github/workflows/about.toml

identity-iota/src/tangle/message_version.rs
olivereanderson: In my understanding associated functions are mostly used as constructors (in Rust) and it feels more idiomatic the way things are now, but I would be happy to be proven wrong. Granted moving the `CURRENT_MESSAGE_VERSION` constant into `MessageVersion` and doing something like this 
```
impl MessageVersion {
  pub(crate) const CURRENT_MESSAGE_VERSION: Self = MessageVersion::V1;
  
/// Adds the current message version flag at the beginning of arbitrary data.
  pub(crate) fn add_version_flag(mut data: Vec<u8>) -> Vec<u8> {
    let version_flag = Self::CURRENT_MESSAGE_VERSION as u8;
    data.splice(0..0, [version_flag].iter().cloned());
    data
  }
  ```
  is maybe neater than how it is now, but I don't see why it should be `MessageVersion`'s "responsibility" to check a version flag, i.e. it feels strange to have `check_version_flag`as an associated function, in which case I would rather have both of them as plain functions defined in this module the way it is now. 

Note that this is mega bike shedding, and I will not complain if @abdulmth refactors it the way you suggested.

Comment (olivereanderson) [reply identity-iota/src/tangle/message_version.rs:26]: Update. Having thought about this a bit more I think the following is the "correct" version:

```rust
#[derive(Copy, Clone)]
pub(crate) enum MessageVersion {
  V1 = 1,
}

impl MessageVersion {
  pub(crate) const CURRENT_MESSAGE_VERSION: Self = MessageVersion::V1;
  
/// Adds the current message version flag at the beginning of arbitrary data.
  pub(crate) fn add_version_flag(mut data: Vec<u8>) -> Vec<u8> {
    let version_flag = Self::CURRENT_MESSAGE_VERSION as u8;
    data.splice(0..0, [version_flag].iter().cloned());
    data
  }

  /// Checks if flag matches message_version.
  pub(crate) fn check_version_flag(&self, flag: &u8) -> Result<(), Error> {
    if *self as u8 == *flag {
      Ok(())
    } else {
      Err(Error::InvalidMessageFlags)
    }
  }
}
```

olivereanderson: Update to the update. Having thought about it even more, it is "correct" the way it is now, except that the 
`add_version_flag` function should probably be renamed to `add_current_version_flag`, and the `CURRENT_MESSAGE_VERSION` constant should be an associated constant `CURRENT` of `MessageVersion`. So all in all it would be :
```rust
#[derive(Copy, Clone)]
pub(crate) enum MessageVersion {
  V1 = 1,
}

impl MessageVersion {
  const CURRENT: Self = MessageVersion::V1; 
}

/// Adds the current message version flag at the beginning of arbitrary data.
pub(crate) fn add_current_version_flag(mut data: Vec<u8>) -> Vec<u8> {
  let version_flag = MessageVersion::CURRENT as u8;
  data.splice(0..0, [version_flag].iter().cloned());
  data
}

/// Checks if flag matches message_version.
pub(crate) fn check_version_flag(flag: &u8, message_version: MessageVersion) -> Result<(), Error> {
  if message_version as u8 == *flag {
    Ok(())
  } else {
    Err(Error::InvalidMessageFlags)
  }
}
```

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Adds the current encoding flag at the beginning of arbitrary data.
+pub(crate) fn add_encoding_version_flag(mut data: Vec<u8>) -> Vec<u8> {
+  let encoding_version = CURRENT_ENCODING_VERSION as u8;
+
+  data.splice(0..0, [encoding_version].iter().cloned());
+  data
+}
+
+/// Decompresses a message depending on the encoding flag.
+pub(crate) fn get_decompressed_message_data<T: AsRef<[u8]>>(encoding_flag: &u8, data: T) -> Result<String, Error> {
```
olivereanderson, line 24:
```suggestion
decompress_message<T: AsRef<[u8]>>(encoding_flag: &u8, data: T) -> Result<String, Error> {
```

olivereanderson (changes requested):
Great work! 

Most of it is good as it is, just a few things that should be renamed, and you can consider incorporating the encoding, and versioning constants into the `MessageEncoding` and `MessageVersion`types as associated constants.

identity-iota/src/tangle/compression_brotli.rs
```diff
@@ -0,0 +1,48 @@
...
+  let mut compressor = brotli::CompressorReader::new(input.as_ref(), BUFFER_SIZE, QUALITY, WINDOWS_SIZE);
+  compressor
+    .read_to_end(&mut result)
+    .map_err(|_| Error::CompressionError)?;
+  Ok(result)
+}
+
+pub(crate) fn decompress_brotli<T: AsRef<[u8]> + ?Sized>(input: &T) -> Result<Vec<u8>, Error> {
+  let mut decompressor = brotli::Decompressor::new(input.as_ref(), BUFFER_SIZE);
+  let mut s = Vec::new();
```
olivereanderson, line 23:
```suggestion
let mut buff = Vec::new();
```

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,48 @@
...
+/// Adds the current encoding flag at the beginning of arbitrary data.
+pub(crate) fn add_encoding_version_flag(mut data: Vec<u8>, encoding: MessageEncodingVersion) -> Vec<u8> {
+  let encoding_version = encoding as u8;
+
+  data.splice(0..0, [encoding_version].iter().cloned());
+  data
+}
+
+/// Decompresses a message depending on the encoding flag.
+pub(crate) fn get_decompressed_message_data(encoding_flag: &u8, data: &[u8]) -> Result<Vec<u8>, Error> {
```
olivereanderson, line 24:
```suggestion
pub(crate) fn decompress_message(encoding_flag: &u8, data: &[u8]) -> Result<Vec<u8>, Error> {
```

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,48 @@
...
+  return if *encoding_flag == MessageEncodingVersion::JsonBrotli as u8 {
+    decompress_brotli(data.as_ref())
+  } else if *encoding_flag == MessageEncodingVersion::Json as u8 {
+    Ok(data.to_vec()) //todo prevent copying the slice.
+  } else {
+    Err(Error::InvalidMessageFlags)
+  };
```
olivereanderson, line 31:
```suggestion
 if *encoding_flag == MessageEncodingVersion::JsonBrotli as u8 {
    decompress_brotli(data.as_ref())
  } else if *encoding_flag == MessageEncodingVersion::Json as u8 {
    Ok(data.to_vec()) //todo prevent copying the slice.
  } else {
    Err(Error::InvalidMessageFlags)
  }
  ```

File: identity-iota/src/tangle/client.rs [hunk 20]
```diff
@@ -74,23 +76,37 @@ impl Client {
...
+      compressed_data = MessageEncoding::add_encoding_version_flag(compressed_data, MessageEncoding::Json);
+    }
+    compressed_data = MessageVersion::add_version_flag(compressed_data, MessageVersion::CURRENT);
+
     self
       .client
       .message()
       .with_index(index)
-      .with_data(data.to_json_vec()?)
+      .with_data(compressed_data)
```
olivereanderson, line 109:
```suggestion
    let encoded_message_data = {
      if compression {
        MessageEncoding::compress_message(&data.to_json()?, MessageEncoding::JsonBrotli).map(|compressed_data| {
          MessageEncoding::add_encoding_version_flag(compressed_data, MessageEncoding::JsonBrotli)
        })?
      } else {
        data.to_json_vec().map(|uncompressed_data| {
          MessageEncoding::add_encoding_version_flag(uncompressed_data, MessageEncoding::Json)
        })?
      }
    };
    let message_data = MessageVersion::add_version_flag(encoded_message_data, MessageVersion::CURRENT);

    self
      .client
      .message()
      .with_index(index)
      .with_data(message_data)
```
is how I would have written this, but you can keep it the way it is if you prefer.

identity-iota/src/tangle/compression_brotli.rs
```diff
@@ -0,0 +1,50 @@
...
+use crate::error::Error;
+use crate::error::Error::CompressionError;
+use std::io::prelude::*;
+
+const BUFFER_SIZE: usize = 4096;
+const QUALITY: u32 = 5; // compression level
+const WINDOWS_SIZE: u32 = 22;
+
+pub(crate) fn compress_brotli<T: AsRef<[u8]>>(input: T) -> Result<Vec<u8>, Error> {
+  let mut result = Vec::new();
```
olivereanderson, line 13: nit: Consider renaming `result` to `buf`

identity-iota/src/tangle/message_version.rs
```diff
@@ -0,0 +1,41 @@
...
+  /// Checks if flag matches message_version.
+  pub(crate) fn check_version_flag(flag: &u8, message_version: MessageVersion) -> Result<(), Error> {
+    if message_version as u8 == *flag {
+      Ok(())
+    } else {
+      Err(Error::InvalidMessageFlags)
+    }
+  }
+}
+
```
olivereanderson, line 30: It feels very adhoc and non-idiomatic Rust to have these as associated functions. I would suggest to either (1): turn them into methods so one would instead have: 
```rust
fn add version_flag(&self, mut data: Vec<u8>) -> Vec<u8>

fn check_version_flag(&self, flag: &u8) -> Result<(), Error>
```
or alternatively one can (2): rename the module to `message_versioning`
and move `add_version_flag` and `check_version_flag` out of `MessageVersion` and then one would call them 
with `message_versioning::add_version_flag`, `message_versioning::check_version_flag` elsewhere in the crate. 

What do you think @cycraig ?

identity-iota/src/tangle/client.rs
```diff
@@ -74,23 +76,37 @@ impl Client {
...
+      .await
   }
 
-  /// Publishes arbitrary JSON data to the specified index on the Tangle.
-  pub async fn publish_json<T: ToJson>(&self, index: &str, data: &T) -> Result<Receipt> {
+  /// Compresses and Publishes arbitrary JSON data to the specified index on the Tangle.
+  pub async fn publish_json<T: ToJson>(&self, index: &str, data: &T, compression: bool) -> Result<Receipt> {
+    let mut compressed_data: Vec<u8>;
+    if compression {
+      compressed_data = MessageEncoding::compress_message(&data.to_json()?, MessageEncoding::JsonBrotli)?;
```
olivereanderson, line 97: The double `MessageEncoding` looks a bit weird.

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  /// Compresses a message depending on current encoding version.
+  pub(crate) fn compress_message<T: AsRef<[u8]>>(message: T, encoding: MessageEncoding) -> Result<Vec<u8>, Error> {
+    match encoding {
+      MessageEncoding::JsonBrotli => compress_brotli(message),
+      MessageEncoding::Json => Err(CompressionError),
+    }
+  }
```
olivereanderson, line 41: My comment regading `message_version.rs` also applies here.

identity-iota/src/tangle/client.rs
```diff
@@ -74,23 +76,37 @@ impl Client {
...
+  pub async fn publish_diff(&self, message_id: &MessageId, diff: &DocumentDiff, compression: bool) -> Result<Receipt> {
+    self
+      .publish_json(&IotaDocument::diff_index(message_id)?, diff, compression)
+      .await
   }
 
-  /// Publishes arbitrary JSON data to the specified index on the Tangle.
-  pub async fn publish_json<T: ToJson>(&self, index: &str, data: &T) -> Result<Receipt> {
+  /// Compresses and Publishes arbitrary JSON data to the specified index on the Tangle.
+  pub async fn publish_json<T: ToJson>(&self, index: &str, data: &T, compression: bool) -> Result<Receipt> {
```
olivereanderson, line 94: If this feature: https://blog.rust-lang.org/inside-rust/2021/09/06/Splitting-const-generics.html#featureconst_generics_defaults lands before our 1.0 release, then we can use that to essentially set `compression = true` as a default, and then one has to write a bit of boiler plate in the rare cases where no compression is desired. For the time being it is probably fine to have `compression` as a parameter.

identity-iota/src/tangle/client_map.rs
```diff
@@ -94,6 +101,10 @@ impl ClientMap {
 
     Ok(client)
   }
+
+  pub fn disable_compression(&mut self) {
+    self.compression = false;
+  }
```
cycraig, line 107: This should be `set_compression`, otherwise there is no way to re-enable compression in a Client.

identity-iota/src/tangle/client.rs
```diff
@@ -74,23 +76,37 @@ impl Client {
...
-    self.publish_json(&IotaDocument::diff_index(message_id)?, diff).await
+  pub async fn publish_diff(&self, message_id: &MessageId, diff: &DocumentDiff, compression: bool) -> Result<Receipt> {
+    self
+      .publish_json(&IotaDocument::diff_index(message_id)?, diff, compression)
+      .await
   }
 
-  /// Publishes arbitrary JSON data to the specified index on the Tangle.
-  pub async fn publish_json<T: ToJson>(&self, index: &str, data: &T) -> Result<Receipt> {
+  /// Compresses and Publishes arbitrary JSON data to the specified index on the Tangle.
```
cycraig, line 93: I think `compression` should be changed to be on the `Client` instead of the `ClientMap`. None of these methods would need a `compression` parameter then and `publish_json` would be the single place it is accessed.

identity-iota/src/error.rs
```diff
@@ -53,4 +53,8 @@ pub enum Error {
   NoExplorerURLSet,
   #[error("Invalid Explorer Url")]
   InvalidExplorerURL,
+  #[error("compression error")]
+  CompressionError,
+  #[error("invalid message flags")]
+  InvalidMessageFlags,
```
cycraig, line 59: Do we want to be more specific with `InvalidMessageVersion` and `InvalidMessageEncoding`?

identity-iota/src/tangle/client_map.rs
```diff
@@ -23,19 +23,26 @@ type State = DashMap<NetworkName, Arc<Client>>;
 #[derive(Debug)]
 pub struct ClientMap {
   data: State,
+  compression: bool,
 }
```
cycraig, line 27: 1. `compression` should be set on the `Client` instead of the `ClientMap` to allow different settings per private Tangle, for example.
2. `compression` could be a `MessageEncoding` enum instead for forwards compatibility with any new encodings that are added.

identity-iota/src/tangle/client.rs
```diff
@@ -185,7 +201,7 @@ impl Client {
   }
 }
 
-#[async_trait::async_trait(?Send)]
+#[async_trait::async_trait(? Send)]
```
cycraig, line 204: nit: IDE formatting thing.

```suggestion
#[async_trait::async_trait(?Send)]
```

identity-iota/src/tangle/client_map.rs
```diff
@@ -102,7 +113,7 @@ impl Default for ClientMap {
   }
 }
 
-#[async_trait::async_trait(?Send)]
+#[async_trait::async_trait(? Send)]
```
cycraig, line 116: nit: IDE formatting thing.

```suggestion
#[async_trait::async_trait(?Send)]
```

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,54 @@
...
+use crate::error::Error::CompressionError;
+use crate::tangle::compression_brotli::compress_brotli;
+use crate::tangle::compression_brotli::decompress_brotli;
+use crate::Error;
+
+#[derive(Copy, Clone)]
+pub(crate) enum MessageEncoding {
+  Json = 0,
+  JsonBrotli = 1,
+}
```
cycraig, line 13: I wonder if we need to separate `MessageEncoding` and `MessageCompression` into two flags since they can be applied separately? That would avoid duplicating the compression options for each encoding (e.g. `Json, JsonBrotli, Cbor, CborBrotli`). Just a thought, I doubt we'll need a different encoding other than JSON, hopefully.

identity-iota/src/tangle/message_version.rs
```diff
@@ -0,0 +1,41 @@
...
+impl MessageVersion {
+  pub(crate) const CURRENT: Self = MessageVersion::V1;
+
+  /// Adds the current message version flag at the beginning of arbitrary data.
+  pub(crate) fn add_version_flag(mut data: Vec<u8>, message_version: MessageVersion) -> Vec<u8> {
+    let version_flag = message_version as u8;
+    data.splice(0..0, [version_flag].iter().cloned());
+    data
+  }
```
cycraig, line 19: Since both `add_version_flag` and `add_encoding_version_flag` both use `splice` it adds extra allocations unnecessarily. The methods can be combined and optimised by pre-allocating a vec with the data first, in `publish_json` or a new `pack_message` etc.

identity-iota/src/tangle/encoding.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::error::Error::CompressionError;
+use crate::tangle::compression_brotli::compress_brotli;
+use crate::tangle::compression_brotli::decompress_brotli;
+use crate::Error;
+
+#[derive(Copy, Clone)]
+pub(crate) enum MessageEncoding {
```
cycraig, line 10: Should we rename `MessageEncoding` and `MessageVersion` to `DIDMessageEncoding` and `DIDMessageVersion` to avoid confusion with actual Tangle messages?

cycraig (changes requested):
More changes.

identity-iota/src/tangle/client.rs
cycraig: The extra brackets in the suggestion after 

```rust
let encoded_message_data = { 
  if compression {
```
can be collapsed to 

```rust
let encoded_message_data = if compression {
```

identity-iota/src/tangle/client.rs
abdulmth: The way the clients are stored inside the `ClientMap` makes them hard to mutate. Storing them in a `Mutex` for example looked more complicated.

cycraig: Fair point.

identity-iota/src/tangle/encoding.rs
abdulmth: If that will be really necessary in the future, the message version can indicate that. For example starting from message version 2, the second byte is for `MesageEncoding` and the third is for `MessageCompression`. But I doubt using many combinations.

identity-iota/src/tangle/client.rs
abdulmth: The compression argument is removed from the methods in `client.rs`, instead the state of the `Client` is used.

identity-iota/src/tangle/client.rs
abdulmth: Now, disabling compression can only be done through the `Client` directly. And not through the `ClientMap`. But a `Client` can be added to a `ClientMap` manually. For the WASM bindings, it can be set through `Config`.

identity-iota/src/error.rs
abdulmth: Since there's too many Errors already, and they probably will get refactored anyway, maybe we can leave it as is for now.

identity-iota/src/tangle/message_version.rs
abdulmth: I'm not sure where the unnecessary allocations happen, to my understanding, splice doesn't copy the entire vector in order to prepend items. And the idea of calling splice twice was for the sake of separation of concerns.

cycraig: > splice doesn't copy the entire vector in order to prepend items

Pre-pending with `splice` needs to allocate a new `Vec` with enough space and move the existing `Vec` into it as far as I know. It's efficient but this is still done twice which is unnecessary.

> And the idea of calling splice twice was for the sake of separation of concerns.

They're both operations that only happen when packing a message to be published, they can be combined.

cycraig: From https://doc.rust-lang.org/std/vec/struct.Vec.html#method.splice


> The input iterator replace_with is only consumed when the Splice value is dropped.
> 
> This is optimal if:
> 
> - The tail (elements in the vector after range) is empty,
> - or replace_with yields fewer or equal elements than range’s length
> - or the lower bound of its size_hint() is exact.
> 
> Otherwise, a temporary vector is allocated and the tail is moved twice.

identity-iota/src/tangle/client_builder.rs
```diff
@@ -36,6 +39,12 @@ impl ClientBuilder {
     self
   }
 
+  /// Disables message compression.
+  pub fn disable_compression(mut self) -> Self {
+    self.encoding = DIDMessageEncoding::Json;
+    self
+  }
```
cycraig, line 46: I still think this should be `set_compression` that takes the `DIDMessageEncoding` enum for forwards compatibility.

abdulmth: I changed it now. 
The idea for using `disable_compression`  is that it implicitly means that the default is enabling compression + it is called on a builder and a config, so disabling compression then enabling it doesn't make much sense.

identity-iota/src/tangle/client.rs
```diff
@@ -183,6 +198,19 @@ impl Client {
...
+  /// Adds message version flag and encoding flag at the beginning of arbitrary message data.
+  fn add_flags_to_message(
+    mut message_data: Vec<u8>,
+    message_version: DIDMessageVersion,
+    encoding: DIDMessageEncoding,
+  ) -> Vec<u8> {
+    let message_version_flag = message_version as u8;
+    let encoding_flag = encoding as u8;
+
+    message_data.splice(0..0, [message_version_flag, encoding_flag].iter().cloned());
```
olivereanderson, line 211:
```suggestion
message_data.splice(0..0, [message_version_flag, encoding_flag].into_iter());
``` 
This works now because we are using the 2021 edition of Rust.

identity-iota/src/tangle/client_builder.rs
cycraig: Thanks for the change. 

I agree `disable_compression` is more explanative for examples but if/when more than two options are available in the future I would prefer to use the same `set_compression` function to choose between them. 
Come to think of it, the function should actually be called `set_encoding` or `set_message_encoding` since it sets the `DIDMessageEncoding`. Yeah I should have really picked up on the naming thing before, my bad.

identity-iota/src/tangle/did_encoding.rs
```diff
@@ -0,0 +1,32 @@
...
+/// Decompresses a message depending on the encoding flag.
+pub(crate) fn decompress_message(encoding_flag: &u8, data: &[u8]) -> Result<Vec<u8>, Error> {
+  if *encoding_flag == DIDMessageEncoding::JsonBrotli as u8 {
+    decompress_brotli(data)
+  } else if *encoding_flag == DIDMessageEncoding::Json as u8 {
+    Ok(data.to_vec()) //todo prevent copying the slice.
+  } else {
+    Err(Error::InvalidMessageFlags)
+  }
+}
```
cycraig, line 24: ```//todo prevent copying the slice.```

Maybe we should embed this directly in `parse_data` to avoid cloning uncompressed messages? Most messages will be compressed so I do not anticipate this to be a major issue, but it is annoying.

identity-iota/src/tangle/did_encoding.rs
```diff
@@ -0,0 +1,32 @@
...
+pub(crate) enum DIDMessageEncoding {
+  Json = 0,
+  JsonBrotli = 1,
+}
+
+/// Decompresses a message depending on the encoding flag.
+pub(crate) fn decompress_message(encoding_flag: &u8, data: &[u8]) -> Result<Vec<u8>, Error> {
+  if *encoding_flag == DIDMessageEncoding::JsonBrotli as u8 {
+    decompress_brotli(data)
+  } else if *encoding_flag == DIDMessageEncoding::Json as u8 {
```
cycraig, line 19: Checking the flag can be made easier by implementing `TryFrom<u8>` or `FromPrimitive` on the `DIDMessageEncoding` enum.

identity-iota/src/tangle/did_message_versioning.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Copy, Clone)]
+pub(crate) enum DIDMessageVersion {
+  V1 = 1,
+}
+
+pub(crate) const CURRENT_MESSAGE_VERSION: DIDMessageVersion = DIDMessageVersion::V1;
+
+/// Checks if flag matches message_version.
+pub(crate) fn check_version_flag(flag: &u8, message_version: DIDMessageVersion) -> Result<(), Error> {
+  if message_version as u8 == *flag {
```
cycraig, line 15: Same as `DIDMessageEncoding`, this would benefit from a  `TryFrom<u8>` or `FromPrimitive` implementation.

identity-iota/src/tangle/client.rs
cycraig: Just the following worked due to the bounds on splice (and Rust 2021):
```rust
message_data.splice(0..0, [message_version_flag, encoding_flag]);
```

identity-iota/src/tangle/message/message_version.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use crate::Error;
+
+/// Indicates the version of a DID Message.
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, num_derive::FromPrimitive)]
+pub enum DIDMessageVersion {
+  V1 = 1,
+}
+
+pub(crate) const CURRENT_MESSAGE_VERSION: DIDMessageVersion = DIDMessageVersion::V1;
```
olivereanderson, line 14: Consider making this an associated constant of the `DIDMessageVersion` struct like this:
```rust
impl DIDMessageVersion {
    const CURRENT: Self = DIDMessageVersion::V1 
}
```

olivereanderson (changes requested):
I think it would be neater to change the module's constant to be an associated constant, apart from that this looks good to me.

cycraig (approved):

identity-iota/src/tangle/message/message_ext.rs
```diff
@@ -0,0 +1,191 @@
...
+  message_version: DIDMessageVersion,
+  encoding: DIDMessageEncoding,
+) -> Vec<u8> {
+  let mut buffer: Vec<u8> = Vec::with_capacity(1 + DID_MESSAGE_MARKER.len() + 1 + data.len());
+  buffer.push(message_version as u8);
+  buffer.extend_from_slice(DID_MESSAGE_MARKER);
+  buffer.push(encoding as u8);
+  buffer.append(&mut data);
+  buffer
+}
```
olivereanderson, line 107: mega nit: One could also write this as:
```suggestion
fn add_flags_to_message(data: Vec<u8>, message_version: DIDMessageVersion, encoding: DIDMessageEncoding) -> Vec<u8> {
  std::iter::once(message_version as u8)
    .chain(DID_MESSAGE_MARKER.into_iter().copied())
    .chain(std::iter::once(encoding as u8))
    .chain(data)
    .collect()
}
```

olivereanderson (approved):

identity-iota/src/tangle/message/message_ext.rs
cycraig: Does `chain` pre-allocate a `Vec` of the exact size needed? The above was to avoid unnecessary memory allocations.

abdulmth: Since everyone agrees that the original is good enough, I will merge for now since this PR has been open for a long time.

PR #477, opened 2021-11-02
Add new images

Dr-Electron:
# Description of change

New images from @JSto91
As always: [before/after](https://www.figma.com/file/WxlD0ZV9x0IAKUObrUY1c7/Wiki-images-update?node-id=203%3A26)
[Wiki](https://dr-electron.github.io/iota-wiki/identity.rs/specs/didcomm/protocols/connection)

## Type of change

- [x] Documentation Fix

## Change checklist


JelleMillenaar:
We are not sure if we can merge this. While we appreciate the improved images, the images are currently generated from a draw.io file, part of the repo. This allows anyone to submit a PR for suggested changes to DIDComm, including updated graphics. Merging this in removes that flexibility and might "centralize" updates to DIDComm to only IOTA Foundation members.

JSto91:
Nice, thanks for the information :)
I'll take a look at draw.io. Maybe it can be stylized. If you are ok with it :)

Dr-Electron:
> We are not sure if we can merge this. While we appreciate the improved images, the images are currently generated from a draw.io file, part of the repo. This allows anyone to submit a PR for suggested changes to DIDComm, including updated graphics. Merging this in removes that flexibility and might "centralize" updates to DIDComm to only IOTA Foundation members.

Hmm. I saw that it's draw.io graphs but didn't see the file. So i thought it was closed development. Thanks for the heads up :)

Dr-Electron:
> Thanks for the information :)
> I'll take a look at draw.io. Maybe it can be stylized. If you are ok with it :)

It is. To a certain degree at least. I never really changed the styling 😅

JSto91:
> > Thanks for the information :)
> > I'll take a look at draw.io. Maybe it can be stylized. If you are ok with it :)
> 
> It is. To a certain degree at least. I never really changed the styling 😅

Yeah i looked into it. You can predefine colors etc into the palette and also the basic shapes. Don't know if you can add/import custom shapes. Need to dig in further 😄. I'll look if i can come up with some solution.

JSto91:
**Update:** So i looked for solutions and there are 3 that could work:

**Option 1:**

Create custom Icon/Shape Library for draw.io which is located in the Repository

How this would work (Icon/Shape Library already imported):
	- User forks this Repo
	- Opens draw.io
	- imports Library via Github (selects library file from his repo)
	- has access to the Icon/Shapes

Pro:
	- draw.io is free
	- no signup required
	- can be linked with Github

Con:
	- still limited style option
	- color (for fonts) would need to be changed manually
	- needs a "readme" somewhere for the colors (Color Hex)
	- different font

---

**Option 2:**

Publish seperate Figma-File so anyone can duplicate it and do changes

Pro:
	- free
	- fully customizable

Con:
	- Needs to upload manually to Github (which isn't a big deal imo)
	- Needs to be a link to the published Figma somewhere in the wiki
	- Signup required (They have the option for SAML SSO but i am not familiar with that. Don't know if this makes it any better)

---

**Option 3:**

Upload the images and leave the original (from draw.io) in another folder in the Repo.
If someone wants to change something in the flowchart, he can do it with the original-file in draw.io and replace the current-image with the updated-image.
After the updated img is online, i can stylize it.
In my opinion the best option.

But what do you think? @Dr-Electron @JelleMillenaar

eike-hass:
Hi guys,
really appreciate the effort! My thoughts are as following:
Option 1 - We shouldn't change the shapes, they have meaning in the BPMN Notation. A thick outline means something different then a thin one, then a dotted one. If we can enhance the diagrams with changing colors, or more generally changes that don't change the meaning that might an interesting option.
Option 2 - since we want to stick to BPMN and potentially utilize other features or "shapes" it would mean to recreate the whole spec in Figma. I think that would be an insane amount of work.
Option 3 - this puts a lot of work on you and also creates a dependency. How long does it take you to respond to changes on average? Is there a fallback? Also we shouldn't deviate from the shapes, as they have meaning. So it would only be minor recolor, maybe different backgrounds and fonts.

I lean towards Option 1 if it improves the look without changing the meaning. We need to keep in mind that we would want to use the whole spec, so be able to drag in other shapes at any time. This Option is only worthwhile if it doesn't complicate the workflow significantly.
Option 2 and 3 create too much overhead in my Opinion.

I think we should consider leaving it as is as a very valid option. We can spice it up on a per diagram basis for special use cases like presentations or blog post, but leaving the spec as is has the most streamlined workflow.

JSto91:
Hi Eike,
thank you for the information :). In this case your Option 1 makes more sense i agree. I will use the original flowchart and just change some colors (Background etc) and leave the shapes/line-style/etc as they are. If that is fine.

Dr-Electron:
> Hi guys, really appreciate the effort! My thoughts are as following: Option 1 - We shouldn't change the shapes, they have meaning in the BPMN Notation. A thick outline means something different then a thin one, then a dotted one. If we can enhance the diagrams with changing colors, or more generally changes that don't change the meaning that might an interesting option. Option 2 - since we want to stick to BPMN and potentially utilize other features or "shapes" it would mean to recreate the whole spec in Figma. I think that would be an insane amount of work. Option 3 - this puts a lot of work on you and also creates a dependency. How long does it take you to respond to changes on average? Is there a fallback? Also we shouldn't deviate from the shapes, as they have meaning. So it would only be minor recolor, maybe different backgrounds and fonts.
> 
> I lean towards Option 1 if it improves the look without changing the meaning. We need to keep in mind that we would want to use the whole spec, so be able to drag in other shapes at any time. This Option is only worthwhile if it doesn't complicate the workflow significantly. Option 2 and 3 create too much overhead in my Opinion.
> 
> I think we should consider leaving it as is as a very valid option. We can spice it up on a per diagram basis for special use cases like presentations or blog post, but leaving the spec as is has the most streamlined workflow.

Everything you say makes sense. But also boring specs have the right to look beautiful :trollface: 
So let's try option 1

JSto91:
Hey 😃 
so i re-colored the current draw.io flowchart.
Example: https://critical94.github.io/iota-wiki/learn/future/future-of-iota

Question: We used the max-width of the content so the image isn't too small anymore. I would like to resize/move the content of the flowchart a little bit so it doesn't look cramped.

What do you think about the colors? Something you don't like? Let me know :)

eike-hass:
> Hey 😃 so i re-colored the current draw.io flowchart. Example: https://critical94.github.io/iota-wiki/learn/future/future-of-iota
> 
> Suggestion: We used the max-width of the content so the image isn't too small anymore. I would like to resize/move the content of the flowchart a little bit so it doesn't look cramped.
> 
> What do you think about the colors? Something you don't like? Let me know :)

Looking very good! Couple of issues came up discussing it internally:

- the second decision block for ´Require information?´ does not have an orange arrow in the negative case.
- some blocks are yellow and some are teal inconsistently. ´Sign verifiable credential´ might be meant to be yellow instead of teal?
-  Maybe the send and receive blocks should be the same color. The contrast on the receive blocks are too low to read comfortably.

JelleMillenaar:
Any update on this PR. Do we want to stick with our current images or incorporate the feedback from Eike. @JSto91 & @Dr-Electron :)

Dr-Electron:
I really love the new banners.
Also really cool that we now have a flowchart lib. But shouldn't we make it more general? So a basic IOTA lib?
Or are there thing that will be different for identity? Do other frameworks even use flowchart so that we could reuse?

PR #481, opened 2021-11-05
Enable deep equality in `OrderedSet`

PhilippGackstatter:
# Description of change

The current implementation of `CoreDocument`, which uses an `OrderedSet<DIDKey<T>>` to represent verification methods, does not work as intended when rotating the key material of a given verification method. In this case, the diff between the old and the updated document is empty, and `old_doc == updated_doc` returns `true`. This is because `DIDKey` overwrites the equality implementation to compare the contents of the set by key only (i.e. the DID url).

To fix this, the `DIDKey` wrapper is removed, such that normal equality checks work as expected and intended. To disallow adding multiple methods with the same fragment but, for example, different key material -- which would be allowed with regular equality checks -- a `KeyComparable` trait is introduced which is used in the `contains` check. This trait requires elements to return a `Key` by which they are compared. A default implementation is provided such that types that implement `AsRef<CoreDIDUrl>` will use that url as the key. This approach should be flexible enough to allow other keys in the future.

Note to reviewers: I would appreciate a careful check of the changed trait bounds on the `OrderedSet` methods, to make sure they make sense.

## Type of change

- [ ] Documentation Fix

Existing tests all pass, and new tests were added to check the introductory example works as intended.

## Change checklist


cycraig (changes requested):
I really like the simplification and removal of `DIDKey`!

There is a problem with the `OrderedSet::update` and `OrderedSet::replace` methods since they still use the old equality comparsions. 

Suggested changes:

```rust
  /// Replaces a `current` value with the given `update` value; returns `true`
  /// if the value was successfully replaced.
  #[inline]
  pub fn replace<U>(&mut self, current: &U, update: T) -> bool
  where
    T: KeyComparable,
    U: KeyComparable<Key = T::Key>,
  {
    self.change(update, |item, update| item.as_key() == current.as_key() || item.as_key() == update.as_key())
  }

  /// Updates an existing value in the `OrderedSet`; returns `true` if the value
  /// was successfully updated.
  #[inline]
  pub fn update(&mut self, update: T) -> bool
  where
    T: KeyComparable,
  {
    self.change(update, |item, update| item.as_key() == update.as_key())
  }
```


Here are the relevant unit tests which fail with the current PR:
```rust

  #[derive(Clone, Copy, PartialEq, Eq)]
  struct ComparableStruct {
    key: u8,
    value: i32,
  }

  impl KeyComparable for ComparableStruct {
    type Key = u8;

    fn as_key(&self) -> &Self::Key {
      &self.key
    }
  }

  #[test]
  fn test_ordered_set_replace() {
    let mut set = OrderedSet::new();

    // Create two structs with the same key.
    let cs1 = ComparableStruct {
      key: 0,
      value: 10,
    };
    let cs2 = ComparableStruct {
      key: 0,
      value: 20,
    };

    // Try replace it with the second.
    // This should succeed because the keys are equivalent.
    assert!(set.append(cs1));
    assert_eq!(set.len(), 1);

    assert!(set.replace(&cs1, cs2));
    assert_eq!(set.len(), 1);
    assert_eq!(set.head().unwrap().key, cs2.key);
    assert_eq!(set.head().unwrap().value, cs2.value);
  }

  #[test]
  fn test_ordered_set_replace_all() {
    let mut set = OrderedSet::new();
    let cs1 = ComparableStruct {
      key: 0,
      value: 10,
    };
    let cs2 = ComparableStruct {
      key: 1,
      value: 20,
    };
    assert!(set.append(cs1));
    assert!(set.append(cs2));
    assert_eq!(set.len(), 2);

    // Now replace cs1 with something that has the same key as cs2.
    // This should replace BOTH cs1 AND cs2.
    let cs3 = ComparableStruct {
      key: 1,
      value: 30,
    };
    assert!(set.replace(&cs1, cs3));
    assert_eq!(set.len(), 1);
    assert_eq!(set.head().unwrap().key, cs3.key);
    assert_eq!(set.head().unwrap().value, cs3.value);
  }

  #[test]
  fn test_ordered_set_update() {
    let mut set = OrderedSet::new();
    let cs1 = ComparableStruct {
      key: 0,
      value: 10,
    };
    assert!(set.append(cs1));
    assert_eq!(set.len(), 1);

    // This should update the value of cs1 since the keys are the same.
    let cs2 = ComparableStruct {
      key: 0,
      value: 20,
    };
    assert!(set.update(cs2));
    assert_eq!(set.len(), 1);
    assert_eq!(set.head().unwrap().key, cs2.key);
    assert_eq!(set.head().unwrap().value, cs2.value);

    // This should NOT update anything since the key does not match.
    let cs3 = ComparableStruct {
      key: 1,
      value: 20,
    };
    assert!(!set.update(cs3));
    assert_eq!(set.len(), 1);
    assert_eq!(set.head().unwrap().key, cs2.key);
    assert_eq!(set.head().unwrap().value, cs2.value);
  }
```

identity-iota/src/did/doc/iota_document.rs
```diff
@@ -1593,19 +1592,45 @@ mod tests {
...
+    let keypair: KeyPair = KeyPair::new_ed25519().unwrap();
+    let verif_method2: IotaVerificationMethod =
+      IotaVerificationMethod::from_did(doc1.id().to_owned(), &keypair, "test-0").unwrap();
+
+    let was_inserted = doc2.insert_method(verif_method2, MethodScope::CapabilityInvocation);
+
+    // Nothing was inserted, because a method with the same fragment already existed.
+    assert!(!was_inserted);
+    assert_eq!(doc1, doc2);
+  }
```
cycraig, line 1635: nit: variable names

```suggestion
  #[test]
  fn test_document_equality() {
    let keypair1: KeyPair = KeyPair::new_ed25519().unwrap();
    let method1: IotaVerificationMethod = IotaVerificationMethod::from_keypair(&keypair1, "test-0").unwrap();

    let original_doc = IotaDocument::from_verification_method(method1).unwrap();

    let mut doc1 = original_doc.clone();

    // Update the key material of the existing verification method test-0.
    let keypair2: KeyPair = KeyPair::new_ed25519().unwrap();
    let method2: IotaVerificationMethod =
      IotaVerificationMethod::from_did(doc1.id().to_owned(), &keypair2, "test-0").unwrap();

    doc1.remove_method(doc1.id().to_url().join("#test-0").unwrap()).unwrap();
    doc1.insert_method(method2, MethodScope::CapabilityInvocation);

    // Even though the method fragment is the same, the key material has been updated
    // so the two documents are expected to be not equal.
    assert_ne!(original_doc, doc1);

    let mut doc2 = doc1.clone();
    let keypair3: KeyPair = KeyPair::new_ed25519().unwrap();
    let method3: IotaVerificationMethod =
      IotaVerificationMethod::from_did(doc1.id().to_owned(), &keypair3, "test-0").unwrap();

    let was_inserted = doc2.insert_method(method3, MethodScope::CapabilityInvocation);

    // Nothing was inserted, because a method with the same fragment already existed.
    assert!(!was_inserted);
    assert_eq!(doc1, doc2);
  }
```

cycraig (approved):
Looks good to me.

HenriqueNogara (approved):

PR #491, opened 2021-11-12
Make create_signature and sign Async for RemoteEd25519

### LINKED ISSUE #441: [Bug] Deadlock in `RemoteSign` when using `Stronghold`
## Bug description

When using the `RemoteSign` type with the `Stronghold` storage implementation concurrently from two different tokio tasks, a deadlock can occur. The `MemStore` is unaffected.

The stronghold `Context` has an `AsyncMutex` that is acquired on every call to `scope`:

https://github.com/iotaledger/identity.rs/blob/ee4cce5dbff12ad6d5a3ee5e380866ab31f7e7ec/identity-account/src/stronghold/context.rs#L115-L127

The actual problem is in the `RemoteSign` implementation:

https://github.com/iotaledger/identity.rs/blob/ee4cce5dbff12ad6d5a3ee5e380866ab31f7e7ec/identity-account/src/crypto/remote.rs#L48-L54

Since the trait method is not `async`, but the call to `key_sign` requires an `async` context, the `futures_executor::block_on` function is used to drive the future to completion. It blocks the current thread, where the tokio runtime is running. The `future` that is run isn't always able to acquire the lock. Since it's hard to figure out when a `Mutex` is dropped, I'm speculating, but the only reasonable explanation I can come up with, is that a tokio task still holds the lock to the vault. Since the tokio runtime is blocked, it cannot make progress on the tasks and the lock is never released, and so the task in the executor is never able to acquire the lock.

### Fix

I'm not sure what the best way to fix this is. I'm not aware of a way to run the task on the currently running tokio runtime. For example, `tokio::runtime::Handle::current().block_on(future)` fails with "Cannot start a runtime from within a runtime. This happens because a function (like block_on) attempted to block the current thread while the thread is being used to drive asynchronous tasks." Generally speaking, using a separate executor to drive the future to completion should be removed for efficiency alone.

This leaves making the `Sign` trait `async` as an option. That would add a dependency on `async_trait` in `identity-core`. There are only two implementations of this trait. However, since this is a fairly low-level method, a lot of calling code would have to become `async` in the `core` and `did` crate, which seems not ideal. For example, this would also require the `identity_core::crypto::signature::traits::core::Signer` trait to become `async`, since it calls a `Sign` implementation. The `Signer` trait is in turn used in `CoreDocument`, making some of its methods `async`. 

## Language version

- Language: Rust
- Version: `1.55.0`

## IOTA Identity version

- Version (version number, commit, or branch): latest dev, `ee4cce5dbff12ad6d5a3ee5e380866ab31f7e7ec`

## Steps To reproduce the bug
Explain how the maintainer can reproduce the bug.

```rust
#[tokio::test]
async fn test_account_shared_ref() -> Result<()> {

  let storage = Stronghold::new(&stronghold_path, Some("my-password")).await?;
  let account: Arc<Account> = Arc::new(Account::with_config(storage, Config::new().testmode(true)).await?);

  let identity1 = account.create_identity(IdentityCreate::default()).await?;
  let iota_did1: IotaDID = identity1.try_did()?.to_owned();

  let identity2 = account.create_identity(IdentityCreate::default()).await?;
  let iota_did2: IotaDID = identity2.try_did()?.to_owned();

  let account_clone = Arc::clone(&account);

  let task = tokio::task::spawn(async move {
    update(account_clone.as_ref(), iota_did1).await;
  });

  update(account.as_ref(), iota_did2).await;

  task.await.unwrap();

  Ok(())
}

async fn update(account: &Account, did: IotaDID) {
  for i in 0..2 {
    account
      .update_identity(&did)
      .create_method()
      .fragment(format!("key-{}", i))
      .apply()
      .await
      .unwrap();
  }
}
```

HenriqueNogara:
# Description of change
In order to avoid a deadlock caused by the `sign` method in `RemoteSign`, `create_signature` and `sign` were changed to async.

closes #441 

## Type of change

- [ ] Documentation Fix

https://github.com/iotaledger/identity.rs/compare/fix/deadlock-sign?expand=1#diff-04c5e8aa10e6c0269fe9db78cd00b75b070e9b60854f7bf23bb3743e48cea160R449

## Change checklist


identity-account/Cargo.toml
```diff
@@ -13,6 +13,7 @@ description = "Secure storage for Decentralized Identifiers and Verifiable Crede
 [dependencies]
 actix = { version = "0.12.0", optional = true }
 async-trait = { version = "0.1", default-features = false }
+crossbeam-utils = { version = "0.8" }
```
HenriqueNogara, line 16: This will be removed

identity-account/src/tests/commands.rs
```diff
@@ -442,3 +446,40 @@ async fn test_delete_method() -> Result<()> {
 
   Ok(())
 }
+#[tokio::test]
+async fn test_account_shared_ref() -> Result<()> {
+  let stronghold_path: PathBuf = "./example-strong.hodl".into();
+
```
PhilippGackstatter, line 452: I don't think we need to explicitly test for the absence of the deadlock.

PhilippGackstatter (approved):
Per our discussion, we can't get around the code duplication, in order to fix the deadlock, so this seems fine for now.

identity-account/src/identity/identity_state.rs
```diff
@@ -45,7 +49,39 @@ use crate::types::KeyLocation;
 type Properties = VerifiableProperties<BaseProperties>;
 type BaseDocument = CoreDocument<Properties, Object, Object>;
 
-pub type RemoteEd25519<'a, T> = JcsEd25519<RemoteSign<'a, T>>;
+pub struct RemoteEd25519;
+
+impl Named for RemoteEd25519 {
+  const NAME: &'static str = "JcsEd25519Signature2020";
+}
```
cycraig, line 56: Nit: maybe we can re-use the name from `JcsEd25519`?

E.g.
```rust
impl Named for RemoteEd25519 {
  const NAME: &'static str = JcsEd25519::NAME;
}
```

identity-account/src/identity/identity_state.rs
```diff
@@ -45,7 +49,39 @@ use crate::types::KeyLocation;
 type Properties = VerifiableProperties<BaseProperties>;
 type BaseDocument = CoreDocument<Properties, Object, Object>;
 
-pub type RemoteEd25519<'a, T> = JcsEd25519<RemoteSign<'a, T>>;
+pub struct RemoteEd25519;
```
cycraig, line 52: Nit: can/should we move the definition and implementation of `RemoteEd25519` from `identity_state.rs` to `remote.rs`?

identity-account/src/identity/identity_state.rs
```diff
@@ -45,7 +49,39 @@ use crate::types::KeyLocation;
...
+
+  pub async fn sign<T, X>(data: &X, remote_key: &RemoteKey<'_, T>) -> Result<SignatureValue>
+  where
+    X: Serialize,
+    T: Storage,
+  {
+    let message: Vec<u8> = data.to_jcs()?;
+    let signature: Vec<u8> = RemoteSign::sign(&message, remote_key).await?;
+    let signature: String = encode_b58(&signature);
+    Ok(SignatureValue::Signature(signature))
```
cycraig, line 82: Discussion: to prevent the complete duplication of fundamental signing operations we could split them up into steps in the `Sign` trait and re-use them rather than copy-paste? E.g. have a `prepare_data` function which converts it to JCS, a `generate_signature` function which actually signs it, and a default `sign` method which calls both of those.

This may be overcomplicating it and doesn't really solve the duplication issue, it just enables some re-use of the code from `JcsEd25519`. Though since so little code is actually duplicated I think it's fine to leave as-is.

cycraig (approved):
Looks good to me.

Minor comments.

PR #500, opened 2021-11-19
Move drop trait to stronghold

### LINKED ISSUE #452: [Bug] Deadlock in `Account` when drop-saving with `Stronghold` storage
## Bug description

When an `Account` is dropped, the `dropsave` is executed, which flushes all pending changes in the storage to disk. Since `drop` is not an `async` method, currently we run `futures_executor::executor::block_on` that runs the `storage.flush_changes` future to completion. With this similar setup as in #441, this can also deadlock when another account concurrently holds a lock in stronghold, for example when updating an identity.

## Fix

We can try to move the dropsave execution to our `Stronghold` wrapper, which would only run when the last of the accounts that use this storage instance have been dropped.

## Language version

- Language: Rust
- Version: 1.55.0

## IOTA Identity version

- Version (version number, commit, or branch): Current `epic/account-refactor` (`4ce6295a96b79747f9e7cd0676319da1b8ab0c37`) (possibly future `dev`).

## Steps To reproduce the bug

```rust
#[tokio::test]
async fn test_account_dropsave_deadlock() -> Result<()> {

  let storage = Stronghold::new(&stronghold_path, Some("my-password")).await?;

  let setup = AccountSetup::new(Arc::new(storage)).config(AccountConfig::new().testmode(true));

  let account: Account = Account::create_identity(setup.clone(), IdentitySetup::default()).await?;
  let mut account2: Account = Account::create_identity(setup.clone(), IdentitySetup::default()).await?;

  let task = tokio::task::spawn(async move {
    for i in 0..10 {
      add_method(&mut account2, i).await;
    }
  });

  let did = account.did().to_owned();
  std::mem::drop(account);

  for _ in 0..10 {
    let account = Account::load_identity(setup.clone(), did.clone()).await?;
    // Force the account to run the dropsave
    // We could also run an update here, but that could deadlock because of #441...
    account.actions.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
  }

  task.await.unwrap();

  Ok(())
}

async fn add_method(account: &mut Account, iteration: usize) {
  account
    .update_identity()
    .create_method()
    .fragment(format!("key-{}", iteration))
    .apply()
    .await
    .unwrap();
}
```

## Expected behaviour

No deadlock occurs, and the account flushes the changes.

## Actual behaviour

Deadlock.

HenriqueNogara:
# Description of change
Moves the drop trait to `Stronhold`.

closes #452 

## Type of change

- [X] Bug fix (a non-breaking change which fixes an issue)
- [ ] Documentation Fix

Multiple accounts with the same `AccountSetup` were created, and performed multiple actions concurrently, before being dropped. 

## Change checklist

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/identity-account/src/account/builder.rs
+++ b/identity-account/src/account/builder.rs
@@ -85,14 +85,6 @@ impl AccountBuilder {
     self
   }
 
-  /// Save the account state on drop.
-  ///
-  /// See the config's [`dropsave`][AccountConfig::dropsave] documentation for details.
-  pub fn dropsave(mut self, value: bool) -> Self {
-    self.config = self.config.dropsave(value);
-    self
-  }
-
   /// Save a state snapshot every N actions.
   pub fn milestone(mut self, value: u32) -> Self {
     self.config = self.config.milestone(value);
--- a/identity-account/src/account/builder.rs
+++ b/identity-account/src/account/builder.rs
@@ -33,7 +33,7 @@ use super::config::AutoSave;
 pub enum AccountStorage {
   Memory,
   #[cfg(feature = "stronghold")]
-  Stronghold(PathBuf, Option<String>),
+  Stronghold(PathBuf, Option<String>, Option<bool>),
   Custom(Arc<dyn Storage>),
 }
 
@@ -112,9 +112,9 @@ impl AccountBuilder {
         self.storage = Some(storage);
       }
       #[cfg(feature = "stronghold")]
-      Some(AccountStorage::Stronghold(snapshot, password)) => {
+      Some(AccountStorage::Stronghold(snapshot, password, dropsave)) => {
         let passref: Option<&str> = password.as_deref();
-        let adapter: Stronghold = Stronghold::new(&snapshot, passref).await?;
+        let adapter: Stronghold = Stronghold::new(&snapshot, passref, dropsave).await?;
 
         if let Some(mut password) = password {
           password.zeroize();
--- a/examples/account/config.rs
+++ b/examples/account/config.rs
@@ -38,7 +38,6 @@ async fn main() -> Result<()> {
     .autosave(AutoSave::Every) // save immediately after every action
     .autosave(AutoSave::Batch(10)) // save after every 10 actions
     .autopublish(true) // publish to the tangle automatically on every update
-    .dropsave(true) // save the account state on drop
     .milestone(4) // save a snapshot every 4 actions
     .storage(AccountStorage::Memory) // use the default in-memory storage
     // configure a mainnet Tangle client with node and permanode
--- a/examples/account/create_did.rs
+++ b/examples/account/create_did.rs
@@ -21,7 +21,6 @@ async fn main() -> Result<()> {
   // It implements best practices for security and is the recommended way of handling private keys.
   let stronghold_path: PathBuf = "./example-strong.hodl".into();
   let password: String = "my-password".into();
-  let dropsave: bool = false;
 
   // Create a new identity using Stronghold as local storage.
   //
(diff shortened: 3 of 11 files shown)


cycraig (commented):
As discussed last week, I think we should keep the `dropsave` configuration option and just move it to `Stronghold`. I'm not sure why one would need to disable `dropsave` (maybe to avoid blocking on drop) but it's an existing feature that doesn't cost us too much to keep for now.

cycraig (approved):
Looks good to me. Thanks for the changes.

identity-account/src/account/builder.rs
```diff
@@ -33,7 +33,7 @@ use super::config::AutoSave;
 pub enum AccountStorage {
   Memory,
   #[cfg(feature = "stronghold")]
-  Stronghold(PathBuf, Option<String>),
+  Stronghold(PathBuf, Option<String>, Option<bool>),
```
PhilippGackstatter, line 36: I'm a bit concerned that it's unclear to users what the `bool` flag is for. We could solve this in a few ways.

1. Add dedicated constructors to `AccountStorage`, e.g.

```rust
impl AccountStorage {
  pub const fn memory() -> AccountStorage {
    Self::Memory
  }

  pub fn stronghold(path: PathBuf, password: Option<String>) -> AccountStorage {
    Self::Stronghold(path, password)
  }

  pub fn custom(storage: Arc<dyn Storage>) -> AccountStorage {
    Self::Custom(storage)
  }
}
```

Now the question is, should we add `dropsave` to `AccountStorage::stronghold` and the enum variant? Probably not, to be ergonomic. If someone wants to disable `dropsave`, which should be the exception, then they can create their own `Stronghold` and use `AccountStorage::Custom`.

2. Don't add the flag at all in `AccountStorage` and use the default when creating the `Stronghold`. Same workaround as in 1 to set the option.
3. Get rid of `AccountStorage` entirely. I don't really see what benefit that enum brings, other than simplifying the build process slightly. Why have two ways of constructing a type. Especially when introducing options, constructing the type directly is more flexible than an enum variant, which we can't easily change breakingly to add more options. Without that type, we could just take an `impl Storage` (or `Arc<dyn Storage>` not sure) in the builder. It's slightly more effort to construct your own `Stronghold::new(&snapshot, passref).await?` but it's as ergonomic as `AccountStorage`. Same for `MemStore`. Setting the dropsave could just be `stronghold.set_dropsave(false)`, which is fine given how rare it probably is. Exposing the type more prominently like that means it's easier to add more setters for options in the future. Even bindings would benefit, because in Wasm only C-style enums are exposable, so we'd have to find a different solution anyway.

I would personally prefer 2 for this PR, or nothing at all if we decide to get rid of `AccountStorage` entirely (3) in a future PR. Any thoughts?

PhilippGackstatter (approved):
Looks good, just one comment on the flag in `AccountStorage` and how we should proceed there. I'll approve for now, since I think the best solution is to fix it in a future PR.

identity-account/src/account/builder.rs
HenriqueNogara: I agree with the third option. I think we can get rid of the enum without adding complexity.

PhilippGackstatter: I think it's unergonomic to set in the constructor every time, e.g. `Stronghold::new("./strong.hodl", "password", true).await?` because almost always you want it to be `true`? What if we add more options in the future? We cannot break the constructor. I think we should choose a solution that is extensible. I'm also fine with

```rust
// Can be extended in the future
pub struct StrongholdOptions {
   dropsave: bool,
}

impl StrongholdOptions {
    // acts like a builder
    fn dropsave(self, value: bool) -> Self { ... }
}

impl Stronghold {
    pub async fn new<'a, T, U>(snapshot: &T, password: U) -> Result<Self>;
    pub async fn new_with_options<'a, T, U>(snapshot: &T, password: U, options: StrongholdOptions) -> Result<Self>;
}
```

We used the `new_with_options` in the past in `IotaDocument` (though slightly differently) and I think that was a good solution.

HenriqueNogara: I really liked that

PR #409, opened 2021-09-22
Rename SecretKey to PrivateKey throughout identity.rs code base.

m-renaud:
# Description of change

Rename `SecretKey` to `PrivateKey` throughout the code base.

- Renames the `SecretKey` to `PrivateKey` in `identity-core/src/crypto/key/key.rs`
- Updates `PrivateKey` variable names from `secret`->`private` and `secret_key`->`private_key`
- Updates comments (unless I've missed some)
- Does **not** rename `Secret` to `Private` in `libjose/src/utils/crypto/key_repr.rs`
- Does **not** rename `MethodSecret` to `MethodPrivate` (can be done in a follow-up if desired)

Note that the `crypto` crate and various key signature algorithms within appear to use `Secret` and `SecretKey` to refer to the private part of the "secret" in some situations. For consistency I did not change any of the types or names in the `libjose` directory.

## Links to any relevant issues

#349

## Type of change

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] Enhancement (a non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

## How the change has been tested

`cargo build`, `cargo test` in both the root, examples, and `bindings/wasm/` directories.

Please let me know if there are other tests I should run.

## Change checklist

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #510, opened 2021-11-25
Enforce Ed25519 private key length

olivereanderson:
# Description of change
The ed25519 private keys generated and parsed by this library all satisfy the definition given in [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2). However it is quite common for other implementations to use another format as explained in [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/). 

Due to the various conventions in use it is important for us to document which private key format our library assumes which is what this PR aims to do.  This is done by both including more documentation, but also by modifying some parsing checks to be more precise. 

## Type of change

- [x] Documentation Fix

The relevant tests have been modified to reflect the new changes. All tests run locally. 

## Change checklist

### DIFF
--- a/identity-core/src/crypto/key/pair.rs
+++ b/identity-core/src/crypto/key/pair.rs
@@ -38,6 +38,10 @@ impl KeyPair {
   }
 
   /// Reconstructs the [`Ed25519`][`KeyType::Ed25519`] [`KeyPair`] from a private key.
+  ///
+  /// The private key must be exactly 32 bytes. Other implementations have different conventions for private keys, but
+  /// we follow RFC 8032. What we and the aforementioned RFC call a private key, is called a *secret key* in the paper [High-speed high-security signatures](http://ed25519.cr.yp.to/ed25519-20110926.pdf).
+  /// The following [blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) renames the aforementioned private key/secret key to be called a *seed*.
   pub fn try_from_ed25519_bytes(private_key_bytes: &[u8]) -> Result<Self, crypto::Error> {
     let private_key_bytes: [u8; ed25519::SECRET_KEY_LENGTH] = private_key_bytes
       .try_into()
--- a/identity-core/src/crypto/key/pair.rs
+++ b/identity-core/src/crypto/key/pair.rs
@@ -39,9 +39,8 @@ impl KeyPair {
 
   /// Reconstructs the [`Ed25519`][`KeyType::Ed25519`] [`KeyPair`] from a private key.
   ///
-  /// The private key must be exactly 32 bytes. Other implementations have different conventions for private keys, but
-  /// we follow RFC 8032. What we and the aforementioned RFC call a private key, is called a *secret key* in the paper [High-speed high-security signatures](http://ed25519.cr.yp.to/ed25519-20110926.pdf).
-  /// The following [blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) renames the aforementioned private key/secret key to be called a *seed*.
+  ///  The private key must be a 32-byte seed in compliance with [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
+/// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
   pub fn try_from_ed25519_bytes(private_key_bytes: &[u8]) -> Result<Self, crypto::Error> {
     let private_key_bytes: [u8; ed25519::SECRET_KEY_LENGTH] = private_key_bytes
       .try_into()
--- a/identity-core/src/crypto/key/pair.rs
+++ b/identity-core/src/crypto/key/pair.rs
@@ -40,7 +40,7 @@ impl KeyPair {
   /// Reconstructs the [`Ed25519`][`KeyType::Ed25519`] [`KeyPair`] from a private key.
   ///
   ///  The private key must be a 32-byte seed in compliance with [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
-/// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
+  /// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
   pub fn try_from_ed25519_bytes(private_key_bytes: &[u8]) -> Result<Self, crypto::Error> {
     let private_key_bytes: [u8; ed25519::SECRET_KEY_LENGTH] = private_key_bytes
       .try_into()
--- a/identity-core/src/crypto/proof/jcs_ed25519.rs
+++ b/identity-core/src/crypto/proof/jcs_ed25519.rs
@@ -85,7 +85,7 @@ mod tests {
   use crate::crypto::Signer as _;
   use crate::crypto::Verifier as _;
   use crate::json;
-  use crate::utils::decode_b58;
+  use crate::utils;
 
   type Signer = JcsEd25519<Ed25519<PrivateKey>>;
 
@@ -103,8 +103,11 @@ mod tests {
   #[test]
   fn test_tvs() {
     for tv in TVS {
-      let public: PublicKey = decode_b58(tv.public).unwrap().into();
-      let private: PrivateKey = decode_b58(tv.private).unwrap().into();
+      let public: PublicKey = utils::decode_b58(tv.public).unwrap().into();
+      // The test vectors have been taken from: [here](https://github.com/decentralized-identity/JcsEd25519Signature2020/tree/master/signature-suite-impls/test-vectors),
+      // and use [GO crypto/ed25519](https://pkg.go.dev/crypto/ed25519#pkg-types)'s convention of representing an Ed25519 private key as: seed (secret key in the original paper) followed by public key (computed from the seed)
+      // We follow the convention from RFC [8032](https://datatracker.ietf.org/doc/html/rfc8032) and need the seed.
+      let private: PrivateKey = (utils::decode_b58(tv.private).unwrap()[..32]).to_vec().into();
       let badkey: PublicKey = b"IOTA".to_vec().into();
 
       let input: Object = Object::from_json(tv.input).unwrap();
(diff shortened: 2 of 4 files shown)


identity-core/src/utils/ed25519.rs
```diff
@@ -29,6 +32,24 @@ pub fn keypair_from_ed25519_private_key(private_key: ed25519::SecretKey) -> (Pub
 }
 
 /// Generates a list of public/private ed25519 keys.
+///
+/// Note that the ed25519 private keys used in this crate comply with the definition given in [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
+/// Other implementations often have another format for ed25519 private keys. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanations.
```
cycraig, line 37:
```suggestion
/// Generates a list of public/private Ed25519 keys.
///
/// See [`generate_ed25519_keypair`].
```

identity-core/src/utils/ed25519.rs
```diff
@@ -29,6 +32,24 @@ pub fn keypair_from_ed25519_private_key(private_key: ed25519::SecretKey) -> (Pub
...
 pub fn generate_ed25519_keypairs(count: usize) -> Result<Vec<(PublicKey, PrivateKey)>> {
   (0..count).map(|_| generate_ed25519_keypair()).collect()
 }
+
+#[cfg(test)]
+mod tests {
+  use super::generate_ed25519_keypair;
+
+  #[test]
+  fn generate_private_key_has_expected_length() {
```
cycraig, line 47:
```suggestion
  fn generate_ed25519_keypair_has_expected_length() {
```

identity-core/src/utils/ed25519.rs
```diff
@@ -18,8 +21,8 @@ pub fn generate_ed25519_keypair() -> Result<(PublicKey, PrivateKey)> {
   Ok((public, private))
 }
 
-/// Reconstructs the ed25519 public key given the private key.
-pub fn keypair_from_ed25519_private_key(private_key: ed25519::SecretKey) -> (PublicKey, PrivateKey) {
+// Reconstructs a pair of public/private ed25519 keys from an ed25519::SecretKey.
+pub(crate) fn keypair_from_ed25519_private_key(private_key: ed25519::SecretKey) -> (PublicKey, PrivateKey) {
```
cycraig, line 25:
```suggestion
// Reconstructs a pair of public/private Ed25519 keys from an ed25519::SecretKey.
pub(crate) fn keypair_from_ed25519_private_key(private_key: ed25519::SecretKey) -> (PublicKey, PrivateKey) {
```

identity-core/src/utils/ed25519.rs
```diff
@@ -8,6 +8,9 @@ use crate::crypto::PublicKey;
 use crate::error::Result;
 
 /// Generates a new pair of public/private ed25519 keys.
+///
+/// Note that the ed25519 private keys used in this crate comply with the definition given in [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
+/// Other implementations often have another format for ed25519 private keys. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanations.
 pub fn generate_ed25519_keypair() -> Result<(PublicKey, PrivateKey)> {
```
cycraig, line 14:
```suggestion
/// Generates a new pair of public/private Ed25519 keys.
///
/// Note that the private key is a 32-byte seed in compliance with [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
/// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
pub fn generate_ed25519_keypair() -> Result<(PublicKey, PrivateKey)> {
```

identity-core/src/crypto/signature/ed25519.rs
```diff
@@ -23,7 +23,11 @@ where
 {
   type Private = T;
   type Output = [u8; SIGNATURE_LENGTH];
-
+  /// signs a message.
+  ///
+  /// The private key is expected to be exactly 32 bytes. Other implementations have different conventions for private
+  /// keys, but we follow RFC 8032. What we and the aforementioned RFC call a private key, is called a *secret key* in the paper [High-speed high-security signatures](http://ed25519.cr.yp.to/ed25519-20110926.pdf).
+  /// The following [blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) renames the aforementioned private key/secret key to be called a *seed*.
```
cycraig, line 30: Nit: clarify that this only generates a signature.

Does the original [High-speed high-security signatures](http://ed25519.cr.yp.to/ed25519-20110926.pdf) paper add anything extra or just confirm where the RFC defintion is from? Can we remove it or should we link it everywhere else too? I found the secret key distinction confusing here, but maybe that's just me.

Aside: we moved away from "secret key" in #409 as it was pointed out by the community to be inaccurate for asymmetric keys.

```suggestion
  /// Computes an EdDSA/Ed25519 signature.
  ///
  ///  The private key must be a 32-byte seed in compliance with [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
/// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
```

identity-core/src/crypto/key/pair.rs
```diff
@@ -38,6 +38,10 @@ impl KeyPair {
   }
 
   /// Reconstructs the [`Ed25519`][`KeyType::Ed25519`] [`KeyPair`] from a private key.
+  ///
+  /// The private key must be exactly 32 bytes. Other implementations have different conventions for private keys, but
+  /// we follow RFC 8032. What we and the aforementioned RFC call a private key, is called a *secret key* in the paper [High-speed high-security signatures](http://ed25519.cr.yp.to/ed25519-20110926.pdf).
+  /// The following [blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) renames the aforementioned private key/secret key to be called a *seed*.
```
cycraig, line 44:
```suggestion
  ///  The private key must be a 32-byte seed in compliance with [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
/// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
```

identity-core/src/crypto/proof/jcs_ed25519.rs
```diff
@@ -103,8 +103,11 @@ mod tests {
   #[test]
   fn test_tvs() {
     for tv in TVS {
-      let public: PublicKey = decode_b58(tv.public).unwrap().into();
-      let private: PrivateKey = decode_b58(tv.private).unwrap().into();
+      let public: PublicKey = utils::decode_b58(tv.public).unwrap().into();
+      // The test vectors have been taken from: [here](https://github.com/decentralized-identity/JcsEd25519Signature2020/tree/master/signature-suite-impls/test-vectors),
+      // and use [GO crypto/ed25519](https://pkg.go.dev/crypto/ed25519#pkg-types)'s convention of representing an Ed25519 private key as: seed (secret key in the original paper) followed by public key (computed from the seed)
+      // We follow the convention from RFC [8032](https://datatracker.ietf.org/doc/html/rfc8032) and need the seed.
+      let private: PrivateKey = (utils::decode_b58(tv.private).unwrap()[..32]).to_vec().into();
```
cycraig, line 110:
```suggestion
      // The test vectors are from [JcsEd25519Signature2020](https://github.com/decentralized-identity/JcsEd25519Signature2020/tree/master/signature-suite-impls/test-vectors),
      // and use [Go crypto/ed25519](https://pkg.go.dev/crypto/ed25519#pkg-types)'s convention of representing an Ed25519 private key as: 32-byte seed concatenated with the 32-byte public key (computed from the seed).
      // We follow the convention from [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2) and extract the 32-byte seed as the private key.
      let public: PublicKey = utils::decode_b58(tv.public).unwrap().into();
      let private: PrivateKey = (utils::decode_b58(tv.private).unwrap()[..32]).to_vec().into();
```

cycraig (approved):
Looks good to me.

Only minor suggestions from my side to try make all the comments consistent.

cycraig:
Should we change the title of this PR to e.g. `Enforce Ed25519 private key length`, since that's the actual breaking change?

identity-core/src/crypto/signature/ed25519.rs
olivereanderson: > Does the original High-speed high-security signatures paper add anything extra or just confirm where the RFC defintion is from?

Yes, I wanted to point out that even the RFC deviates from the original phrasing, but realise now that the documentation for a function is not the right place to do that so I will remove it.  

In principle I like the phrasing in your suggestion, but I am wondering if "32-byte seed" should be replaced with "cryptographically secure 32-byte seed" or is that already implied/clear?

cycraig: What does "cryptographically secure" mean for the seed in this context though, that it was obtained from cryptographically secure pseudorandom number generator, can we even enforce that?

Do we just want to advertise that the private keys we (or rather the zebra library) generates are theoretically from a cryptographically strong source? On reflection we can't even guarantee that since it's based on the operating system I believe.

cycraig: > but I am wondering if "32-byte seed" should be replaced with "cryptographically secure 32-byte seed" or is that already implied/clear?

I think it's implied just by being a cryptographic private key basically.

HenriqueNogara (approved):

identity-core/src/crypto/signature/ed25519.rs
olivereanderson: >What does "cryptographically secure" mean for the seed in this context though, that it was obtained from cryptographically secure pseudorandom number generator, can we even enforce that? 

By cryptographically secure, I believe one usually means fit for cryptographic purposes, which can indeed mean obtained from a cryptographically secure source.  This is indeed something we cannot enforce programmatically (at least not in this context) so it is maybe better not to give anyone the wrong impression of this being checked on our side. 

>Do we just want to advertise that the private keys we (or rather the zebra library) generate are theoretically from a cryptographically strong source? On reflection we can't even guarantee that since it's based on the operating system I believe.

From what I recall the 32 byte seed is actually created by `iota-crypto` before it is passed to the zebra library, and `iota-crypto` does indeed get its input from the operating system.  I don't know if we should say a few words about how the keys are generated, but this is maybe something that can be handled in a separate PR anyway?

If you don't have any objections I will simply apply your suggestion for now.

cycraig: > I don't know if we should say a few words about how the keys are generated, but this is maybe something that can be handled in a separate PR anyway?

Most of these comments are hidden deep in the code. I don't think a comment is necessary but if we were to add one on what key generation we use I would say somewhere like `KeyPair::new` is more visible.

Issue #508, opened 2021-11-24
[Task] Derive `KeyLocation` from inherent fields of methods


PhilippGackstatter:
## Description

Refactor the `KeyLocation` to remove their dependence on generations. The generations currently ensure that two keys with the same fragment can co-exist. To allow co-existence in the future, we can base the key location on the fragment _and_ a hash of the `MethodData`. Two keys with different key materials can share the same fragment without producing the same location. This enables us to derive the key location entirely from the document and does not require additional state keeping.

This can be implemented once https://github.com/iotaledger/stronghold.rs/issues/283 is implemented. Prior to that, we need to migrate to the newer stronghold version, which has a refactored procedure API and refactored error types.

## Motivation

The locations are the last remaining place where generations are needed in the account. To reduce complexity and the amount of state to keep, we want to get rid of the generations.

## To-do list

This issue can be implemented in two separate PRs.
- [ ] Refactor the `KeyLocation`s
    - [ ] Base the `KeyLocation` on fragment and a hash of `MethodData`.
    - [ ] Make use of the new key migration feature in stronghold to allow generating a key in stronghold and migrating it to a location based on its public key (`MethodData`).
    - [ ] Update `CreateIdentity` and `CreateMethod` to make use of this feature, in particular the initial DID generation.
    - [ ] Remove the `IdentityState` (which includes the generations).
    - [ ] Implement `Storage::purge` for `MemStore` and `Stronghold`
    - [ ] Implement deletion of keys of deleted verification methods (while being careful with methods that are used to sign the update that removes that method).

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

Issue #497, opened 2021-11-16
[Task] Detect and abort publishing if document is outdated


PhilippGackstatter:
## Description

We support distributed setups, where multiple accounts, e.g. on different machines, can update the same identity. One of those accounts can make updates to a locally stored document, without publishing them (e.g. with `autopublish = false`). If one of them updates an outdated locally cached document we should ideally detect that case before publishing and abort.

## Motivation

Better support distributed setups.
## To-do list
Create a task-specific to-do list. Please link PRs that match the To-do list item behind the item after it has been submitted.

This list proposes a mechanism for implementing re-application of updates in the IOTA Chrysalis DID method, it is left here for reference.

- [ ] Detect the outdated case during publication and return an error.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

Issue #475, opened 2021-11-01
[Task] Provide low-level API access from within Account


PhilippGackstatter:
## Description

As was discussed in #419, the low-level API will be accessible via "Option 2", a mutable DID Document to be returned by the `Account`. This function will return either an `IotaDocument` directly or an `IdentityState`, depending on #462. 

Additionally, a `publish_unchecked` method is needed, to publish a mutated document directly, to bypass the regular account functionality, like automatic signing and figuring out whether to publish as diff or int. It should be possible to force publish as an integration message, but not as a diff message. This should also allow modifying the document's metadata.

In addition, the `Account` will need to be able to import a DID Document, publish it to the Tangle, storing it in the `Storage` adapter and loading it back into the `Account`. All these actions should try to avoid putting restrictions on the state of the DID Document, such that developers have more freedom than just the `Account`. It needs to be researched how we can allow DID Documents with a state that the `Account` doesn't understand to once again become part of the `Account`. 

## Motivation

Right now, the low-level API for DID Document management does not overlap with the `Account` API. This means that either API must be chosen for the management of the Identity for the entire application. This Issue helps to make the API that is used interchangeable. 

## Resources

* #419 

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Add a method to access the document stored in the account, and be able to modify it (returning a `&mut IotaDocument`)
- [ ] Add a method to publish the document without the typical account functionality and which exposes more options for how to publish
    - [ ] Allow modification of the metadata through this method, too
- [ ] Add import functionality that can take an existing `IotaDocument` with a list of its private keys and make it managable by the `Account`

PR #522, opened 2021-12-02
Enable access to the low-level API from the `Account`

PhilippGackstatter:
# Description of change

These changes enable users to modify the account state using the low-level API. The process would generally follow this simple pattern: Retrieve a copy of the current account state, i.e. the `IotaDocument` or the `ChainState`, and modify it. Then use `Account::update_document_unchecked` or `Account::set_chain_state_unchecked` to overwrite the account's state. The document method is prefixed with `update` instead of `set` because I think it captures what the method is doing better. It updates the document by setting it explicitly rather than using by `update_identity`. It will also most likely become fallible in the future, when that method will calculate the diff between the previous document version and the new version (#497). Setters are usually not fallible. `IdentityState` is not exposed, because it is expected to be removed with #508 anyway.

These methods are highly unsafe. No validation is done in the `_unchecked` methods. Using this correctly comes with some caveats, such as not modifying the account's state in between retrieving the copy and setting it. Users could easily mess up the state by setting a document with a different did, or setting a document with verification methods for which no corresponding key exists in storage. A good understanding of the account internals is required to use this safely, which is why we will need documentation that explains just that. Because these methods are not a threat to memory safety, they are not marked `unsafe`.

The PR also adds `Account::publish_with_options`, which is a safe API but provides options to force publish an update as an integration update, and to choose the method with which to sign the update.

Some alternative designs have been considered, but were discarded in favor of the above solution.

- One option is to give out a mutable reference directly, as was originally suggested in the discussion. The issue is, that the account can not explicitly react to the changed document. This is required in order to store the update made via `update_document_unchecked` as a diff in the future (#497). When giving out a mutable reference, no such method would be required, and thus it would be much more complex to detect when an update was made.
- Another option would be to take a closure in `update_document_unchecked` that is passed a mutable reference to the current document. This way, the account can run logic pre- and post-update. This could be less easy to work with in bindings, however, it is also a viable option, I think.

The PR is marked as a breaking change, because `Account::publish_updates` was renamed to `Account::publish`, in order for `Account::publish_with_options` not having to include `update` in the name.

implements part 1/2 of issue #475 

## Type of change

- [ ] Documentation Fix

Added a superficial test with `test_account_publish_with_options`.

## Change checklist


cycraig (commented):
Changes look good to me. Minor comments.

>  [Account::update_document_unchecked] will also most likely become fallible in the future, when that method will calculate the diff between the previous document version and the new version (#497).

I thought only the `synchronise` or e.g. `reapply_updates` method would be fallible based on the local document state, not `update_document_unchecked`? 

> One option is to give out a mutable reference directly, as was originally suggested in the discussion. The issue is, that the account can not explicitly react to the changed document. 

I am still of the opinion that the `Account` does not need to react to unchecked updates when they are performed, but only at other points like `synchronise`. I think returning a mutable reference might actually be safer because one cannot do completely wrong things like change the DID of the document from a mutable reference (since `IotaDocument` prevents that), while one could pass a completely different DID document from a different identity to `update_document_unchecked`.

> This is required in order to store the update made via update_document_unchecked as a diff in the future ([Task] Enable reapplication of local updates to a changed base document #497).

I still believe any document diff reapplication strategy would not address the complex challenge of merging multiple (possibly conflicting) updates to a document from distributed sources. For example, when trying to merge two updates to the revocation status of a MerkleKeyCollection bitset, I believe the diff string only represents the final state of the bitset not the index changed. Therefore merging a conflicting diff in that case would always overwrite one of the changes, causing a lost update.
The point is that I do not believe it is necessary to create an "incremental" diff in `update_document_unchecked` since it would be no different than a single, "total" document diff generated between the new and old local documents and applied to a remote document on `synchronise`. Unless I'm mistaken?

> Another option would be to take a closure in update_document_unchecked that is passed a mutable reference to the current document. This way, the account can run logic pre- and post-update. This could be less easy to work with in bindings, however, it is also a viable option, I think.

Without knowing whether this can be implemented in current and future bindings I would not go with this option either.

identity-account/src/account/account.rs
```diff
@@ -247,54 +276,66 @@ impl Account {
...
-    }
+      old_state.document()
+    };
+
+    let signing_method: &IotaVerificationMethod = match fragment {
+      Some(fragment) => signing_document
+        .resolve_method_with_scope(fragment, MethodScope::capability_invocation())
+        .ok_or(Error::DIDError(identity_did::Error::MethodNotFound))?,
+      None => signing_document.default_signing_method()?,
+    };
```
cycraig, line 293: This should also probably validate that the method is an `Ed25519VerificationKey2018`, similar to `IotaDocument::check_signing_key`. Right now a `MerkleKeyCollection2021` is rejected by `sign_data` simply because it is unimplemented in the `Account`.

This restriction is from the IOTA DID method specification: https://wiki.iota.org/identity.rs/specs/did/iota_did_method_spec#standardized-verification-method-types

PhilippGackstatter:
On the mutable reference vs setter approach: You're right, I still had the older approach in mind (need to update #497). It doesn't need to calculate the diff, but it needs to store the document. Basically, the dual-document state is not in `dev` yet, but would only be implemented in #497 by storing the latest on-tangle document, and the latest locally changed document (currently only the latter). On synchronize, make the diff and attempt to re-apply to the changed document. So yes, no need for individual diffs. When making an update on an offline device, the unpublished update should be stored when using `update_identity` and when using `update_document_unchecked`, to have consistency between these approaches. There is no easy way to store the update when giving out mutable references. I'm also not sure if the mutable reference works across bindings. The getter/setter approach is more universal.

You're absolutey right, diffing and merging are hard, and detecting merge conflicts in the first place will be the hardest part of #497. For now, I would make changes to the account that are compatible with that issue, and if we end up not being able to support reapplication, we can still make some simplifications.

identity-account/src/account/account.rs
```diff
@@ -162,6 +163,28 @@ impl Account {
...
+  /// Sets the [`ChainState`] for the identity this account manages, **without doing any validation**.
+  ///
+  /// # Warning
+  ///
+  /// This method is dangerous and can easily mess up the internal state,
+  /// making the identity potentially unusable. Only use it if you have
+  /// a good grasp of what the implications are.
+  pub fn set_chain_state_unchecked(&mut self, chain_state: ChainState) {
+    self.chain_state = chain_state
+  }
```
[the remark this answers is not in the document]
PhilippGackstatter: Thanks, made it spicier :hot_pepper:

olivereanderson (approved):
Looks good. I appreciate the thoroughness of your considerations of the various possible approaches.

identity-iota/src/document/iota_document.rs
```diff
@@ -458,6 +458,21 @@ impl IotaDocument {
     self.document.try_resolve_method_mut(query).map_err(Into::into)
   }
 
+  /// Returns the first [`IotaVerificationMethod`] with an `id` property matching the provided `query`
+  /// and the verification relationship specified by `scope`.
+  pub fn resolve_method_with_scope<'query, Q>(&self, query: Q, scope: MethodScope) -> Option<&IotaVerificationMethod>
```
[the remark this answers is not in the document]
PhilippGackstatter: I think returning `Option` for methods that find things fits better with the upcoming error refactor, so I would prefer leaving it as-is.

identity-iota/src/document/iota_document.rs
```diff
@@ -242,6 +242,11 @@ impl IotaDocument {
     Ok(())
   }
 
+  /// Returns whether the given method can be used to sign document updates.
+  pub fn is_signing_method(method: &IotaVerificationMethod) -> bool {
+    method.key_type() == MethodType::Ed25519VerificationKey2018
+  }
```
cycraig, line 248: This should probably use `UPDATE_METHOD_TYPES` or `IotaVerificationMethod::is_signing_method` if that's not removed. I was more thinking that this could be something like `resolve_signing_method` to perform the capability invocation method check too. If not, then `check_signing_method` should be updated to use `is_signing_method` internally to avoid duplicate code.

identity-account/src/account/account.rs
```diff
@@ -367,16 +428,26 @@ impl Account {
...
     let mut diff: DiffMessage = DiffMessage::new(old_doc, new_doc, *previous_message_id)?;
 
-    let method: &IotaVerificationMethod = old_state.document().default_signing_method()?;
+    let signing_method: &IotaVerificationMethod = match signing_method_query {
+      Some(fragment) => old_state
+        .document()
+        .resolve_method_with_scope(fragment, MethodScope::capability_invocation())
+        .ok_or(Error::DIDError(identity_did::Error::MethodNotFound))?,
+      None => old_state.document().default_signing_method()?,
+    };
```
cycraig, line 438: This should use `is_signing_method` too since `sign_data` bypasses the check added in `sign_self`.

cycraig (changes requested):

cycraig:
#526 fixes the failing coverage action.

PhilippGackstatter:
I realized that to make `update_document_unchecked` really consistent with how the account works so far, it needs to mirror `update_identity`. This means it now also publishes (respecting  `autopublish`) and stores the update. If the method is only a setter, then the update isn't stored in storage, contrary to updates made with `update_identity`. I'm hoping this change makes the unchecked and safe API easier to work with, because there are fewer differences.

I've added an example to show _very_ basic usage of the new functionality. This PR also starts cleaning up the low-level API examples, starting with `diff_chain` and `manipulate_did`. Publishing to a diff chain explicitly is discouraged, and might even be removed entirely, depending on what happens with the `Client` type. Similarly, manipulating a DID should be done through the account, rather than through the `IotaDocument`.

identity-iota/src/document/iota_document.rs
PhilippGackstatter: Addressed the duplication. I changed it to only require the `MethodType`, because of problems converting the `method` in `check_signing_method` to an `IotaVerificationMethod` due to generics.

examples/account/unchecked.rs
```diff
@@ -0,0 +1,70 @@
...
+  // Set a custom timestamp.
+  document.set_updated(timestamp);
+
+  // Set a custom property.
+  document.properties_mut().insert("key".into(), "value".into());
+
+  // Update the identity without validation by the account, and publish the result (depending on autopublish).
+  // The responsibility is on us to provide a valid document, which the account can continue to use.
+  // Failing to do so can corrupt the identity; use with caution!
+  account.update_document_unchecked(document).await?;
```
cycraig, line 53: Optional: minor changes to the example comments.

```suggestion
  // Get a copy of the document this account manages.
  // We will apply updates to the document and overwrite the account's current document.
  let mut document: IotaDocument = account.document().clone();

  // Add a custom property to the document.
  document.properties_mut().insert("myCustomPropertyKey".into(), "value".into());

  // Override the updated field timestamp to 24 hours (= 86400 seconds) in the future,
  // because we can. This is usually set automatically by Account::update_identity.
  let timestamp: Timestamp = Timestamp::from_unix(Timestamp::now_utc().to_unix() + 86400);
  document.set_updated(timestamp);

  // Update the identity without validation and publish the result to the Tangle 
  // (depending on the account's autopublish setting).
  // The responsibility is on the caller to provide a valid document which the account 
  // can continue to use. Failing to do so can corrupt the identity; use with caution!
  account.update_document_unchecked(document).await?;
```

examples/account/unchecked.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example account_advanced
```
cycraig, line 4:
```suggestion
//! cargo run --example account_unchecked
```

identity-account/src/account/account.rs
```diff
@@ -212,8 +241,14 @@ impl Account {
...
-  pub async fn publish_updates(&mut self) -> Result<()> {
-    self.publish(true).await?;
+  pub async fn publish(&mut self) -> Result<()> {
+    self.publish_internal(true, PublishOptions::default()).await?;
+
+    Ok(())
+  }
+
+  pub async fn publish_with_options(&mut self, options: PublishOptions) -> Result<()> {
+    self.publish_internal(true, options).await?;
```
cycraig, line 251: Doc comment public function.

```suggestion
  /// Push all unpublished changes to the Tangle in a single message, optionally choosing
  /// the signing key used or forcing an integration chain update.
  ///
  /// See [`PublishOptions`].
  pub async fn publish_with_options(&mut self, options: PublishOptions) -> Result<()> {
    self.publish_internal(true, options).await?;
```

identity-account/src/account/account.rs
```diff
@@ -247,54 +282,66 @@ impl Account {
     &self,
     old_state: &IdentityState,
     new_state: &IdentityState,
+    fragment: &Option<String>,
```
cycraig, line 285: Opinion: rename `signing_method_query` for consistency?

```suggestion
    signing_method_query: &Option<String>,
```

identity-account/src/account/account.rs
```diff
@@ -247,54 +282,66 @@ impl Account {
...
-        .sign_data(self.did(), self.storage(), &location, document)
-        .await?;
-    }
+      (old_state, old_state.document())
+    };
+
+    let signing_method: &IotaVerificationMethod = match fragment {
+      Some(fragment) => signing_document.resolve_signing_method(fragment)?,
+      None => signing_document.default_signing_method()?,
+    };
```
cycraig, line 297: Opinion: using just the `signing_state` might be cleaner? Not sure, completely optional.

E.g.
```rust
    let signing_state: &IdentityState = if self.chain_state().is_new_identity() {
      new_state
    } else {
      old_state
    };

    let signing_method: &IotaVerificationMethod = match fragment {
      Some(fragment) => signing_state.document().resolve_signing_method(fragment)?,
      None => signing_state.document().default_signing_method()?,
    };
```

identity-iota/src/document/iota_document.rs
```diff
@@ -234,14 +235,18 @@ impl IotaDocument {
...
+      return Err(Error::InvalidDocumentSigningMethodType);
     }
 
     Ok(())
   }
 
+  /// Returns whether the given [`MethodType`] can be used to sign document updates.
+  pub fn is_signing_method_type(method_type: MethodType) -> bool {
+    UPDATE_METHOD_TYPES.contains(&method_type)
+  }
```
cycraig, line 248: The other usages of `UPDATE_METHOD_TYPES` should probably be replaced with `IotaDocument::is_signing_method_type`, like in `PublishType` and Account `create_identity`.

cycraig (approved):
Thanks for the changes. I think I agree with making `update_document_unchecked` also save and publish the document.

I'm a bit hesitant about removing the low-level `diff_chain` and `manipulate_did` examples right now since we're keeping parity between the Rust and Wasm examples. I would be fine with removing almost all the low-level examples once the `Account` has been migrated to the Wasm bindings.

examples/README.md
```diff
@@ -29,19 +29,17 @@ The following examples are available for using the basic account (A high-level A
 | 5 | [account_lazy](./account/lazy.rs) | How to take control over publishing DID updates manually, instead of the default automated behavior. |
 | 6 | [account_signing](./account/signing.rs) | Using a DID to sign arbitrary statements and validating them. |
 | 7 | [account_multiple](./account/multiple_identities.rs) | How to create multiple identities from a builder and how to load existing identities into an account. |
-
+| 8 | [account_unchecked](./account/unchecked.rs) | Making an update to the custom properties of a document, by using the account's unchecked methods. |
```
cycraig, line 32: I'm not sure what phrasing we're supposed to use for example descriptions, it's pretty inconsistent.

```suggestion
| 8 | [account_unchecked](./account/unchecked.rs) | How to update the custom properties of a DID document directly by using the account's unchecked methods. |
```

JelleMillenaar (approved):
Reviewed only the example to check how it would work from an application developer perspective, and that looks solid :)

PR #529, opened 2021-12-06
migrate to time 

olivereanderson:
# Description of change
This PR removes all our direct dependencies on the `chrono` crate and introduces a dependency on `time` instead. 

The changes introduced here are mainly visible internally in `identity-core` as I have tried to keep the changes to the `Timestamp` API to a minimum. 

This makes progress on https://github.com/iotaledger/identity.rs/issues/442, but does NOT fix it. 

## Type of change

- [ ] Documentation Fix

Runs under `cargo test`

## Change checklist

### DIFF
--- a/identity-core/Cargo.toml
+++ b/identity-core/Cargo.toml
@@ -13,7 +13,6 @@ description = "The core traits and types for the identity-rs library."
 [dependencies]
 base64 = { version = "0.13", default-features = false, features = ["std"] }
 bs58 = { version = "0.4", default-features = false, features = ["std"] }
-chrono = { version = "0.4", default-features = false, features = ["clock", "std"] }
 hex = { version = "0.4", default-features = false }
 identity-diff = { version = "=0.4.0", path = "../identity-diff", default-features = false }
 multibase = { version = "0.9", default-features = false, features = ["std"] }
@@ -24,6 +23,7 @@ serde_json = { version = "1.0", default-features = false, features = ["std"] }
 strum = { version = "0.21", features = ["derive"] }
 subtle = { version = "2.4", default-features = false }
 thiserror = { version = "1.0", default-features = false }
+time = { version = "0.3.5", features = ["std", "serde", "parsing", "formatting"] }
 typenum = { version = "1.13", default-features = false }
 url = { version = "2.2", default-features = false, features = ["serde"] }
 zeroize = { version = "1.4", default-features = false }
--- a/identity-core/src/common/timestamp.rs
+++ b/identity-core/src/common/timestamp.rs
@@ -7,12 +7,9 @@ use core::fmt::Display;
 use core::fmt::Formatter;
 
 use core::str::FromStr;
-
-use chrono::DateTime;
-use chrono::NaiveDateTime;
-use chrono::SecondsFormat;
-use chrono::Timelike;
-use chrono::Utc;
+use time::format_description::well_known::Rfc3339;
+use time::Duration;
+use time::OffsetDateTime;
 
 use crate::error::Error;
 use crate::error::Result;
@@ -21,44 +18,45 @@ use crate::error::Result;
 #[derive(Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
 #[repr(transparent)]
 #[serde(try_from = "String", into = "String")]
-pub struct Timestamp(DateTime<Utc>);
+pub struct Timestamp(OffsetDateTime);
 
 impl Timestamp {
   /// Parses a `Timestamp` from the provided input string.
   pub fn parse(input: &str) -> Result<Self> {
-    let datetime: DateTime<Utc> = DateTime::parse_from_rfc3339(input)?.into();
-    let datetime: DateTime<Utc> = Self::truncate(datetime);
-
-    Ok(Self(datetime))
+    let offset_date_time = OffsetDateTime::parse(input, &Rfc3339).map_err(time::Error::from)?;
+    Ok(Timestamp(truncate(offset_date_time)))
   }
 
   /// Creates a new `Timestamp` with the current date and time.
   pub fn now_utc() -> Self {
-    Self(Self::truncate(Utc::now()))
-  }
-
-  /// Returns the `Timestamp` as a Unix timestamp.
-  #[allow(clippy::wrong_self_convention)]
-  pub fn to_unix(&self) -> i64 {
-    self.0.timestamp()
-  }
-
-  /// Creates a new `Timestamp` from the given Unix timestamp.
-  pub fn from_unix(seconds: i64) -> Self {
-    Self(DateTime::from_utc(NaiveDateTime::from_timestamp(seconds, 0), Utc))
+    Self(truncate(OffsetDateTime::now_utc()))
   }
 
   /// Returns the `Timestamp` as an RFC 3339 `String`.
   ///
   /// See: https://tools.ietf.org/html/rfc3339
-  #[allow(clippy::wrong_self_convention)]
   pub fn to_rfc3339(&self) -> String {
-    self.0.to_rfc3339_opts(SecondsFormat::Secs, true)
+    // expect is OK, because all constructors provide well defined RFC 3339 compatible OffSetDateTime
+    // if this is to be made fallible instead then we need to change some things in our interface such as the
+    // From<Timestamp> for String implementation.
+    self
+      .0
+      .format(&Rfc3339)
+      .expect("unexpected Timestamp incompatibility with RFC 3339")
   }
 
-  fn truncate(value: DateTime<Utc>) -> DateTime<Utc> {
-    // Safe to unwrap because 0 is a valid nanosecond
-    value.with_nanosecond(0).unwrap()
+  /// Returns the `Timestamp` as a Unix timestamp.
+  pub fn to_unix(&self) -> i64 {
+    self.0.unix_timestamp()
+  }
+
+  /// Creates a new `Timestamp` from the given Unix timestamp.
+  pub fn from_unix(seconds: i64) -> Result<Self> {
+    let offset_date_time = OffsetDateTime::from_unix_timestamp(seconds).map_err(time::error::Error::from)?;
+    if !(0..10_000).contains(&offset_date_time.year()) {
+      return Err(time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into());
+    }
+    Ok(Self(offset_date_time))
   }
 }
 
@@ -110,6 +108,10 @@ impl FromStr for Timestamp {
   }
 }
 
+// truncate the nanoseconds within the second in the stored offset
+fn truncate(offset_date_time: OffsetDateTime) -> OffsetDateTime {
+  offset_date_time - Duration::nanoseconds(offset_date_time.nanosecond() as i64)
+}
 #[cfg(test)]
 mod tests {
   use crate::common::Timestamp;
--- a/identity-account/src/tests/updates.rs
+++ b/identity-account/src/tests/updates.rs
@@ -72,8 +72,8 @@ async fn test_create_identity() -> Result<()> {
   assert!(account.load_state().await.is_ok());
 
   // Ensure timestamps were recently set.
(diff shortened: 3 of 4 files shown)


identity-core/src/common/timestamp.rs
```diff
@@ -21,44 +18,45 @@ use crate::error::Result;
...
+  /// Returns the `Timestamp` as a Unix timestamp.
+  pub fn to_unix(&self) -> i64 {
+    self.0.unix_timestamp()
+  }
+
+  /// Creates a new `Timestamp` from the given Unix timestamp.
+  pub fn from_unix(seconds: i64) -> Result<Self> {
+    let offset_date_time = OffsetDateTime::from_unix_timestamp(seconds).map_err(time::error::Error::from)?;
+    if !(0..10_000).contains(&offset_date_time.year()) {
+      return Err(time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into());
```
cycraig, line 57: Maybe add a comment that this check is to prevent runtime errors in `to_rfc3339`?

This also needs unit tests for posterity. `OffsetDateTime::from_unix_timestamp` will already error on any timestamp outside the range `[-377705116800, 253402300799]` (i.e. `["-9999-01-01 0:00:00.0 +00:00:00", "9999-12-31T23:59:59Z"]`). So test some values in that range.

```suggestion
    // Reject years outside of the range 0000AD - 9999AD per Rfc3339
    // upfront to prevent conversion errors in to_rfc3339().
    // https://datatracker.ietf.org/doc/html/rfc3339#section-1
    if !(0..10_000).contains(&offset_date_time.year()) {
      return Err(time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into());
```

identity-core/src/common/timestamp.rs
```diff
@@ -21,44 +18,45 @@ use crate::error::Result;
...
 impl Timestamp {
   /// Parses a `Timestamp` from the provided input string.
   pub fn parse(input: &str) -> Result<Self> {
-    let datetime: DateTime<Utc> = DateTime::parse_from_rfc3339(input)?.into();
-    let datetime: DateTime<Utc> = Self::truncate(datetime);
-
-    Ok(Self(datetime))
+    let offset_date_time = OffsetDateTime::parse(input, &Rfc3339).map_err(time::Error::from)?;
+    Ok(Timestamp(truncate(offset_date_time)))
   }
```
cycraig, line 28: This needs to convert the timezone/offset to UTC as per the DID specification and prior behaviour.

E.g.
```rust
  /// Parses a `Timestamp` from the provided input string, normalized to UTC+00:00 with fractional
  /// seconds truncated.
  ///
  /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
  pub fn parse(input: &str) -> Result<Self> {
    let offset_date_time: OffsetDateTime = OffsetDateTime::parse(input, &Rfc3339)
      .map_err(time::Error::from)?
      .to_offset(UtcOffset::UTC);
    Ok(Timestamp(truncate_fractional_seconds(offset_date_time)))
  }
```

identity-core/src/common/timestamp.rs
```diff
@@ -21,44 +18,45 @@ use crate::error::Result;
...
-  #[allow(clippy::wrong_self_convention)]
-  pub fn to_unix(&self) -> i64 {
-    self.0.timestamp()
-  }
-
-  /// Creates a new `Timestamp` from the given Unix timestamp.
-  pub fn from_unix(seconds: i64) -> Self {
-    Self(DateTime::from_utc(NaiveDateTime::from_timestamp(seconds, 0), Utc))
+    Self(truncate(OffsetDateTime::now_utc()))
   }
```
cycraig, line 33: Doc comment suggestion:

```rust
  /// Creates a new `Timestamp` with the current date and time, normalized to UTC+00:00 with
  /// fractional seconds truncated.
  ///
  /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
  pub fn now_utc() -> Self {
    Self(Self::truncate(Utc::now()))
  }
```

identity-core/src/common/timestamp.rs
```diff
@@ -110,6 +108,10 @@ impl FromStr for Timestamp {
...
 }
 
+// truncate the nanoseconds within the second in the stored offset
+fn truncate(offset_date_time: OffsetDateTime) -> OffsetDateTime {
+  offset_date_time - Duration::nanoseconds(offset_date_time.nanosecond() as i64)
+}
 #[cfg(test)]
 mod tests {
   use crate::common::Timestamp;
```
cycraig, line 117: Needs more unit tests.

E.g. that the offset is normalised to UTC.
```rust
  #[test]
  fn test_parse_valid_offset_normalised() {
    let original = "1937-01-01T12:00:27.87+00:20";
    let expected = "1937-01-01T11:40:27Z";
    let timestamp = Timestamp::parse(original).unwrap();

    assert_eq!(timestamp.to_rfc3339(), expected);
  }
```

cycraig (changes requested):
Thanks for the change.

`parse` needs to be updated to normalise the timezone to UTC+0 as per the DID specification and `Timestamp` needs more unit tests in general.

cycraig:
This isn't really a breaking change is it? No APIs are changed since we didn't expose `chrono` from `Timestamp` unless I'm mistaken?

olivereanderson:
> This isn't really a breaking change is it? No APIs are changed since we didn't expose `chrono` from `Timestamp` unless I'm mistaken?

There are a few breaking changes introduced in this PR:  One can no longer convert a `chrono::Error` to `Error` [caused by this change](https://github.com/iotaledger/identity.rs/pull/529/files#diff-631f7126efbbcb42d913186be3cf616160d856a6424b819f00e96371fa6207f4L50) and the aforementioned change also replaces the wrapped value which is also part of the public API as far as I know. Furthermore the `from_unix` method is [now fallible](https://github.com/iotaledger/identity.rs/pull/529/files#diff-4f63f539d24a115bfdd1a8cb8b6665748dd6bb09d4779e8e2ca3e457662c3959L47).

cycraig:
Good points, I missed `from_unix` in particular.

identity-core/Cargo.toml
```diff
@@ -24,6 +23,7 @@ serde_json = { version = "1.0", default-features = false, features = ["std"] }
 strum = { version = "0.21", features = ["derive"] }
 subtle = { version = "2.4", default-features = false }
 thiserror = { version = "1.0", default-features = false }
+time = { version = "0.3.5", features = ["std", "serde", "parsing", "formatting"] }
```
cycraig, line 26: I know we actually use the default feature set, we should just be consistent to make it easier to determine that we aren't pulling in unnecessary crates or code bloat from features etc. (some of the dependencies I've added also need this change).

```suggestion
time = { version = "0.3.5", default-features = false, features = ["std", "serde", "parsing", "formatting"] }
```

identity-core/src/common/timestamp.rs
```diff
@@ -110,12 +123,40 @@ impl FromStr for Timestamp {
...
+fn truncate_fractional_seconds(offset_date_time: OffsetDateTime) -> OffsetDateTime {
+  offset_date_time - Duration::nanoseconds(offset_date_time.nanosecond() as i64)
+}
+
 #[cfg(test)]
 mod tests {
+  use quickcheck::Arbitrary;
+  use quickcheck_macros::quickcheck;
+  const LAST_VALID_UNIX_TIMESTAMP: i64 = 253402300799;
+  const FIRST_VALID_UNIX_TIMESTAMP: i64 = -62167219200;
```
cycraig, line 136:
```suggestion
  const LAST_VALID_UNIX_TIMESTAMP: i64 = 253402300799;  // 9999-12-31T23:59:59Z
  const FIRST_VALID_UNIX_TIMESTAMP: i64 = -62167219200; // 0000-01-01T00:00:00Z
```

identity-core/src/common/timestamp.rs
```diff
@@ -110,12 +123,40 @@ impl FromStr for Timestamp {
...
+      let markers: Vec<i64> = (FIRST_VALID_UNIX_TIMESTAMP..LAST_VALID_UNIX_TIMESTAMP)
+        .step_by(interval_length as usize)
+        .collect();
+      let random_marker = *g.choose(&markers).unwrap();
+      // add another randomly chosen number between 1 and 1000 for extra randomness
+      let first_thousand: Vec<i64> = (1..=1000).collect();
+      let random_number = *g.choose(&first_thousand).unwrap();
+      Self(random_marker + random_number)
+    }
+  }
```
cycraig, line 158: Wow `quickcheck` is a lot of work. I would have just shoved it in a proptest range. You don't have to change this, I just prefer `proptest`.

E.g.
```rust
  proptest::proptest! {
    #[test]
    fn test_from_unix_to_rfc3339_valid_no_panic(seconds in FIRST_VALID_UNIX_TIMESTAMP..=LAST_VALID_UNIX_TIMESTAMP) {
      let timestamp = Timestamp::from_unix(seconds).unwrap();
      let expected_length = "dddd-dd-ddTdd:dd:ddZ".len();
      assert_eq!(timestamp.to_rfc3339().len(), expected_length);
    }

    #[test]
    fn test_from_unix_no_panic(seconds: i64) {
      let _ = Timestamp::from_unix(seconds);
    }
  }
```

Could also add a parse fuzz test then (in the same `proptest` block).

E.g.
```rust
    #[test]
    fn test_parse_no_panic(s in "\\PC*") {
      let _ = Timestamp::parse(s);
    }
```

identity-core/src/common/timestamp.rs
```diff
@@ -21,44 +19,59 @@ use crate::error::Result;
...
-  #[allow(clippy::wrong_self_convention)]
   pub fn to_rfc3339(&self) -> String {
-    self.0.to_rfc3339_opts(SecondsFormat::Secs, true)
+    // expect is OK, because all constructors provide well defined RFC 3339 compatible OffSetDateTime
+    // if this is to be made fallible instead then we need to change some things in our interface such as the
+    // From<Timestamp> for String implementation.
+    self
+      .0
+      .format(&Rfc3339)
+      .expect("unexpected Timestamp incompatibility with RFC 3339")
```
cycraig, line 54: As much as I hate unwrap/expect it seems to fit here. Small suggestion to simplify comment.

```suggestion
    // expect is okay, constructors ensure RFC 3339 compatible timestamps.
    // Making this fallible would break our interface such as From<Timestamp> for String.
    self
      .0
      .format(&Rfc3339)
      .expect("Timestamp incompatible with RFC 3339")
```

identity-core/src/common/timestamp.rs
```diff
@@ -21,44 +19,59 @@ use crate::error::Result;
...
-    value.with_nanosecond(0).unwrap()
+  /// Returns the `Timestamp` as a Unix timestamp.
+  pub fn to_unix(&self) -> i64 {
+    self.0.unix_timestamp()
+  }
+
+  /// Creates a new `Timestamp` from the given Unix timestamp.
+  ///
+  /// # Errors
+  /// Fails if `seconds` is outside of the interval [-62167219200,253402300799].
```
cycraig, line 65: Optional, trying to give more context.

```suggestion
  /// Creates a new `Timestamp` from the given Unix timestamp.
  ///  
  /// The timestamp must be in the valid range for [RFC 3339](https://tools.ietf.org/html/rfc3339).
  ///
  /// # Errors
  /// [`Error::InvalidTimestamp`] if `seconds` is outside of the interval [-62167219200,253402300799].
```

cycraig (approved):
Thanks for the changes. Minor comments.

identity-core/src/common/timestamp.rs
olivereanderson: I also think `proptest` fits better here. I went with `quickcheck` because I saw that we were already using that in this crate, but since you would also prefer proptest, I will change this. 

>Edit2: doesn't the random selection exclude numbers that are not within 1000 of any of the intervals?

Yes it does. As far as I can tell one either has to allocate a vector with the entire range to make it available to `quickcheck`'s `choose` method, or bring in your own random generator.

PR #541, opened 2021-12-10
Fix `Timestamp` in the Wasm bindings

olivereanderson:
# Description of change
After migrating to `time` we got the following error `panicked at 'time not implemented on this platform', library/std/src/sys/wasm/../unsupported/time.rs:39:9` when running the node examples. This PR fixes that. 

## Type of change

- [ ] Documentation Fix

Runs successfully under 
`npm run build && npm run build:examples && npm run test:node && npm run test:browser` called from the `bindings/wasm` directory. Also passes `cargo test` called from the root directory. 

## Change checklist

### DIFF
--- a/identity-core/Cargo.toml
+++ b/identity-core/Cargo.toml
@@ -15,7 +15,7 @@ base64 = { version = "0.13", default-features = false, features = ["std"] }
 bs58 = { version = "0.4", default-features = false, features = ["std"] }
 hex = { version = "0.4", default-features = false }
 identity-diff = { version = "=0.4.0", path = "../identity-diff", default-features = false }
-js-sys = { version = "0.3.55", optional = true }
+js-sys = { version = "0.3.55", default-features = false, optional = true }
 multibase = { version = "0.9", default-features = false, features = ["std"] }
 roaring = { version = "0.7", default-features = false }
 serde = { version = "1.0", default-features = false, features = ["std", "derive"] }
--- a/identity-core/Cargo.toml
+++ b/identity-core/Cargo.toml
@@ -15,6 +15,7 @@ base64 = { version = "0.13", default-features = false, features = ["std"] }
 bs58 = { version = "0.4", default-features = false, features = ["std"] }
 hex = { version = "0.4", default-features = false }
 identity-diff = { version = "=0.4.0", path = "../identity-diff", default-features = false }
+js-sys = { version = "0.3.55", optional = true }
 multibase = { version = "0.9", default-features = false, features = ["std"] }
 roaring = { version = "0.7", default-features = false }
 serde = { version = "1.0", default-features = false, features = ["std", "derive"] }
@@ -38,6 +39,9 @@ proptest = "1.0.0"
 quickcheck = { version = "1.0" }
 quickcheck_macros = { version = "1.0" }
 rand = { version = "0.8" }
+
+[features]
+wasm = ["js-sys"]
 
 [package.metadata.docs.rs]
 # To build locally:
--- a/identity-core/src/common/timestamp.rs
+++ b/identity-core/src/common/timestamp.rs
@@ -37,8 +37,22 @@ impl Timestamp {
   /// fractional seconds truncated.
   ///
   /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
+  #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasm")))]
   pub fn now_utc() -> Self {
     Self(truncate_fractional_seconds(OffsetDateTime::now_utc()))
+  }
+
+  /// Creates a new `Timestamp` with the current date and time, normalized to UTC+00:00 with
+  /// fractional seconds truncated.
+  ///
+  /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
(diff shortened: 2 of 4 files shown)


identity-core/src/common/timestamp.rs
```diff
@@ -37,10 +37,24 @@ impl Timestamp {
...
+  /// fractional seconds truncated.
+  ///
+  /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
+  #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasm"))]
+  pub fn now_utc() -> Self {
+    let date = js_sys::Date::new_0();
+    let milliseconds_since_unix_epoch: u64 = date.get_time() as u64;
+    let seconds = milliseconds_since_unix_epoch / 1000;
+    // expect is okay, since we assume that the current time is less than 9999AD
+    Self::from_unix(seconds as i64).expect("Failed to accurately convert host Timestamp")
```
cycraig, line 55: The `Date` timestamp could be negative (yes, only if someone screws with their system datetime) and we convert to i64 at the end, so shouldn't the intermediate conversion also be i64 instead of u64?

```suggestion
    let milliseconds_since_unix_epoch: i64 = js_sys::Date::new_0().get_time() as i64;
    let seconds: i64 = milliseconds_since_unix_epoch / 1000;
    // expect is okay, we assume the current time is between 0AD and 9999AD
    Self::from_unix(seconds).expect("Timestamp failed to convert system datetime")
```

identity-core/Cargo.toml
```diff
@@ -15,6 +15,7 @@ base64 = { version = "0.13", default-features = false, features = ["std"] }
 bs58 = { version = "0.4", default-features = false, features = ["std"] }
 hex = { version = "0.4", default-features = false }
 identity-diff = { version = "=0.4.0", path = "../identity-diff", default-features = false }
+js-sys = { version = "0.3.55", optional = true }
```
cycraig, line 18: Even if a crate doesn't have features it just helps to make it easier to check that we're not pulling stuff in unnecessarily.

```suggestion
js-sys = { version = "0.3.55", default-features = false, optional = true }
```

cycraig (approved):
Tests and Wasm examples pass my side.

identity-core/src/common/timestamp.rs
```diff
@@ -37,10 +37,24 @@ impl Timestamp {
...
   }
 
+  /// Creates a new `Timestamp` with the current date and time, normalized to UTC+00:00 with
+  /// fractional seconds truncated.
+  ///
+  /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
+  #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasm"))]
+  pub fn now_utc() -> Self {
+    let date = js_sys::Date::new_0();
+    let milliseconds_since_unix_epoch: u64 = date.get_time() as u64;
```
PhilippGackstatter, line 52: Nit: You could use [`Date::now`](https://docs.rs/js-sys/latest/js_sys/struct.Date.html#method.now) here.

identity-core/src/common/timestamp.rs
```diff
@@ -37,10 +37,24 @@ impl Timestamp {
...
+  #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasm")))]
   pub fn now_utc() -> Self {
     Self(truncate_fractional_seconds(OffsetDateTime::now_utc()))
   }
 
+  /// Creates a new `Timestamp` with the current date and time, normalized to UTC+00:00 with
+  /// fractional seconds truncated.
+  ///
+  /// See the [`datetime` DID-core specification](https://www.w3.org/TR/did-core/#production).
+  #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasm"))]
```
PhilippGackstatter, line 49: I think I would prefer having that fixed in the wasm bindings rather than in core. Any advantage of having it here? I guess there is this 9999AD invariant that you want to uphold, which is easier to state here because of the "proximity" to the other code that upholds that invariant. But aside from that, I would prefer having binding-specific code in the bindings directory, otherwise this sets a sort of precedent on where binding-specific issues are fixed.

PhilippGackstatter (changes requested):

identity-core/src/common/timestamp.rs
cycraig: It needs to be in `identity-core` because `IotaDocument::new()` calls `Timestamp::now_utc()` internally (and probably in other places in the code too). Changing only external `WasmTimestamp::now_utc()` calls in the Wasm bindings would not fix the current problem as far as I know.

abdulmth (approved):

PhilippGackstatter (approved):

identity-core/src/common/timestamp.rs
PhilippGackstatter: Ah thanks, that makes sense.

olivereanderson: @PhilippGackstatter when I find a bit more time I will see if I can make the `time` crate compatible with `wasm-bindgen` itself and then we can remove this code. For now I think it is OK to have it here for the reasons @cycraig suggested.

PR #544, opened 2021-12-13
Add account synchronization method

### LINKED ISSUE #140: Account API
Add an API for managing Identity-related data

### TODO

- [x] Add an optional background-sync process
- [x] Add automatic auth-chain publishing
  - [ ] Ability to change publish threshold
  - [x] Ability to publish with/without key change

HenriqueNogara:
# Description of change
Adds a function to synchronize the account state with the changes found on the tangle.

closes #140

## Type of change
- [ ] Documentation Fix

Case 0: Nothing found on the tangle (local state does not change);
Case 1: Tangle and account are already in sync (local state does not change);
Case 2: Account is ahead of the tangle (local state does not change);
Case 3: Integration msg has been published on the tangle (local state must be updated);
Case 4: Diff msg has been found on the tangle (local state must be updated);

## Change checklist


identity-account/src/account/account.rs
```diff
@@ -471,6 +472,29 @@ impl Account {
     Ok(())
   }
 
+  /// Fetches the latest changes for the account did from the tangle and
+  /// updates the local state
+  pub async fn synchronize_state(&mut self) -> Result<()> {
```
HenriqueNogara, line 477: Should we add the force flag in case the user wants to save the incoming changes in the storage?

identity-account/src/account/account.rs
```diff
@@ -471,6 +472,29 @@ impl Account {
...
+    }
+    // Overwrite the current state with the most recent document
+    self.state.set_document(document_chain.current().clone());
+    self
+      .chain_state
+      .set_last_integration_message_id(*document_chain.integration_message_id());
+    self
+      .chain_state
+      .set_last_diff_message_id(*document_chain.diff_message_id());
+    Ok(())
```
olivereanderson, line 495: The call to `clone` on line 488 can be avoided as follows: 
```suggestion 
    let iota_did: &IotaDID = self.did();
    let mut document_chain: DocumentChain = self.client_map.read_document_chain(iota_did).await?;
    // Checks if the local document is up to date
    if document_chain.integration_message_id() == self.chain_state.last_integration_message_id()
      && (document_chain.diff().is_empty()
        || document_chain.diff_message_id() == self.chain_state.last_diff_message_id())
    {
      return Ok(());
    }
    // Overwrite the current state with the most recent document
    self
      .chain_state
      .set_last_integration_message_id(*document_chain.integration_message_id());
    self
      .chain_state
      .set_last_diff_message_id(*document_chain.diff_message_id());
    std::mem::swap(self.state.document_mut(), document_chain.current_mut());
    Ok(())
```

identity-account/src/identity/identity_state.rs
```diff
@@ -81,6 +81,10 @@ impl IdentityState {
     &mut self.document
   }
 
+  pub fn set_document(&mut self, document: IotaDocument) {
+    self.document = document;
```
olivereanderson, line 85: Discussion: Do we really want trivial setters?
Hiding the mutation of a field inside a function can be good for future proofing, but introduces some indirection when for those reading/auditing code (one has to check that the `set_document` function does not have any other side effects or does other non-trivial things in addition to checking the field). Hence if one is sure about the internal presentation it is perhaps better to just declare that field public.  

Note also that in this case we already have `document_mut` and then one can just compose that with [mem::replace](https://doc.rust-lang.org/std/mem/fn.replace.html) to obtain the same functionality as this setter, so there is even less reason to introduce an extra method here.

HenriqueNogara: Personally, I don't mind having this setter. I find that having to use `document_mut` + `mem::replace` would be a bit annoying.

PhilippGackstatter: As a compromise, I would be perfectly fine with making such fields `pub(crate)`, but not `pub`. With the latter, everything is even more likely to be a breaking change, and we retain less control of the code. Also note that `IdentityState` will likely be removed with #508 anyway.

PhilippGackstatter (changes requested):
I think we need to always store state after synchronization, and we may want to rename synchronize to fetch.

identity-account/src/identity/identity_state.rs
PhilippGackstatter: As a compromise, I would be perfectly fine with making such fields pub(crate), but not pub. With the latter, everything is even more likely to be a breaking change, and we retain less control of the code. Also note that IdentityState will likely be removed with #508 anyway.

olivereanderson: If one wants to mutate a field defined in the same crate I would recommend declaring the field `pub(crate)` rather than introducing a setter.  For fields one might want to mutate from outside of the crate in which they are defined it can be okay to define a setter if one thinks the internal representation might change in the future, but I would argue for another name in that case, because if the internal representation does in fact change then `set_field_name` becomes a bit misleading. 

Regardless of our preferences, with the `mem::swap` change I suggested above I don't think this method is used (by us) anymore (or am I wrong about that?).

identity-account/src/account/account.rs
```diff
@@ -257,6 +258,30 @@ impl Account {
...
+      return Ok(());
+    }
+    // Overwrite the current state with the most recent document
+    self
+      .chain_state
+      .set_last_integration_message_id(*document_chain.integration_message_id());
+    self
+      .chain_state
+      .set_last_diff_message_id(*document_chain.diff_message_id());
+    std::mem::swap(self.state.document_mut(), &mut document_chain.current_mut().document);
```
olivereanderson, line 280: Nit: 
`document_chain` does not need to be mutable before line 280. I would prefer to first declare it as immutable on line 265, then redeclare it as mutable just before line 280 as this makes it easier for the reader to understand what is happening.

olivereanderson (approved):

identity-account/src/account/account.rs
```diff
@@ -257,6 +258,30 @@ impl Account {
     Ok(())
   }
 
+  /// Fetches the latest changes for the account did from the tangle and
+  /// updates the local state
```
PhilippGackstatter, line 262:
```suggestion
  /// Fetches the latest changes from the tangle and **overwrites** the local document.
  /// 
  /// If a DID is managed from distributed accounts, this should be called before making changes
  /// to the identity, to avoid publishing updates that would be ignored.
```

identity-account/src/account/account.rs
```diff
@@ -257,6 +258,30 @@ impl Account {
...
+    }
+    // Overwrite the current state with the most recent document
+    self
+      .chain_state
+      .set_last_integration_message_id(*document_chain.integration_message_id());
+    self
+      .chain_state
+      .set_last_diff_message_id(*document_chain.diff_message_id());
+    std::mem::swap(self.state.document_mut(), &mut document_chain.current_mut().document);
+    self.save(true).await?;
```
PhilippGackstatter, line 281:
```suggestion
    self.increment_actions();
    self.store_state().await?;
```

We have to call `store_state` in order to write the changes into storage. `store_state` respects autosave settings, so we can increment actions for autosave to properly kick in, when appropriate.

We should probably rename `Account::save` to `Account::flush_changes` or something, to make these methods more distinct.

identity-account/src/tests/account.rs
```diff
@@ -322,3 +335,172 @@ async fn test_account_publish_options_force_integration() -> Result<()> {
...
+
+      account.fetch_state().await.unwrap();
+      assert!(account.state().document().properties().contains_key("foo"));
+      assert!(account.state().document().properties().contains_key("bar"));
+      assert_eq!(
+        account.chain_state().last_integration_message_id(),
+        chain.integration_message_id()
+      );
+      assert_eq!(account.chain_state().last_diff_message_id(), chain.diff_message_id());
+      Ok(())
```
PhilippGackstatter, line 425:
```suggestion
      assert_eq!(account.chain_state().last_diff_message_id(), chain.diff_message_id());
      
      // Ensure state was written into storage.
      let storage_state: IdentityState = account.load_state().await?;
      assert_eq!(&storage_state, account.state());
      
      Ok(())
```

Test that fetched state was written into storage. Maybe also set `AutoSave::Every` explicitly in the `AccountConfig` (line 394) to make it explicit that this test requires that setting.

PhilippGackstatter (changes requested):
Thanks for the changes! I think we should add a check in the tests that the state was successfully stored.

PhilippGackstatter (approved):
Thanks again for the changes, looks good!

PR #564, opened 2021-12-20
Feat/VC and VP specified error codes and validation options 

olivereanderson:
# Description of change
Adds specific error codes for what went wrong during VC and VP validation. Also provides the ability to specify which deficiencies (expired, not yet active, etc) are acceptable during validation. 

## Type of change

- [ ] Documentation Fix

## Change checklist


identity-iota/src/credential/validator.rs
```diff
@@ -87,53 +115,129 @@ impl<'a, R: TangleResolve> CredentialValidator<'a, R> {
...
-    let credential_verified: bool = issuer_doc.document.document.verify_data(&credential, options).is_ok();
+    issuer.document.document.verify_data(&credential, signature_options)?;
 
-    // Check if all subjects have valid signatures
-    let subjects_verified: bool = subjects.values().all(|subject| subject.verified);
+    // Verify the expiration date
+    if let Some(expiration_date) = credential.expiration_date {
+      if Timestamp::now_utc() > expiration_date {
+        deficiency_handler(CredentialDeficiencyEvent::Expired(expiration_date))?;
+        // encountered_refutation_categories.insert(CredentialRefutationCategory::Expired);
```
PhilippGackstatter, line 207: Leftover comment.

identity-iota/src/credential/validator.rs
```diff
@@ -87,53 +115,129 @@ impl<'a, R: TangleResolve> CredentialValidator<'a, R> {
...
+      }
+    }
 
-    // The credential is truly verified if all associated documents are verified
-    let verified: bool = issuer_doc.verified && credential_verified && subjects_verified;
+    // Verify that the credential has been activated
+    if Timestamp::now_utc() < credential.issuance_date {
+      deficiency_handler(CredentialDeficiencyEvent::Dormant(credential.issuance_date))?;
+      // encountered_refutation_categories.insert(CredentialRefutationCategory::Dormant);
```
PhilippGackstatter, line 214: Leftover comment.

identity-iota/src/credential/validator.rs
```diff
@@ -87,53 +115,129 @@ impl<'a, R: TangleResolve> CredentialValidator<'a, R> {
...
     credential: Credential<T>,
-    options: VerifierOptions,
+    signature_verification_options: VerifierOptions,
+    allowed_deficiencies: CredentialDeficiencySet,
   ) -> Result<CredentialValidation<T>>
   where
     T: Serialize,
   {
+    // setup deficiency event handler
+    let mut encountered_deficiencies = CredentialDeficiencySet::empty();
```
PhilippGackstatter, line 125: Do we actually need to return that when the caller has already stated which deficiencies are (not) allowed?

identity-iota/src/credential/validator.rs
```diff
@@ -19,28 +30,29 @@ use crate::error::Result;
...
-  pub issuer: DocumentValidation,
-  pub subjects: BTreeMap<String, DocumentValidation>,
-  pub verified: bool,
+// used internally during credential validation
+#[allow(clippy::large_enum_variant)] // The smaller variants occur at most once each
+enum CredentialDeficiencyEvent<'a> {
+  // the Url corresponds to the url of the deactivated document
+  DeactivatedSubjectDocument((&'a Url, DocumentValidation<Deactivated>)),
+  Expired(Timestamp),
+  Dormant(Timestamp),
```
PhilippGackstatter, line 39: Looks like those timestamps aren't really used? The deficiency handler just converts it into the `CredentialDeficiency` anyway without looking at the timestamp.

identity-iota/src/credential/validator.rs
```diff
@@ -19,28 +30,29 @@ use crate::error::Result;
...
 }
-
-#[derive(Clone, Debug, PartialEq, Serialize)]
-pub struct PresentationValidation<T = Object, U = Object> {
-  pub presentation: Presentation<T, U>,
-  pub holder: DocumentValidation,
-  pub credentials: Vec<CredentialValidation<U>>,
-  pub verified: bool,
+impl CredentialDeficiencyEvent<'_> {
+  fn associated_category(&self) -> CredentialDeficiency {
```
PhilippGackstatter, line 42:
```suggestion
  fn as_deficiency(&self) -> CredentialDeficiency {
```
Or `to_deficiency`? Not sure.

identity-iota/src/credential/validated_data.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub(super) mod document_state {
+  // Used to parameterise whether a resolved document is active.
+  pub trait Sealed {}
+  #[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+  pub struct Active;
+  #[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+  pub struct Deactivated;
+  impl Sealed for Active {}
+  impl Sealed for Deactivated {}
```
PhilippGackstatter, line 29: Not sure I see the advantage of encoding the state of the document in the type rather than in a `bool`. What is the benefit of this approach?

identity-iota/src/credential/deficiencies.rs
```diff
@@ -0,0 +1,188 @@
...
+impl FromIterator<CredentialDeficiency> for CredentialDeficiencySet {
+  fn from_iter<T: IntoIterator<Item = CredentialDeficiency>>(iter: T) -> Self {
+    let mut deficiencies = Self::empty();
+    deficiencies.extend(iter);
+    deficiencies
+  }
+}
+
+#[cfg(test)]
+mod tests {
```
PhilippGackstatter, line 114: Test methods in this module still have the old refutation naming.

identity-iota/src/credential/deficiencies.rs
```diff
@@ -0,0 +1,188 @@
...
+
+impl Display for CredentialDeficiency {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    write!(f, "{}", self.description())
+  }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+/// A set containing zero or more variants of [`CredentialDeficiency`]
+pub struct CredentialDeficiencySet {
```
PhilippGackstatter, line 44: Discussion: I'm not sure about reimplementing collections. I guess this is a bit more ergonomic than taking `&[CredentialDeficiency]` everywhere? But that seems fine to me, and wouldn't require implementing collection-specific things. Perhaps I'm missing some usecase? Not a strong opinion overall.

PhilippGackstatter (changes requested):
Looks nice! Mostly bikeshedding below. There are quite a few occurrences of the old "refute" naming that should be updated. Also missing tests for the deficiency validation, which is the only thing that keeps me from approving.

identity-iota/src/credential/validated_data.rs
olivereanderson: It provides more type safety as it is now impossible to use them in the same context and this is also more backward compatible if we want to say make `Deactivated`  unrepresentable, as the type itself although public is not reachable. I actually started using const generics here, so it was `DocumentValidation<true>` and `DocumentValidation<false>` but it is less flexible in terms of backwards compatibility so I dropped that idea.

PhilippGackstatter: Fair enough, more type safety is nice to have in this case.

identity-iota/src/credential/validator.rs
olivereanderson: I think we do since the caller can only specify the kind of deficiency, but not be more precise than that. Maybe it would be better to use some other collection for recording the encountered allowed `Deficiencies` that also keeps track of the specifics however.

identity-iota/src/credential/deficiencies.rs
olivereanderson: Good points. You have made me realise that `CredentialDeficiencySet` needs rethinking.

identity-iota/src/credential/validator.rs
PhilippGackstatter: Then again, as you pointed out, the caller can look at the presentation/credential themselves to see, e.g. whether they accept an expiration one day ago.

olivereanderson:
Closed in favour of: https://github.com/iotaledger/identity.rs/pull/599

Issue #304, opened 2021-07-09
[Task] Implement the Account high-level API in WASM


JelleMillenaar:
## Description

Implement the high-level API of the account into WASM. This requires WASM bindings to be implemented for the account API and examples to be added to the node and web. This binding may completely exclude the Stronghold integration in Rust as that will be a separate topic / issue. 

## Motivation

Improve the WASM developer experience

## Resources
Link to any resources relevant for the task such as Issues, PRs, reference implementations, or specifications.

- Depends on #291 

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [X] Implement WASM bindings for the account https://github.com/iotaledger/identity.rs/pull/681
- [X] Write matching examples to the Rust account examples for both node.js and web environments. https://github.com/iotaledger/identity.rs/pull/681
- [ ] Configure a project for both the browser and Node.js with proper TypeScript configuration and linting (make sure code completion works, source map ..etc). 
- [X] Add nightly example testing to CI. (Examples will now run in our regular `build-and-test` job) https://github.com/iotaledger/identity.rs/pull/681
- [ ] Migrate the needed low level examples to the new project.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

eike-hass:
I can take care of running the examples in CI.

PR #574, opened 2022-01-10
WASM Bindings for the Account

abdulmth:
# Description of change
Adds WASM bindings for the account with a TypeScript minimal project and examples.

#304 

## Type of change

- [ ] Documentation Fix

Locally run the Typescript examples.


identity/Cargo.toml
```diff
@@ -21,14 +21,16 @@ identity-iota = { version = "=0.5.0-dev.3", path = "../identity-iota", default-f
 
 [dev-dependencies]
 criterion = { version = "0.3" }
-tokio = { version = "1.5", features = ["full"] }
+tokio = { version = "1.14", features = ["full"] }
```
abdulmth, line 24: This might need correction.

abdulmth:
Please add documentation suggestions where you see fit.

identity/Cargo.toml
cycraig: It's a dev-dependency so it doesn't really matter. `identity-account` still uses `1.5` but ideally all the projects should migrate to the latest version of `tokio 1.15`.

identity-iota/src/tangle/client.rs
```diff
@@ -30,7 +30,7 @@ use crate::tangle::TangleRef;
 use crate::tangle::TangleResolve;
 
 /// Client for performing IOTA Identity operations on the Tangle.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
HenriqueNogara, line 33: It will break otherwise
```
/// Client for performing IOTA Identity operations on the Tangle.
#[cfg_attr(feature = "wasm", derive(Debug, Clone))]
#[cfg_attr(not(feature = "wasm"), derive(Debug))]
pub struct Client {
  pub(crate) client: IotaClient,
  pub(crate) network: Network,
  pub(crate) encoding: DIDMessageEncoding,
}
```

identity/Cargo.toml
```diff
@@ -37,7 +39,7 @@ async = ["identity-iota/async"]
 wasm = ["identity-iota/wasm", "identity-comm/wasm", "identity-core/wasm"]
 
 # Enables support for secure storage of DID Documents
-account = ["identity-account"]
+account = ["identity-account", "wasm"]
```
HenriqueNogara, line 42: "wasm" should not be added here.

identity/Cargo.toml
```diff
@@ -21,14 +21,16 @@ identity-iota = { version = "=0.5.0-dev.3", path = "../identity-iota", default-f
...
-tokio = { version = "1.5", features = ["full"] }
+tokio = { version = "1.14", features = ["full"] }
 
 [[bench]]
 name = "benchmark"
 harness = false
 
 [features]
-default = ["async"]
+default = ["async", "stronghold"]
```
HenriqueNogara, line 31: I don't think we need to add the stronghold as a default feature

identity/src/lib.rs
```diff
@@ -90,7 +90,6 @@ pub mod account {
   pub use identity_account::error::*;
   pub use identity_account::identity::*;
   pub use identity_account::storage::*;
-  pub use identity_account::stronghold::*;
```
HenriqueNogara, line 93:
```
pub use identity_account::stronghold::*;
 ```

File: identity/Cargo.toml [hunk 6]
```diff
@@ -37,7 +39,7 @@ async = ["identity-iota/async"]
 wasm = ["identity-iota/wasm", "identity-comm/wasm", "identity-core/wasm"]
```
HenriqueNogara, line 39:
```suggestion
wasm = ["identity-iota/wasm", "identity-comm/wasm", "identity-core/wasm", "identity-account/wasm"]
```

bindings/wasm/src/account/update/create_service.rs
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    let endpoint = Url::parse(endpoint.as_str()).wasm_result()?;
+
+    let update = Update::CreateService {
+      fragment,
+      type_: service_type,
+      endpoint: ServiceEndpoint::from(endpoint),
+      properties: None, //ToDo
+    };
```
cycraig, line 47: TODO properties. What should we actually pass in for `properties`, just `JsValue`/`any`? That's basically the equivalent...

bindings/wasm/src/account/update/create_method.rs
```diff
@@ -0,0 +1,96 @@
...
+        scope: method_scope,
+      };
+
+      account
+        .as_ref()
+        .borrow_mut()
+        .process_update(update)
+        .await
+        .wasm_result()
+        .and_then(|output| JsValue::from_serde(&output).wasm_result())
```
cycraig, line 63: Can we use `JsValue::undefined()` here instead?

bindings/wasm/src/account/update/delete_method.rs
```diff
@@ -0,0 +1,35 @@
...
+    let promise: Promise = future_to_promise(async move {
+      let update = Update::DeleteMethod { fragment };
+
+      account
+        .as_ref()
+        .borrow_mut()
+        .process_update(update)
+        .await
+        .wasm_result()
+        .and_then(|output| JsValue::from_serde(&output).wasm_result())
```
cycraig, line 30: Can we use `JsValue::undefined()` here instead?

bindings/wasm/src/account/update/delete_service.rs
```diff
@@ -0,0 +1,35 @@
...
+    let update = Update::DeleteService { fragment };
+
+    let promise: Promise = future_to_promise(async move {
+      account
+        .as_ref()
+        .borrow_mut()
+        .process_update(update)
+        .await
+        .wasm_result()
+        .and_then(|output| JsValue::from_serde(&output).wasm_result())
```
cycraig, line 30: Can we use `JsValue::undefined()` here instead?

bindings/wasm/src/account/update/create_service.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let promise: Promise = future_to_promise(async move {
+      account
+        .as_ref()
+        .borrow_mut()
+        .process_update(update)
+        .await
+        .wasm_result()
+        .and_then(|output| JsValue::from_serde(&output).wasm_result())
+    });
```
cycraig, line 57: Can we use `JsValue::undefined()` here instead?

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+        .wasm_result()
+    });
+    promise.unchecked_into::<PromiseResolvedDocument>()
+  }
+
+  /// Removes the identity from the local storage entirely.
+  ///
+  /// Note: This will remove all associated document updates and key material - recovery is NOT POSSIBLE!
+  #[wasm_bindgen(js_name = deleteIdentity)]
+  pub fn delete_identity(self) -> Promise {
```
cycraig, line 94: Another candidate for `PromiseVoid` (along with most methods to be fair).

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+  #[wasm_bindgen(js_name = createSignedCredential)]
+  pub fn create_signed_credential(
+    &self,
+    fragment: String,
+    credential: &WasmCredential,
+    signature_options: WasmSignatureOptions,
+  ) -> PromiseCredential {
+    let account = self.0.clone();
+    let mut cred: Credential = credential.0.clone();
+    let options: SignatureOptions = SignatureOptions::from(signature_options);
```
cycraig, line 169: Opinion: I think it might be better to take a reference to `WasmSignatureOptions` and clone it to avoid nulling-out its pointer in Javascript.

E.g.
```suggestion
    signature_options: &WasmSignatureOptions,
  ) -> PromiseCredential {
    let account = self.0.clone();
    let mut cred: Credential = credential.0.clone();
    let options: SignatureOptions = signature_options.0.clone();
```

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+        .publish_with_options(options)
+        .await
+        .map(|_| JsValue::undefined())
+        .wasm_result()
+    })
+  }
+
+  /// Signs a {@link Credential} with the key specified by `fragment`.
+  #[wasm_bindgen(js_name = createSignedCredential)]
+  pub fn create_signed_credential(
```
cycraig, line 161: Opinion: I still prefer `sign_credential` for parity, even with the differences...

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+  #[wasm_bindgen(js_name = createSignedDocument)]
+  pub fn create_signed_document(
+    &self,
+    fragment: String,
+    document: &WasmDocument,
+    signature_options: WasmSignatureOptions,
+  ) -> PromiseDocument {
+    let account = self.0.clone();
+    let mut doc: IotaDocument = document.0.clone();
+    let options: SignatureOptions = SignatureOptions::from(signature_options);
```
cycraig, line 194: Same as above, consider taking a reference to `WasmSignatureOptions`.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+  #[wasm_bindgen(js_name = createSignedPresentation)]
+  pub fn create_signed_presentation(
+    &self,
+    fragment: String,
+    presentation: &WasmPresentation,
+    signature_options: WasmSignatureOptions,
+  ) -> PromisePresentation {
+    let account = self.0.clone();
+    let mut pres: Presentation = presentation.0.clone();
+    let options: SignatureOptions = SignatureOptions::from(signature_options);
```
cycraig, line 219: Same as above, consider taking a reference to `WasmSignatureOptions`.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+  #[wasm_bindgen(js_name = createSignedData)]
+  pub fn create_signed_data(
+    &self,
+    fragment: String,
+    data: &JsValue,
+    signature_options: WasmSignatureOptions,
+  ) -> Result<Promise> {
+    let account = self.0.clone();
+    let mut verifiable_properties: VerifiableProperties = data.into_serde().wasm_result()?;
+    let options: SignatureOptions = SignatureOptions::from(signature_options);
```
cycraig, line 247: Same as above, consider taking a reference to `WasmSignatureOptions`.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+        .map(|_| JsValue::undefined())
+        .wasm_result()?;
+      JsValue::from_serde(&pres).wasm_result()
+    });
+    promise.unchecked_into::<PromisePresentation>()
+  }
+
+  /// Signs arbitrary `data` with the key specified by `fragment`.
+  #[wasm_bindgen(js_name = createSignedData)]
+  pub fn create_signed_data(
```
cycraig, line 239: This should be called `signData` because it has the same behaviour as `WasmDocument::signData`, developers will not expect a difference.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+  ///
+  /// # WARNING
+  ///
+  /// This method is dangerous and can easily corrupt the internal state,
+  /// potentially making the identity unusable. Only call this if you fully
+  /// understand the implications!
+  #[wasm_bindgen(js_name = updateDocumentUnchecked)]
+  pub fn update_document_unchecked(&mut self, document: WasmDocument) -> Promise {
+    let account = self.0.clone();
+    future_to_promise(async move {
```
cycraig, line 272: I think this should take a reference to document and clone it to prevent unexpectedly nulling its pointer in Javascript.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+extern "C" {
+  #[wasm_bindgen(typescript_type = "Promise<Presentation>")]
+  pub type PromisePresentation;
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "Promise<Document>")]
+  pub type PromiseDocument;
+}
```
cycraig, line 324: Combine these `extern "C"` sections? Otherwise I wonder if these should all be moved to their respective Wasm types for code organisation.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "PublishOptions")]
+  pub type WasmPublishOptions;
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn forceIntegrationUpdate(this: &WasmPublishOptions) -> Option<bool>;
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn signWith(this: &WasmPublishOptions) -> Option<String>;
```
cycraig, line 335: Define these with snake_case and use `js_name` to rename them with camelCase.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+#[wasm_bindgen(typescript_custom_section)]
+const TS_APPEND_CONTENT_2: &'static str = r#"
+/**
```
cycraig, line 340:
```suggestion
#[wasm_bindgen(typescript_custom_section)]
const TS_PUBLISH_OPTIONS: &'static str = r#"
/**
```

bindings/wasm/src/credential/credential.rs
```diff
@@ -18,7 +18,7 @@ use crate::error::Result;
 use crate::error::WasmResult;
 
 #[wasm_bindgen(js_name = Credential, inspectable)]
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct WasmCredential(pub(crate) Credential);
```
cycraig, line 22: Why does this derive `Serialize` and `Deserialize` (and `PartialEq` for that matter)? We always (de)serialize the inner rust type, not the Wasm wrapper.

examples/Cargo.toml
```diff
@@ -8,7 +8,7 @@ publish = false
 identity = { path = "../identity", features = ["account"] }
 pretty_env_logger = { version = "0.4" }
 rand = { version = "0.8" }
-tokio = { version = "1.10", features = ["full"] }
+tokio = { version = "1.14", features = ["full"] }
```
cycraig, line 11: Should be `1.15` now. I think all `tokio` dependencies should be updated to use `1.15` in my opinion.

identity-account/src/account/account.rs
```diff
@@ -134,6 +134,10 @@ impl Account {
     self.storage.as_ref()
   }
 
+  pub fn storage_arc(&self) -> Arc<dyn Storage> {
+    Arc::clone(&self.storage)
+  }
+
```
cycraig, line 140: Just call this `storage`. Is this completely necessary by the way?

identity-account/src/updates/update.rs
```diff
@@ -102,7 +102,7 @@ pub(crate) async fn create_identity(
 }
 
 #[derive(Clone, Debug)]
-pub(crate) enum Update {
+pub enum Update {
   CreateMethod {
     scope: MethodScope,
     type_: MethodType,
```
cycraig, line 108: Do we really need this public? This significantly affects our Rust users in my opinion since it introduces another method of updating credentials apart from the builder.

Can the Wasm bindings not just call the `update_identity()` builders internally without exposing this?

identity-iota/src/tangle/client.rs
cycraig, line 33: `Clone` should not be on `Client`, either find a way to not do this or restrict it to the Wasm target only.

cycraig (changes requested):
Preliminary comments.

The biggest change I think would be to not make `Update` and `process_update` public from the `Account` since that changes the interface on the Rust side. Try use the `update_identity()` builders internally in the Wasm bindings to avoid this.

bindings/wasm/Cargo.toml
```diff
@@ -18,16 +18,20 @@ crate-type = ["cdylib", "rlib"]
 console_error_panic_hook = { version = "0.1" }
 futures = { version = "0.3" }
 js-sys = { version = "0.3" }
+log = "0.4.6"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", default-features = false }
+serde_repr = "0.1"
 wasm-bindgen = { version = "0.2", features = ["serde-serialize"] }
 wasm-bindgen-futures = { version = "0.4", default-features = false }
+wasm-logger = "0.2.0"
```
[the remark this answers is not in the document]
cycraig: Also, disable default features for `serde_repr`, even if it doesn't have any right now.

```toml
serde_repr = { version = "0.1", default-features = false }
```

bindings/wasm/examples-account/README.md
```diff
@@ -0,0 +1,37 @@
+![banner](./../../../.meta/identity_banner.png)
+
+
+## IOTA Identity Account Examples
+
+This folder provides code examples for you to learn how the IOTA Identity WASM bindings for the Account can be used in JavaScript/Typescript.
+
+The examples are written in Typescript and can be run independently with Node.js.
```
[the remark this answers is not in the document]
abdulmth: The idea is to make a clean, minimal project without the need for webpack configuration to avoid complexity for people just starting to learn about the framework (and the account). Also, this allows auto-completion without the need for extra configuration (on VS Code for example). This format is open for change in a future issue, but it is outside the scope of this PR.

bindings/wasm/examples-account/README.md
```diff
@@ -0,0 +1,37 @@
...
+| # | Name | Details |
+| -------- | -------- | -------- |
+| 1 |[create_did](src/create_did.ts)|Generates and publishes a DID Document, the fundamental building block for decentralized identity.|
+|2| [manipulate_did](src/manipulate_did.ts)|  Adds verification methods and service endpoints to a DID Document and updates an already existing DID Document.|           
+|3| [lazy](src/lazy.ts)| Manipulates a DID Document and publishes multiple changes to the tangle at once.|
+|4| [signing](src/signing.ts) | Creates a DID with a verification method, signs a verifiable credential, verifies the signature.|
+|5| [config](src/config.ts) | Applies different account configurations|
+|6| [unchecked](src/unchecked.ts) | Updates an Identity without validation and publishes it to the tangle.|
+## Browser
+Although the examples should work in browser environment, we don't provide a browser project as for now.
```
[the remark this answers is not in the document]
abdulmth: It's not done, maybe in a future issue depending on if we decide to move them with the low level examples. The code should also be valid JavaScript code. (it also can be transpiled to JavaScript if we decide to add TypeScript code, this shouldn't be complicated). In my opinion, using TypeScript by default is really important because we rely heavily on types. This also could encourage developers to support it if they're new to web development.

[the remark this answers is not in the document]
abdulmth: Yes, it requires `load_identity`, I created an empty example, so we don't forget about it.

bindings/wasm/examples-account/src/config.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import {AccountBuilder, Client, Network, ExplorerUrl, Config, DIDMessageEncoding, AutoSave } from './../../node/identity_wasm.js';
```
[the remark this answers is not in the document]
abdulmth: It's possible, Although it defeats the purpose of making the examples very simple, since this PR if focused on the bindings. But the examples are likely to move in a future PR.

bindings/wasm/examples-account/src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+    // The creation step generates a keypair, builds an identity
+    // and publishes it to the IOTA mainnet.
+    let builder = new AccountBuilder({
+        // never auto-save. rely on the drop save.
+        // use `AutoSave.every()` to save immediately after every action,
+        // and `AutoSave.batch(10)` to save after every 10 actions.
+        autoSave: AutoSave.never(),
+        autopublish: true, // publish to the tangle automatically on every update
+        milestone: 4, // save a snapshot every 4 actions
+        client: client // set client to the previously defined client.
```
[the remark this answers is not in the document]
abdulmth: Needs storage bindings first.

bindings/wasm/examples-account/src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+        let did = account.did();
+
+        // Prints the Identity Resolver Explorer URL.
+        // The entire history can be observed on this page by clicking "Loading History".
+        console.log(`[Example] Explore the DID Document = ${explorer.resolverUrl(did.toString())}`);
+
+    } catch (e: any) {
+        console.log(`[Example] Error: ${e.message}`);
+        console.log(`[Example] Is your Tangle node listening on ${private_node_url}?`);
+    }
```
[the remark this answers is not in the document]
abdulmth: you can't catch a specific error type in TypeScript, see [here](https://github.com/Microsoft/TypeScript/issues/8677#issuecomment-220385124). I changed it a bit to make sure it's an instance of `Error` before trying to get `.message`.

bindings/wasm/examples-account/src/lazy.ts
```diff
@@ -0,0 +1,54 @@
...
+        serviceType: "LinkedDomains",
+        endpoint: "https://example.org"
+    });
+
+    // Delete the previously added service.
+    await account.deleteService("example-service");
+
+    // Publish the updates as one message to the tangle.
+    // Also see: account.publishWithOptions(..)
+    await account.publish();
```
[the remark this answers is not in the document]
abdulmth: I think it's good to give a hint about the existence of the other method, since there's no example for it.

bindings/wasm/examples-account/src/manipulate_did.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    // Add a new service to the identity.
+    await account.createService({
+        fragment: "my-service-1",
+        serviceType: "MyCustomService",
+        endpoint: "https://example.com"
+    })
+
+    // Remove the Ed25519 verification method
+    await account.deleteMethod("my-next-key")
```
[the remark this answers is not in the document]
abdulmth: agree

bindings/wasm/examples-account/src/unchecked.ts
```diff
@@ -0,0 +1,47 @@
...
+    // Get a copy of the document this account manages.
+    // We will apply updates to the document, and overwrite the account's current document.
+    let document = account.document();
+
+    // Print the local state of the DID Document
+    console.log(`Document before update`, document);
+
+    // Override the updated field timestamp to 01.01.1990 00:00:00.
+    // because we can. This is usually set automatically by Account::update_identity.
+    document.metadataUpdated = Timestamp.parse("1900-01-01T00:00:00Z")
```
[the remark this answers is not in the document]
abdulmth: I will create an issue later for this.

bindings/wasm/package.json
```diff
@@ -23,6 +23,7 @@
     "serve": "webpack serve",
     "example:node": "node examples/dist/node.js",
     "example:browser": "http-server ./examples/dist/ -c-1 -o ",
+    "example:account": "ts-node ./examples-account/src/node.ts",
```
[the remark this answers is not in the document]
abdulmth: It's not installed by default, thanks for the catch.

bindings/wasm/src/account/account_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+      let auto_save: AutoSave = match o.autoSave() {
+        Some(save) => save.0,
+        None => default_config.autosave,
+      };
+
+      builder = builder
+        .autopublish(o.autopublish().unwrap_or(default_config.autopublish))
+        .milestone(o.milestone().unwrap_or(default_config.milestone))
+        .autosave(auto_save);
+      //todo storage
```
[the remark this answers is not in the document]
abdulmth: Yes

bindings/wasm/src/account/account_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+    //   builder
+    //     .as_ref()
+    //     .borrow_mut()
+    //     .load_identity(did.0)
+    //     .await
+    //     .map(WasmAccount::from)
+    //     .map(Into::into)
+    //     .wasm_result()
+    // });
+    // Ok(promise.unchecked_into::<PromiseAccount>())
```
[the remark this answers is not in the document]
abdulmth: Depends on #597 .

bindings/wasm/src/account/account_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+     * Client for tangle requests.
+     */
+    client?: Client
+};
+"#;
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "IdentitySetup")]
+  pub type WasmIdentitySetup;
```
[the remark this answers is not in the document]
abdulmth: I will create an issue for that later.

bindings/wasm/examples-account/README.md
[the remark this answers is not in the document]
cycraig: If it's outside the scope of this PR then this PR should follow the existing conventions and match the existing Webpack examples.

bindings/wasm/examples-account/README.md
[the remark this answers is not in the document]
cycraig: Except we can't test the Account Typescript examples in the browser now, leaving us open to them simply not working in that environment. We have had numerous instances of the Wasm bindings working in Node.js just fine and breaking in the browser or vice-versa.

bindings/wasm/examples-account/src/config.ts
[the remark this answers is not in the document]
cycraig: Then add it as a commented-out TODO in this example.

bindings/wasm/examples-account/src/unchecked.ts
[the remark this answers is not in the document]
cycraig: Issue #622 already notes that. I think the example should still be updated with the suggested workaround until a better alternative is added, if ever.

bindings/wasm/examples-account/src/lazy.ts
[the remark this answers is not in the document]
cycraig: It's not mentioned in Rust and I think `publish` and `publishWithOptions` should be combined in the Wasm bindings since similar `*with_options` methods have been combined in other places already since it's easier to omit parameters in Javascript.

bindings/wasm/src/account/method_secret.rs
```diff
@@ -0,0 +1,29 @@
...
+use identity::core::decode_b58;
+use identity::crypto::PrivateKey;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(js_name = MethodSecret)]
+pub struct WasmMethodSecret(pub(crate) MethodSecret);
+
+#[wasm_bindgen(js_class = MethodSecret)]
+impl WasmMethodSecret {
+  /// Creates a {@link MethodSecret} object from base58-encoded Ed25519 private key.
```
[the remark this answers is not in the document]
abdulmth: `@link` makes XYZ clickable for easier navigation.

bindings/wasm/src/account/method_secret.rs
```diff
@@ -0,0 +1,29 @@
...
+  #[wasm_bindgen]
+  pub fn ed25519_base58(private_key: &str) -> Result<WasmMethodSecret> {
```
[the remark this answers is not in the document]
abdulmth: I made it camalCase. I also prefer enum variants to be capitalized, but it's a method here, it can be confusing for a method name to start with a capital letter.

bindings/wasm/src/account/update/attach_method_relationship.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::error::WasmResult;
+
+#[wasm_bindgen(js_class = Account)]
+impl WasmAccount {
+  /// Attaches a single or multiple relationships to the given method, if the method exists.
+  ///
+  /// Note: The method needs to be in the set of verification methods,
+  /// so it cannot be an embedded one.
+  #[wasm_bindgen(js_name = attachMethodRelationships)]
+  pub fn attach_relationships(&mut self, input: &AttachMethodRelationshipOptions) -> Result<Promise> {
```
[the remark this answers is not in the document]
abdulmth: There's a difference here since in TS, it's possible to set an array of relationships, while this is done with a separate method call with the builder pattern in Rust. So making it plural in TS makes more sense in my opinion. Open to change it if people think otherwise.

bindings/wasm/src/account/update/attach_method_relationship.rs
```diff
@@ -0,0 +1,102 @@
...
+    let relationships: Vec<WasmMethodRelationship> = match input.relationships().into_serde().wasm_result()? {
+      One(r) => vec![r],
+      Many(r) => r,
+    };
```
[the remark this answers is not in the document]
abdulmth: It works, changed it. I'm not sure if there's any benefit over the explicit deserialization.

bindings/wasm/src/account/update/attach_method_relationship.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let relationships: Vec<MethodRelationship> = relationships.into_iter().map(Into::into).collect();
+
+    let account = self.0.clone();
+    let fragment = match input.fragment() {
+      Some(value) => value,
+      None => return Err(wasm_error(MissingRequiredField("fragment"))),
+    };
```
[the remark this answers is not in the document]
abdulmth: I added a check for when the resulted `Vec` is empty, I'm not sure how to serialize to `Vec` without using `OneOrMany`. This is the reason we used it in the first place. But now, if types are used correctly, it's hard to mess things up from the TS side.

bindings/wasm/src/account/update/attach_method_relationship.rs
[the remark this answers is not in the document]
cycraig: The point is to not deserialize to a `OneOrMany` then convert that to a `Vec` manually. The case where we should throw an error if the input is empty needs to be handled manually either way, as per another comment somewhere below this.

bindings/wasm/src/account/update/attach_method_relationship.rs
```diff
@@ -0,0 +1,102 @@
...
+#[wasm_bindgen (js_name = MethodRelationship)]
+#[derive(Clone, Copy, Serialize_repr, Deserialize_repr, PartialEq, Debug)]
+#[repr(u8)]
+pub enum WasmMethodRelationship {
+  Authentication = 0,
+  AssertionMethod = 1,
+  KeyAgreement = 2,
+  CapabilityDelegation = 3,
+  CapabilityInvocation = 4,
+}
```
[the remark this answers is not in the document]
abdulmth: correct

bindings/wasm/src/account/update/attach_method_relationship.rs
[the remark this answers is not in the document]
cycraig: No, you're right. The relationships need to be deserialized to `OneOrMany` first. My bad.

The empty check is still necessary in case someone passes an empty `Vec`, e.g. `[]`, which currently deserializes to `Many` I believe.

bindings/wasm/src/account/update/create_method.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::did::WasmMethodType;
+use crate::error::wasm_error;
+use crate::error::Result;
+use crate::error::WasmResult;
+
+#[wasm_bindgen(js_class = Account)]
+impl WasmAccount {
+  /// Adds a new verification method to the DID document.
+  #[wasm_bindgen(js_name = createMethod)]
+  pub fn create_method(&mut self, options: &CreateMethodOptions) -> Result<Promise> {
```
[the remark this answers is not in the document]
abdulmth: Well, `any` can be `undefined`. But it's better if we can return undefined. I will create an issue later for this, since it occurs in other bindings too.

bindings/wasm/src/account/update/create_method.rs
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "CreateMethodOptions")]
+  pub type CreateMethodOptions;
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn fragment(this: &CreateMethodOptions) -> Option<String>;
```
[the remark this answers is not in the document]
abdulmth: in TS you will get a compile time error since `fragment` is not optional. If you do it in JS it panics with 

`Err > TypeError: Cannot read properties of undefined (reading 'length')
    at passStringToWasm0 (C:\iota\identity2\identity.rs\bindings\wasm\node\identity_wasm.js:85:19)
    at module.exports.__wbg_fragment_84b451f64fea39cc (C:\iota\identity2\identity.rs\bindings\wasm\node\identity_wasm.js:4173:16)
    at null.<anonymous> (wasm://wasm/00e35596:1:2451167)
    at null.<anonymous> (wasm://wasm/00e35596:1:1460972)
    at null.<anonymous> (wasm://wasm/00e35596:1:2351346)
    at Account.attachMethodRelationships (C:\iota\identity2\identity.rs\bindings\wasm\node\identity_wasm.js:533:28)
    at manipulateIdentity (C:\iota\identity2\identity.rs\bindings\wasm\examples-account\src\manipulate_did.ts:27:19)
    at async main (C:\iota\identity2\identity.rs\bindings\wasm\examples-account\src\node.ts:24:20)`

Which is not very clean error. It's best to make it optional and check if it's `Some` otherwise throw a proper error.

bindings/wasm/src/account/update/create_method.rs
```diff
@@ -0,0 +1,96 @@
...
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn methodScope(this: &CreateMethodOptions) -> Option<WasmMethodScope>;
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn methodType(this: &CreateMethodOptions) -> Option<WasmMethodType>;
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn methodSecret(this: &CreateMethodOptions) -> Option<WasmMethodSecret>;
```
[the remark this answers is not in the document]
abdulmth: I think it's not necessary since it indicates that you're calling functions from JS and Clippy doesn't complain about it.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,365 @@
...
+  #[wasm_bindgen]
+  pub fn autosave(&self) -> WasmAutoSave {
+    WasmAutoSave(self.0.as_ref().borrow().autosave())
+  }
+
+  /// Returns the total number of actions executed by this instance.
+  #[wasm_bindgen]
+  pub fn actions(&self) -> usize {
+    self.0.as_ref().borrow().actions()
+  }
```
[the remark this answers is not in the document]
abdulmth: I think it's needed when you set `autosave` to `batch(n)`. so it's good to know when an auto save will happen.

identity-account/src/updates/update.rs
abdulmth: The advantage here is that all properties must be exhausted in the bindings, which is important not to forget any of them if we use the builder. But you have a good point here. I can change it if we decide that it's not worth it. But I personally don't see a big issue. Plus, I really dislike not having auto completion for the builder (I think it's only possible when using the Rust Analyzer, which I don't use) because it's implemented in a macro.

bindings/wasm/examples-account/src/lazy.ts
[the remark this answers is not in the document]
abdulmth: I merged both methods.

bindings/wasm/examples-account/src/unchecked.ts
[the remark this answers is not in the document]
abdulmth: I added the workaround to the example.

identity/Cargo.toml
abdulmth: `cargo update` updated this dependency to this version. The newest version I guess i `1.16`. Should we updated them by hand?

bindings/wasm/examples-account/README.md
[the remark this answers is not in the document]
abdulmth: This PR will be merged into an epic branch, this issue will be dealt with before merging into dev. See #304.

bindings/wasm/src/account/account_builder.rs
```diff
@@ -0,0 +1,169 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::account::wasm_account::WasmAccount;
```
HenriqueNogara, line 4: Nit: Usually the imports are separated by blocks. One block with imports from the standard library, one from other libs, and one within the same crate. 

When I merge your branched into mine I changed this, so don't worry, you don't have do it.

bindings/wasm/src/account/account_builder.rs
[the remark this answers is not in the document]
HenriqueNogara: I guess we can do this after both this PR and #597 are merge into the epic branch. I added a storage interface so we can change the builder after that.

identity-account/src/account/account.rs
cycraig: Discussed this, it is necessary. I do wonder if `storage` can instead return `&Arc<dyn Storage>` though.

identity-account/src/updates/update.rs
cycraig: Yeah, we've had discussions and I've been thinking more about this and we might want to expose the update structs even in Rust. Will need to be a joint decision though.

bindings/wasm/package.json
```diff
@@ -62,7 +63,8 @@
     "webpack-dev-server": "^4.7.3"
   },
   "dependencies": {
-    "node-fetch": "^2.6.7"
+    "node-fetch": "^2.6.7",
+    "ts-node": "^10.4.0"
```
abdulmth, line 67: Should be a dev dependency.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,360 @@
...
+      future_to_promise(async move {
+        account
+          .as_ref()
+          .borrow_mut()
+          .publish()
+          .await
+          .map(|_| JsValue::undefined())
+          .wasm_result()
+      })
+    }
```
PhilippGackstatter, line 159:
```suggestion
    let mut options: PublishOptions = PublishOptions::default();

    if let Some(publish_options) = publish_options {
      if let Some(force_integration) = publish_options.forceIntegrationUpdate() {
        options = options.force_integration_update(force_integration);
      }

      if let Some(sign_with) = publish_options.signWith() {
        options = options.sign_with(sign_with);
      };
    };

    future_to_promise(async move {
      account
        .as_ref()
        .borrow_mut()
        .publish_with_options(options)
        .await
        .map(|_| JsValue::undefined())
        .wasm_result()
    })
```

That achieves the same :slightly_smiling_face: `Account::publish` is the same as `publish_with_options` with `PublishOptions::default()`.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,360 @@
...
+      account
+        .as_ref()
+        .borrow_mut()
+        .sign(fragment.as_str(), &mut cred, options)
+        .await
+        .map(|_| JsValue::undefined())
+        .wasm_result()?;
+      JsValue::from_serde(&cred).wasm_result()
+    });
+    promise.unchecked_into::<PromiseCredential>()
```
PhilippGackstatter, line 184: You could reduce duplication in these `create_signed_*` methods, as an example:

```rust
  /// Signs a {@link Document} with the key specified by `fragment`.
  #[wasm_bindgen(js_name = createSignedDocument)]
  pub fn create_signed_document(
    &self,
    fragment: String,
    document: &WasmDocument,
    signature_options: &WasmSignatureOptions,
  ) -> PromiseDocument {
    self
      .create_signed(fragment, document.0.clone(), signature_options)
      .unchecked_into::<PromiseDocument>()
  }

  fn create_signed<U>(&self, fragment: String, mut data: U, signature_options: &WasmSignatureOptions) -> Promise
  where
    U: serde::Serialize + SetSignature + 'static,
  {
    let account = self.0.clone();
    let options: SignatureOptions = SignatureOptions::from(signature_options);

    future_to_promise(async move {
      account
        .as_ref()
        .borrow_mut()
        .sign(fragment.as_str(), &mut data, options)
        .await
        .map(|_| JsValue::undefined())
        .wasm_result()?;
      JsValue::from_serde(&data).wasm_result()
    })
  }
```

Unless I'm missing something, you should be able to call `create_signed` from the other methods as well.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,360 @@
...
+        .map(|_| JsValue::undefined())
+        .wasm_result()?;
+      JsValue::from_serde(&verifiable_properties).wasm_result()
+    });
+    Ok(promise)
+  }
+
+  /// Overwrites the {@link Document} this account manages, **without doing any validation**.
+  ///
+  /// # WARNING
```
PhilippGackstatter, line 264:
```suggestion
  /// ### WARNING
```

This looks a bit weird in the API reference as it currently is.

examples/account/config.rs
```diff
@@ -46,8 +46,7 @@ async fn main() -> Result<()> {
       // Configure a client for the private network
       ClientBuilder::new()
         .network(network.clone())
-        .primary_node(private_node_url, None, None)?
-        // .permanode(<permanode_url>, None, None)? // set a permanode for the same network
+        .primary_node(private_node_url, None, None)?, // .permanode(<permanode_url>, None, None)? // set a permanode for the same network
```
PhilippGackstatter, line 49:
```suggestion
        .primary_node(private_node_url, None, None)?,
         // set a permanode for the same network
         // .permanode(<permanode_url>, None, None)?
```

I think it should be this :laughing: Putting everything in one line is very unreadable, imo.

identity-account/src/account/account.rs
PhilippGackstatter: Good suggestion, I think we should try returning `&Arc<dyn Storage>` from `Account::storage` and get rid of `storage_arc`.

identity-account/src/account/account.rs
```diff
@@ -310,7 +314,7 @@ impl Account {
     self.storage().state(self.did()).await?.ok_or(Error::IdentityNotFound)
   }
 
-  pub(crate) async fn process_update(&mut self, update: Update) -> Result<()> {
+  pub async fn process_update(&mut self, update: Update) -> Result<()> {
```
PhilippGackstatter, line 317: I'm hoping we can reverse this. I see this as an internal method and the proper way is through `update_identity`.

identity-account/src/identity/identity_setup.rs
```diff
@@ -4,8 +4,8 @@
 use identity_core::crypto::KeyType;
 
 use crate::types::MethodSecret;
-
 /// Configuration used to create a new Identity.
+/// Overrides the default creation of private and public keys.
```
PhilippGackstatter, line 8:
```suggestion
/// Allows providing a pre-existing private key using `method_secret`.
```

identity-account/src/updates/update.rs
PhilippGackstatter: Pros:
- Changing an `Update` variant more likely produces a compiler error in the Wasm bindings, e.g. if a field changes from optional to required (unlikely, imo).

Cons:
- Exposes a second method of updating the account which pollutes the API unnecessarily on the Rust side.
- Duplicates default logic (very minor, only affects `Update::CreateMethod::type_`, unlikely to change).

We could make `process_update` public behind the `wasm` feature, but I don't really like that option either, just adds to the fragmentation we will have with `Storage`. I think using the builders is better, as a change to the `Update` enum will most likely require changes to Wasm as well, and would likely also produce a compiler error, and so I find it unlikely that some change slips through unnoticed.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,360 @@
...
+/// publishing to the Tangle.
+#[wasm_bindgen(js_name = Account)]
+pub struct WasmAccount(pub(crate) Rc<WasmRefCell<Account>>);
+
+#[wasm_bindgen(js_class = Account)]
+impl WasmAccount {
+  #[wasm_bindgen(js_name = did)]
+  pub fn did(&self) -> WasmDID {
+    let x = self.0.as_ref().borrow();
+    WasmDID::from(x.document().id().clone())
```
PhilippGackstatter, line 48:
```suggestion
    let account = self.0.as_ref().borrow();
    WasmDID::from(account.document().id().clone())
```

PhilippGackstatter (changes requested):
This looks solid overall!

I have some comments regarding exposing `process_update` and other minor things like code duplication. There's still documentation missing in a few place where you wrote custom TS interfaces. Some have docs, some don't, that should be consistent.

bindings/wasm/src/account/update/attach_method_relationship.rs
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "AttachMethodRelationshipOptions")]
+  pub type AttachMethodRelationshipOptions;
+
+  #[wasm_bindgen(structural, getter, method)]
+  pub fn fragment(this: &AttachMethodRelationshipOptions) -> Option<String>;
```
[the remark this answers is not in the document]
cycraig: I think so, to handle the error cases manually where Javascript developers don't pass that field.

bindings/wasm/src/account/update/create_method.rs
```diff
@@ -0,0 +1,92 @@
...
+impl WasmAccount {
+  /// Adds a new verification method to the DID document.
+  #[wasm_bindgen(js_name = createMethod)]
+  pub fn create_method(&mut self, options: &CreateMethodOptions) -> Result<Promise> {
+    let account: Rc<WasmRefCell<Account>> = Rc::clone(&self.0);
+
+    let method_type: MethodType = options
+      .methodType()
+      .map(|m| m.0)
+      .unwrap_or(MethodType::Ed25519VerificationKey2018);
```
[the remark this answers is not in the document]
cycraig: @abdulmth raised the point that we don't get compilation errors in the bindings with builders if the Rust `Account` adds new fields to them, which we would when using the update structs directly.

I'm still uncertain if we should make them public for Rust too but I'm starting to lean that direction, mostly held back by the status quo of leaving them un-exported to reduce the API surface and ensure there is only one "canonical" method of updating `Accounts` to avoid confusion.

identity-account/src/account/account.rs
abdulmth: I already tried that, the problem is that with `attach_relationship` we get a `Vec` from JavaScript (Since we deserialize to `OneOrMany`). And the builder pattern only takes one `MethodRelationship`, If we try to create a loop, we will get an error for variables being dropped (it doesn't happen if you chain them, but we can chain anymore because we need the loop). The solution for that is by setting them in temporary variables, but that got really messy. Plus, creating an instance of `Update` insures that all properties are set.

PhilippGackstatter: This works fine on my side.

```rust
let promise: Promise = future_to_promise(async move {
      let mut account: RefMut<Account> = account.as_ref().borrow_mut();
      let mut updater: IdentityUpdater<'_> = account.update_identity();
      let mut attach_relationship: AttachMethodRelationshipBuilder<'_> = updater.attach_method_relationship();

      attach_relationship = attach_relationship.fragment(fragment);

      for relationship in relationships {
        attach_relationship = attach_relationship.relationship(relationship);
      }

      attach_relationship
        .apply()
        .await
        .wasm_result()
        .map(|_| JsValue::undefined())
    });
```

> Plus, creating an instance of Update insures that all properties are set.

Yes, that is definitely a good point and a desirable thing to have. But changing a type in some `Update::*` variant on the Rust side automatically changes the type in the builder as well, so we'd get a compiler error for that. If we add a property to the Rust builders, then we'd have to change the Wasm bindings anyway to add support for that property, too. These seem to be the two scenarios where it matters, and I think those are caught in the typical review process. I don't see that as enough of a reason to pollute the Rust API with two ways of updating identities.

identity/Cargo.toml
cycraig: I think to be consistent the versions should be updated manually yes. Not sure why `1.14` was needed specifically.

abdulmth:
Thanks @PhilippGackstatter  @cycraig  and @HenriqueNogara for the initial review I implemented most of the suggested changes.
regarding https://github.com/iotaledger/identity.rs/pull/574#discussion_r796623147  I'm not sure how dropping objects in WASM memory happens. Even forcing garbage collection in JS doesn't `drop` the account in the example. And in JS we can't expect the behavior of the garbage collector anyways, so I don't see any other way for the example but calling the `free()` method.

PhilippGackstatter (approved):
Looks good to me. Thanks for all the changes, especially using the `IdentityUpdater`. Much appreciated.

I still think
-  we shouldn't use single-letter variables like `x`
- and that `attach_relationships` and `detach_relationships` should not require the passed list to be non-empty, to ensure the same behaviour in Wasm and Rust

But I don't insist :smile:. 

The `Arc::storage` change to return `&Arc` rather than `Arc` seems important, though.

identity-account/src/account/account.rs
```diff
@@ -129,9 +129,9 @@ impl Account {
   // Getters & Setters
   // ===========================================================================
 
-  /// Returns a reference to the [Storage] implementation.
-  pub fn storage(&self) -> &dyn Storage {
-    self.storage.as_ref()
+  /// Returns a reference counter to the [Storage] implementation.
+  pub fn storage(&self) -> Arc<dyn Storage> {
+    Arc::clone(&self.storage)
```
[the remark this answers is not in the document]
abdulmth: Thanks @PhilippGackstatter   for the review, I fixed the remaining issues too.

bindings/wasm/src/account/update/detach_method_relationship.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::account::wasm_account::WasmAccount;
+use crate::account::wasm_method_relationship::WasmMethodRelationship;
+use crate::error::Result;
+use crate::error::WasmResult;
+
+#[wasm_bindgen(js_class = Account)]
+impl WasmAccount {
+  /// Detaches the given relationship from the given method, if the method exists.
+  #[wasm_bindgen(js_name = detachMethodRelationships)]
+  pub fn detach_relationships(&mut self, options: &DetachMethodRelationshipOptions) -> Result<Promise> {
```
cycraig, line 28:
```suggestion
  /// Detaches the given relationship from the given method, if the method exists.
  #[wasm_bindgen(js_name = detachMethodRelationships)]
  pub fn detach_method_relationships(&mut self, options: &DetachMethodRelationshipOptions) -> Result<Promise> {
```

bindings/wasm/src/account/update/detach_method_relationship.rs
```diff
@@ -0,0 +1,95 @@
...
+    let relationships: Vec<MethodRelationship> = options
+      .relationships()
+      .into_serde::<OneOrMany<WasmMethodRelationship>>()
+      .map(OneOrMany::into_vec)
+      .wasm_result()?
+      .into_iter()
+      .map(Into::into)
+      .collect();
```
cycraig, line 36: Nit: I prefer explicit `from` calls but totally fine either way.

```suggestion
    let relationships: Vec<MethodRelationship> = options
      .relationships()
      .into_serde::<OneOrMany<WasmMethodRelationship>>()
      .map(OneOrMany::into_vec)
      .wasm_result()?
      .into_iter()
      .map(MethodRelationship::from)
      .collect();
```

bindings/wasm/src/account/update/detach_method_relationship.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let account: Rc<WasmRefCell<Account>> = Rc::clone(&self.0);
+    let fragment: String = options
+      .fragment()
+      .ok_or(MissingRequiredField("fragment"))
+      .wasm_result()?;
+
+    let promise: Promise = future_to_promise(async move {
```
cycraig, line 44: Nit: only clone the `Rc` after input handling.

```suggestion
    let fragment: String = options
      .fragment()
      .ok_or(MissingRequiredField("fragment"))
      .wasm_result()?;

    let account: Rc<WasmRefCell<Account>> = Rc::clone(&self.0);
    let promise: Promise = future_to_promise(async move {
```

bindings/wasm/src/account/update/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+mod attach_method_relationship;
+mod create_method;
+mod create_service;
+mod delete_method;
+mod delete_service;
+mod detach_method_relationship;
```
cycraig, line 9: Aside: still slightly annoyed that `attach_method_relationship` and `detach_method_relationship` are inconsistently pluralised.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,320 @@
...
+///
+/// It handles private keys, writing to storage and
+/// publishing to the Tangle.
+#[wasm_bindgen(js_name = Account)]
+pub struct WasmAccount(pub(crate) Rc<WasmRefCell<Account>>);
+
+#[wasm_bindgen(js_class = Account)]
+impl WasmAccount {
+  #[wasm_bindgen(js_name = did)]
+  pub fn did(&self) -> WasmDID {
```
cycraig, line 46: Nit: missing doc-comment.

```suggestion
  /// Returns the {@link DID} of the managed identity.
  #[wasm_bindgen(js_name = did)]
  pub fn did(&self) -> WasmDID {
```

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,320 @@
...
+    let account: Ref<Account> = self.0.as_ref().borrow();
+    WasmDID::from(account.document().id().clone())
```
cycraig, line 48: Nit: could be condensed?

Aside: pretty sure `.as_ref().borrow()` is unnecessary and can just be `.borrow()` unless it's a `WasmRefCell` thing. Doesn't work if you import `std::borrow::Borrow` though. This applies pretty much everywhere so I won't repeat this comment, assuming it works.

```suggestion
    WasmDID::from(account.borrow().did().clone())
```

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,320 @@
...
+
+  /// Push all unpublished changes to the tangle in a single message.
+  #[wasm_bindgen]
+  pub fn publish(&mut self, publish_options: Option<WasmPublishOptions>) -> Promise {
+    let account = self.0.clone();
+    let mut options: PublishOptions = PublishOptions::default();
+    if let Some(publish_options) = publish_options {
+      options = PublishOptions::from(publish_options);
+    };
+    future_to_promise(async move {
```
cycraig, line 125: Nit: `PublishOptions` implements `Default` so this can be condensed somewhat. Should not be mutable either way even with `if-let` syntax.

```suggestion
    let options: PublishOptions = publish_options.map(PublishOptions::from).unwrap_or_default();
    
    let account: Rc<WasmRefCell<Account>> = Rc::clone(&self.0);
    future_to_promise(async move {
```

bindings/wasm/src/account/wasm_account.rs
cycraig: Point still stands I think.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,320 @@
...
+
+
+    /**
+     *
+     *
+     * Set the fragment of a verification method with which to sign the update.
+     * This must point to an Ed25519 method with a capability invocation
+     * verification relationship.
+     */
+     signWith?: string
```
cycraig, line 302:
```suggestion
    /**
     * Set the fragment of a verification method with which to sign the update.
     * This must point to an Ed25519 method with a capability invocation
     * verification relationship.
     *
     *  If omitted, the default signing method on the Document will be used.
     */
     signWith?: string
```

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,320 @@
...
+    let mut options: PublishOptions = PublishOptions::new();
+
+    if let Some(force_integration) = publish_options.forceIntegrationUpdate() {
+      options = options.force_integration_update(force_integration);
+    }
+
+    if let Some(sign_with) = publish_options.signWith() {
+      let sign_with: String = sign_with;
+      options = options.sign_with(sign_with);
+    };
```
cycraig, line 317: Is there some weird coercion happening here, this can be condensed right?

```suggestion
    if let Some(sign_with) = publish_options.signWith() {
      options = options.sign_with(sign_with);
    };
```

bindings/wasm/src/account/wasm_auto_save.rs
```diff
@@ -0,0 +1,30 @@
...
+use identity::account::AutoSave;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(js_name = AutoSave)]
+pub struct WasmAutoSave(pub(crate) AutoSave);
+
+/// Available auto-save behaviours.
+#[wasm_bindgen(js_class = AutoSave)]
+impl WasmAutoSave {
+  /// Never save
```
cycraig, line 13:
```suggestion
  /// Never save.
```

bindings/wasm/src/account/wasm_auto_save.rs
```diff
@@ -0,0 +1,30 @@
...
+  #[wasm_bindgen]
+  pub fn never() -> WasmAutoSave {
+    Self(AutoSave::Never)
+  }
+
+  /// Save after every action
```
cycraig, line 19:
```suggestion
  /// Save after every action.
```

bindings/wasm/src/account/wasm_auto_save.rs
```diff
@@ -0,0 +1,30 @@
...
+  #[wasm_bindgen]
+  pub fn every() -> WasmAutoSave {
+    Self(AutoSave::Every)
+  }
+
+  /// Save after every N actions
```
cycraig, line 25:
```suggestion
  /// Save after every N actions.
```

bindings/wasm/src/account/wasm_method_relationship.rs
```diff
@@ -0,0 +1,30 @@
...
+  Authentication = 0,
+  AssertionMethod = 1,
+  KeyAgreement = 2,
+  CapabilityDelegation = 3,
+  CapabilityInvocation = 4,
+}
+
+impl From<WasmMethodRelationship> for MethodRelationship {
+  fn from(r: WasmMethodRelationship) -> Self {
+    match r {
```
cycraig, line 22:
```suggestion
  fn from(relationship: WasmMethodRelationship) -> Self {
    match relationship {
```

bindings/wasm/src/account/wasm_method_relationship.rs
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use identity::did::MethodRelationship;
+use serde_repr::Deserialize_repr;
+use serde_repr::Serialize_repr;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen (js_name = MethodRelationship)]
+#[derive(Clone, Copy, Serialize_repr, Deserialize_repr, PartialEq, Debug)]
+#[repr(u8)]
```
cycraig, line 11: Do we still need `Clone`, `Copy`, `PartialEq`, `Debug` on this? Are they required trait bounds from `Serialize_repr`, and `Deserialize_repr`? If not can we remove them?

bindings/wasm/src/crypto/wasm_signature_options.rs
```diff
@@ -37,9 +37,9 @@ impl From<SignatureOptions> for WasmSignatureOptions {
...
 }
 
-impl From<WasmSignatureOptions> for SignatureOptions {
-  fn from(options: WasmSignatureOptions) -> Self {
-    options.0
+impl From<&WasmSignatureOptions> for SignatureOptions {
+  fn from(options: &WasmSignatureOptions) -> Self {
+    options.clone().0
   }
 }
```
cycraig, line 44: Bit odd, maybe make two `From` implementations, one for the borrow and one for owned if you really need this?

identity/Cargo.toml
cycraig: Why was this resolved?

`wasm` should not be added here.

identity/Cargo.toml
```diff
@@ -34,14 +34,17 @@ default = ["async"]
 async = ["identity-iota/async"]
 
 # Enables Web Assembly support
-wasm = ["identity-iota/wasm", "identity-comm/wasm", "identity-core/wasm"]
+wasm = ["identity-iota/wasm", "identity-comm/wasm", "identity-core/wasm", "identity-account/wasm"]
 
 # Enables support for secure storage of DID Documents
-account = ["identity-account"]
+account = ["identity-account", "wasm"]
```
cycraig, line 40: I'll repeat the comment from @HenriqueNogara for visibility: `wasm` should not be added here.

cycraig (changes requested):
Thanks for the many changes from the previous reviews!

Looking good overall but still some minor things that definitely need to be addressed.

bindings/wasm/src/account/account_builder.rs
```diff
@@ -0,0 +1,165 @@
...
+/// An [`Account`] builder for easy account configuration.
+///
+/// To reduce memory usage, accounts created from the same builder share the same `Storage`
+/// used to store identities, and the same {@link Client} used to publish identities to the Tangle.
+///
+/// The configuration on the other hand is cloned, and therefore unique for each built account.
+/// This means a builder can be reconfigured in-between account creations, without affecting
+/// the configuration of previously built accounts.
+#[wasm_bindgen(js_name = AccountBuilder)]
+pub struct WasmAccountBuilder(Rc<WasmRefCell<AccountBuilder>>);
```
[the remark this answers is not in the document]
PhilippGackstatter: From the `WasmRefCell` doc:

> This is just a wrapper around all Rust objects passed to JS intended to guard accidental reentrancy, so this vendored version is intended solely to not panic in libstd. Instead when it "panics" it calls our throw function in this crate which raises an error in JS.

It's basically just a copy of `RefCell` minus the panic machinery. I'm not sure, however, if this is just intended for internal usage in wasm-bindgen or if users are supposed to use it. I assume the panic machinery is included in our code anyway, so using one or the other probably doesn't make a difference for binary size.

bindings/wasm/src/account/account_builder.rs
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * `autopublish == true` the account will publish messages to the tangle on each update.
+     * `autopublish == false` the account will combine and publish message when .publish() is called.
+     */
+    autopublish?: boolean,
+
+    /**
+     * Number of actions required to save a snapshot.
+     */
+    milestone?: number,
```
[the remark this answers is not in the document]
PhilippGackstatter: It appears `milestone` is unused in the Rust account, so I agree we can remove it entirely.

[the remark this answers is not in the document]
cycraig: @abdulmth merge with `dev` branch after #645 is merged.

bindings/wasm/examples-account/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+| # | Name | Details |
+| -------- | -------- | -------- |
+| 1 |[create_did](src/create_did.ts)|Generates and publishes a DID Document, the fundamental building block for decentralized identity.|
+|2| [manipulate_did](src/manipulate_did.ts)|  Adds verification methods and service endpoints to a DID Document and updates an already existing DID Document.|
+|3| [lazy](src/lazy.ts)| Manipulates a DID Document and publishes multiple changes to the tangle at once.|
+|4| [signing](src/signing.ts) | Creates a DID with a verification method, signs a verifiable credential, verifies the signature.|
+|5| [config](src/config.ts) | Applies different account configurations|
+|6| [unchecked](src/unchecked.ts) | Updates an Identity without validation and publishes it to the tangle.|
+|7| [multiple_identities](src/multiple_identities.ts) | Managing multiple identities with the Account.|
```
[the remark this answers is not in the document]
abdulmth: I updated the WASM descriptions to match the rust ones, I'm not sure what do you mean by "Please update the Rust ones too".

bindings/wasm/src/account/account_builder.rs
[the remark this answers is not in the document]
abdulmth: I changed it to use `RefCell`.

bindings/wasm/src/account/wasm_account.rs
abdulmth: No, I think I wanted to write something like 
```rust 
if let Some(sign_with: String) = publish_options.signWith()
```
But apparently this doesn't work in Rust. I removed the redundant line.

bindings/wasm/src/account/wasm_method_relationship.rs
abdulmth: I removed them.

abdulmth:
Thanks @cycraig for the second review, I implemented most of the suggestions. There's a couple of issues that we might wanna still do after this PR, `PromiseVoid` is one of them. 
Regarding to `createSignedData(..)` I didn't change it because we voted on explicit naming. We can still discuss it if you think it is needed.

bindings/wasm/src/account/wasm_account.rs
```diff
@@ -0,0 +1,323 @@
...
```
cycraig, line 314: Line was not removed from previous comment.

```suggestion
    if let Some(sign_with) = publish_options.signWith() {
      options = options.sign_with(sign_with);
    };
```

bindings/wasm/src/account/wasm_account_builder.rs
```diff
@@ -0,0 +1,153 @@
...
+  /// by the {@link Client} used to publish it.
+  ///
+  /// @See {@link IdentitySetup} to customize the identity creation.
+  #[wasm_bindgen(js_name = createIdentity)]
+  pub fn create_identity(&mut self, identity_setup: Option<WasmIdentitySetup>) -> Result<PromiseAccount> {
+    // Create IdentitySetup
+    let setup: IdentitySetup = identity_setup.map(IdentitySetup::from).unwrap_or_default();
+
+    // Call the builder.
+    let builder: Rc<RefCell<AccountBuilder>> = self.0.clone();
```
cycraig, line 97: Nit: the comments here are unnecessary in my opinion.

```suggestion
    let setup: IdentitySetup = identity_setup.map(IdentitySetup::from).unwrap_or_default();

    let builder: Rc<RefCell<AccountBuilder>> = self.0.clone();
```

bindings/wasm/src/crypto/wasm_signature_options.rs
```diff
@@ -37,12 +37,6 @@ impl From<SignatureOptions> for WasmSignatureOptions {
...
 
-impl From<WasmSignatureOptions> for SignatureOptions {
-  fn from(options: WasmSignatureOptions) -> Self {
-    options.0
-  }
-}
```
cycraig, line 44: I don't think this implementation should be removed entirely, the previous comment was about overwriting it with an implementation for a borrowed parameter.

We use `From` for clarity ,compared to accessing the inner field directly with `x.0` everywhere, and with functional-style calls like `map`.

identity-account/src/account/config.rs
```diff
@@ -35,23 +35,19 @@ impl AccountSetup {
...
-pub(crate) struct AccountConfig {
-  pub(crate) autosave: AutoSave,
-  pub(crate) autopublish: bool,
-  pub(crate) testmode: bool,
-  pub(crate) milestone: u32,
+pub struct AccountConfig {
+  pub autosave: AutoSave,
+  pub autopublish: bool,
+  pub testmode: bool,
 }
```
cycraig, line 42: Making `AccountConfig` public means any changes to it will be breaking, and has the side-effect of exposing the `testmode` field which should only be used when compiled for tests, never seen by end-developers.

Does this absolutely need to be `public`? If it does, maybe conditionally compile `testmode`? (Unsure if this code works)

```suggestion
pub struct AccountConfig {
  pub autosave: AutoSave,
  pub autopublish: bool,
  #[cfg(test)]
  pub testmode: bool,
}
```

cycraig (changes requested):
Thanks for the new changes.

> There's a couple of issues that we might wanna still do after this PR, PromiseVoid is one of them.

Indeed, we really need to avoid returning `Promise<any>` if we can avoid it, particularly since it's misleading when the functions don't actually return anything at all.

> Regarding to createSignedData(..) I didn't change it because we voted on explicit naming. We can still discuss it if you think it is needed.

The naming is a divergence from the Rust library which I would like to keep to a minimum, but I understand why it was done.

Minor changes left.

bindings/wasm/examples-account/src/config.ts
```diff
@@ -0,0 +1,65 @@
...
+
+
+    // The creation step generates a keypair, builds an identity
+    // and publishes it to the IOTA mainnet.
+    let builder = new AccountBuilder({
+        // `AutoSave.never()` never auto-saves, relies on the storage drop save.
+        // `AutoSave.every()` saves immediately after every action,
+        autosave: AutoSave.batch(10), // saves after every 10 actions.
+        autopublish: true, // publish to the tangle automatically on every update
+        milestone: 4, // save a snapshot every 4 actions
```
cycraig, line 42: Sorry, missed this in previous review. Remove `milestone`.

```suggestion
```

identity-account/src/account/config.rs
abdulmth: I changed the implementation to use the Builder instead of `AccountConfig`.

bindings/wasm/src/crypto/wasm_signature_options.rs
abdulmth: I added the implementation for the borrowed parameter.

bindings/wasm/src/crypto/wasm_signature_options.rs
```diff
@@ -37,6 +37,12 @@ impl From<SignatureOptions> for WasmSignatureOptions {
   }
 }
 
+impl From<&WasmSignatureOptions> for SignatureOptions {
+  fn from(options: &WasmSignatureOptions) -> Self {
+    options.0.clone()
+  }
+}
+
```
cycraig, line 45: No, I meant leave the original implementation of the owned type: `From<WasmSignatureOptions> for SignatureOptions`.

bindings/wasm/README.md
```diff
@@ -3,7 +3,9 @@
 > This is the beta version of the official WASM bindings for [IOTA Identity](https://github.com/iotaledger/identity.rs).
 
 ## [API Reference](https://wiki.iota.org/identity.rs/libraries/wasm/api_reference)
-## [Examples](https://github.com/iotaledger/identity.rs/blob/main/bindings/wasm/examples/README.md)
+
+## [Account Examples](https://github.com/iotaledger/identity.rs/blob/main/bindings/wasm/examples-account/README.md)
+## [Low-Level Examples](https://github.com/iotaledger/identity.rs/blob/main/bindings/wasm/examples/README.md)
```
cycraig, line 8: Right, I guess we have to use absolute links since I assume the README.md is included in the wiki directly.

No changes requested, just annoying since the link won't work until 0.5 is released to main.

HenriqueNogara (approved):

PR #580, opened 2022-01-11
Use own docker image for changelog generator

eike-hass:
# Description of change
Use our own image with the latest revision from https://github.com/github-changelog-generator/github-changelog-generator to get necessary bugfixes.

## Type of change

- [ ] Documentation Fix

The change has already been on the support branch https://github.com/iotaledger/identity.rs/actions/runs/1683399961

## Change checklist

### DIFF
--- a/.github/actions/release/changelog-generator/action.yml
+++ b/.github/actions/release/changelog-generator/action.yml
@@ -34,7 +34,7 @@ runs:
         echo GITHUB_REPOSITORY_PROJECT=$GITHUB_REPOSITORY_PROJECT >> $GITHUB_ENV
 
     - name: Run github-changelog-generator
-      uses: docker://githubchangeloggenerator/github-changelog-generator:1.16.2
+      uses: docker://eiha/github-changelog-generator:latest
       with:
         args: >
           --output ${{ inputs.changelog-path }}


cycraig:
Given that the [Dockerfile commands](https://github.com/eike-hass/docker-github-changelog-generator/blob/master/Dockerfile) are relatively simple, could we instead install the github-changelog-generator gem inside the GitHub Actions pipeline directly rather than using Docker? Does it take a significant amount of time to install or are there other concerns running it outside of Docker?

eike-hass:
> Given that the [Dockerfile commands](https://github.com/eike-hass/docker-github-changelog-generator/blob/master/Dockerfile) are relatively simple, could we instead install the github-changelog-generator gem inside the GitHub Actions pipeline directly rather than using Docker? Does it take a significant amount of time to install or are there other concerns running it outside of Docker?
> 
> E.g.
> 
> ```
> - name: Set up Ruby (assuming this is necessary to use gem)
>   uses: ruby/setup-ruby@477b21f02be01bcb8030d50f37cfec92bfa615b6
>   with:
>     ruby-version: 2.6
> 
> - name: Install github-changelog-generator
>   run: |
>     REPO="https://github.com/github-changelog-generator/github-changelog-generator.git"
>     REF=594a9ccbf84a3172d05d35bb88a0998ed4a948a0
>     gem install specific_install
>     gem specific_install -l ${REPO} -r ${REF}
> ```
> 
> Then alter the run command to call `github-changelog-generator` directly?

Not a fan of having ruby installed in our pipeline TBH. I think Docker abstracts the underlying runtime away nicely and there is an official image we can migrate to once updated which has a blessed setup and pipeline. Speedwise it might be more the or less same I would assume.

cycraig:
It removes an external dependency on your repository/DockerHub account and in my mind makes it simpler to change the pinned revision if need be without publishing a new Docker image (which will always depend on you as it stands).

I'm alright with the Docker container as a temporary fix, just thinking a year or two down the line when everyone has forgotten about this and it needs updating.

eike-hass:
> It removes an external dependency on your GitHub/DockerHub accounts and in my mind makes it simpler to change the pinned revision if need be without publishing a new Docker image (which will always depend on you as it stands).
> 
> I'm alright with the custom Docker image as a temporary fix, just thinking a year or two down the line when everyone has forgotten about this and it needs updating.

Absolutely agree that this must not be a permanent setup. I pinged the guys at the original repo when we can expect an official release and I would like to migrate to that ASAP, which then would be as easy as changing the image.

cycraig (approved):

Issue #531, opened 2021-12-07
[Task] Refactor `Account` to use a single `Client`


cycraig:
## Description
Replace the `ClientMap` in `Account` with a single `Client`.

## Motivation
With the `Account` refactor to handle a single identity per instance in #436 and #453, it makes more sense for each `Account` to use a single `Client` instead of a shared `ClientMap`, given that each identity is linked to a single network (devnet, mainnet, or a private Tangle). There is no longer a need for an `Account` to resolve from or publish to more than one network now.

This would avoid the possibility of subtle bugs that can occur due to `Account`s from the same `AccountBuilder` sharing an `Arc<ClientMap>` where clients for private networks can be overwritten.

This may also open up the `ClientMap` to a refactor for use in a new `Resolver` API #493.

This can still be discussed.

## To-do list

- [ ] Replace `Account::client_map` with a single `Client`

## Change checklist

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PhilippGackstatter:
Generally, I agree with removing the `ClientMap` because an account only works with a single network. The question is, should we already do this, or wait until we have something like a `ClientSet`? This might end up being a single `Client`, which can choose from a number of node urls. How that `ClientSet` is passed node urls will probably concern the `AccountBuilder`, that's why we may want to hold off on this for now? Just a thought to avoid refactoring twice.

cycraig:
I'm not clear on what a `ClientSet` would be and what problem it would solve? I think the `Client` already supports adding multiple nodes simply by wrapping `iota-client`.

PhilippGackstatter:
In that case, my point is moot. This could be an additional task item in the issue then, "Enable users to set node urls (and perhaps quorums?) in the `AccountBuilder`".

cycraig:
It looks like the `ClientBuilder` already has a lot of options for nodes and quorums, e.g.

```rust
  /// Enables/disables quorum.
  pub fn quorum(mut self, value: bool) -> Self {
    self.builder = self.builder.with_quorum(value);
    self
  }

  /// Sets the number of nodes used for quorum.
[...]

PR #582, opened 2022-01-12
Replace `ClientMap` with `Client` in `Account`

cycraig:
# Description of change
Refactors the `Account` to use a single `Client` rather than a `ClientMap`, since each `Account` only manages a single identity (which can only live on a single network/Tangle). This change slightly simplifies the `Account` and `AccountBuilder`.

This also means an `Account` will not be able to create both mainnet and devnet identities by default, which was a side-effect of `ClientMap` automatically creating clients for both those networks by default.

A minor downside of this refactor is that `AccountSetup` always needs a `Client` on creation, which is an async operation. Previously the `Clients` were built lazily by the `ClientMap`. However, this change only affects our internal tests as far as I can see.

Minor change: the private Tangle examples now set the `primary_node` rather than just `node` since the behaviour for `node` differs between Rust and Wasm: on Wasm the node is considered "synchronised", while in Rust the node is considered "un-synchronised" until a background task (from `iota-client`) marks it as "synchronised". This causes the Rust private Tangle examples to throw an error (when pointing to a running private Tangle or devnet or mainnet) since it considers all nodes to be "un-synchronised".

### Added

- Add `AccountBuilder::client_builder()` (in addition to `AccountBuilder::client()` to allow building a single `Client` or sharing an existing `Arc<Client>`)

### Changed

- Update `AccountBuilder::client()` to take `Arc<Client>` instead of a `ClientBuilder` closure.
- Update `Account::create_identity/load_identity` to throw an error if the DID network does not match the `Client` network, which would throw an error when publishing the document anyway.

### Removed

- Remove `Account::set_client()`

Fixes issue #531 

## Type of change

- [ ] Documentation Fix

All tests and examples pass locally.

## Change checklist

### DIFF
--- a/identity-account/src/account/account.rs
+++ b/identity-account/src/account/account.rs
@@ -1,9 +1,9 @@
 // Copyright 2020-2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
-use std::sync::Arc;
 
 use serde::Serialize;
 
@@ -16,29 +16,27 @@ use identity_iota::document::IotaDocument;
 use identity_iota::document::IotaVerificationMethod;
 use identity_iota::document::ResolvedIotaDocument;
 use identity_iota::tangle::Client;
-use identity_iota::tangle::ClientMap;
 use identity_iota::tangle::MessageId;
 use identity_iota::tangle::MessageIdExt;
 use identity_iota::tangle::PublishType;
-use identity_iota::tangle::TangleResolve;
 
 use crate::account::AccountBuilder;
 use crate::account::PublishOptions;
-use crate::error::Result;
+use crate::Error;
 use crate::identity::ChainState;
 use crate::identity::DIDLease;
 use crate::identity::IdentitySetup;
 use crate::identity::IdentityState;
 use crate::identity::IdentityUpdater;
+use crate::Result;
 use crate::storage::Storage;
 use crate::types::KeyLocation;
 use crate::updates::create_identity;
 use crate::updates::Update;
-use crate::Error;
 
+use super::AccountConfig;
 use super::config::AccountSetup;
 use super::config::AutoSave;
-use super::AccountConfig;
 
 /// An account manages one identity.
 ///
@@ -48,13 +46,12 @@ use super::AccountConfig;
 pub struct Account {
   config: AccountConfig,
   storage: Arc<dyn Storage>,
-  client_map: Arc<ClientMap>,
+  client: Arc<Client>,
   actions: AtomicUsize,
   chain_state: ChainState,
   state: IdentityState,
+  // The lease is not read, but releases the DID storage lease when the Account is dropped.
   _did_lease: DIDLease,
-  /* This field is not read, but has special behaviour on drop which is why it is needed in
-   * the Account. */
 }
 
 impl Account {
@@ -77,7 +74,7 @@ impl Account {
     Ok(Self {
       config: setup.config,
       storage: setup.storage,
-      client_map: setup.client_map,
+      client: setup.client,
       actions: AtomicUsize::new(0),
       chain_state,
       state,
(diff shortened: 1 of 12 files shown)


HenriqueNogara (approved):

identity-account/src/account/builder.rs
```diff
@@ -140,36 +136,52 @@ impl AccountBuilder {
     Ok(Arc::clone(self.storage.as_ref().unwrap()))
   }
 
-  /// Apply configuration to the IOTA Tangle client for the given [`Network`].
+  /// Sets the IOTA Tangle [`Client`], this determines the [`Network`] used by the identity.
+  ///
+  /// NOTE: this overwrites any [`ClientBuilder`] previously set by
+  ///       [`AccountBuilder::client_builder'].
```
PhilippGackstatter, line 142: Intentional whitespace? :slightly_smiling_face:

identity-account/src/account/builder.rs
```diff
@@ -140,36 +136,52 @@ impl AccountBuilder {
     Ok(Arc::clone(self.storage.as_ref().unwrap()))
   }
 
-  /// Apply configuration to the IOTA Tangle client for the given [`Network`].
+  /// Sets the IOTA Tangle [`Client`], this determines the [`Network`] used by the identity.
+  ///
```
PhilippGackstatter, line 140: Consider adding a doc comment that all accounts created from the builder with this client set will share that client? Not as a caveat, but as a feature someone may want to use to save on memory?

PhilippGackstatter (changes requested):
Thanks for the changes, looks good overall. I'm wondering whether we still need `IdentitySetup::network`, though.

identity-account/src/account/builder.rs
cycraig: Yes but I suppose it doesn't matter.

identity-account/src/account/builder.rs
cycraig: There's a doc comment regarding that on the `AccountBuilder` itself, I'll duplicate the relevant part for `client()` and `client_builder()`.

identity-account/src/account/account.rs
```diff
@@ -87,10 +86,18 @@ impl Account {
 
   /// Creates a new identity and returns an [`Account`] instance to manage it.
   /// The identity is stored locally in the [`Storage`] given in [`AccountSetup`], and published
-  /// using the [`ClientMap`].
+  /// using the [`Client`].
   ///
   /// See [`IdentitySetup`] to customize the identity creation.
   pub(crate) async fn create_identity(setup: AccountSetup, input: IdentitySetup) -> Result<Self> {
+    // Error if the DID network does not match the client on the Account.
+    // TODO: automatically use client network?
```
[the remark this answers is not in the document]
cycraig: I thought about it (hence the TODO) and I'm in favour of removing it. The only problem I can see is that the network used depends entirely on the client, which could cause subtle bugs/confusion since the network parameter on `Client`/`ClientBuilder` defaults to the `Mainnet`. If I recall correctly there either isn't an error message when publishing a document to the wrong network or the error is vague. 

I'll make the change with added safeguards in `Client` when publishing and hopefully it won't cause any issues.

cycraig:
Should we expose the `Client` in the `Account` by adding an `Account::client()` method?

PhilippGackstatter:
> Should we expose the `Client` in the `Account` by adding an `Account::client()` method?

What is the advantage of exposing the client? The network can be obtained through `account.did().network()` so I guess it's not because of that?

cycraig:
> > Should we expose the Client in the Account by adding an Account::client() method?
>
> What is the advantage of exposing the client? The network can be obtained through account.did().network() so I guess it's not because of that?

To be able to access the `Client` to pass it to e.g. another `AccountBuilder` or `Resolver` is what I had in mind. I don't think it's necessary at this point so I'll leave the API as-is.

PhilippGackstatter:
> To be able to access the `Client` to pass it to e.g. another `AccountBuilder` or `Resolver` is what I had in mind. I don't think it's necessary at this point so I'll leave the API as-is.

I see, that makes sense. In #574 a getter for the [storage was added](https://github.com/iotaledger/identity.rs/pull/574/files#diff-e1f81a8834ec1f84e3102d6ff186b8f34fc26afb4aaeaa87a2416fc8acd18db5R124-R127), so maybe something similar will be required for the client as well.

Issue #493, opened 2021-11-15
[Task] Implement new Resolver API


JelleMillenaar:
## Description

Old Description: As has been discussed at length in discussion #422, the Resolver logic will be completely refactored with a new API design. It will be able to internally cache the DID message data from the Tangle and return metadata. An additional DID spec compliant API will be introduced in a later issue. 

New Description: We have discussed a new Resolver API, that crates a consistent object that can be used for Verification throughout an application. It works for multiple resolutions, may work with multiple networks (main, dev, shimmer) and can resolve different methods (iota, stardust?, disposable?). The same Verifier API can also take care of VC and VP verification. DID resolution and VC & VP verification may be done with resolution/verification options such as `resolve_at` or `verifiy_at`. Caching is out of scope for this PR. 

For details about the API design, I'd refer to #422 and the comments on this Issue.

## Motivation

Improve developer experience with a clean, easy yet powerful API.

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Implement new API
    - Partially addressed in #594
- [x] Update Examples
    - Requires #599

cycraig:
As per internal discussions and #422, using the initial resolver API might look something like the following:

```rust
// A Resolver should either contain a ClientMap or it could replace it entirely.
let client_map = ClientMap::new().await;
client_map.insert(Client::from_network(Network::Devnet).await?);

// Build a resolver that caches calls to the tangle for this specific did
let resolver = Resolver::new(client_map);

// Resolve the latest version of a DID document (basically the same as Client::resolve right now),
// where IotaTangleDocument is a new wrapper struct from the metadata refactor.
let iota_did: IotaDID = "did:iota:...".parse().unwrap();
let document: IotaTangleDocument = resolver.resolve(&iota_did).await?;

// Try resolve a DID document from a specific version.
let old_document: IotaTangleDocument = resolver.resolve_with_options(
  &iota_did, 
  ResolveOptions::new().version_id("1.5")
).await?;

// Try resolve a DID document from a specific timestamp.
let old_document: IotaTangleDocument = resolver.resolve_with_options(
  &iota_did, 
  ResolveOptions::new().version_time("2021-10-01T00:00:00Z")
).await?;

// Separate API, but can share the network cache of the created resolver instance.
let history: DocumentHistory = resolver.resolve_history(&iota_did).await?;

// Verify a presentation or credential, fetching issuer and subject DID documents automatically.
[...]

Issue #532, opened 2021-12-07
[Task] Remove old resolution module


cycraig:
## Description
~The `identity-did/resolution` module should be split into its own crate, e.g. `identity-resolver` or `identity-resolution`.~

Edit: the `identity-did/resolution` module should be removed.

## Motivation
The module for DID specification-compliant resolution and dereferencing in `identity-did/resolution` is largely unused right now and distinct from the rest of the codebase. ~It's useful for a universal resolver in the future so we should at least retain the code.~

Edit: there doesn't seem to be a reason to keep the code anymore, a universal resolver can be implemented easily without this.

## Resources
https://www.w3.org/TR/did-core/#resolution

## To-do list

- [ ] ~Move `identity-did/resolution` into its own feature-gated crate.~
- [ ] Remove `identity-did/resolution` and all associated code.

## Change checklist

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

cycraig:
Update: it was decided to remove the old resolution code entirely.

PR #594, opened 2022-01-17
Replace `ClientMap` with new `Resolver`

cycraig:
# Description of change
Adds the new `Resolver` API discussed in #422 and #493.

The new `Resolver` replaces the current `ClientMap` and is very similar in design. The differences are:
- `Resolver` does not expose `publish_*` methods. This is to encourage the use of the `Account` over the low-level API.
- `Resolver` does not automatically link to both the Devnet and Mainnet networks (old behaviour of `ClientMap`). It has a default constructor that creates a `Client` for the Mainnet but this can be overridden using the `ResolverBuilder`.
- `Resolver` contains convenience functions for resolving DID Documents associated with verifiable `Credentials` and `Presentations`.

### What is *not* included in this PR?
- Convenience functions for resolving and verifying credentials and presentations at the same time; depends on #599.
- Refactoring the credential and presentations examples; depends on #599.
- The Wasm bindings `Client` functions for `check_credential` and `check_presentation` are not moved to the `Resolver`; depends on #599.
- `Resolver::resolve_with_options` from #493. We can add this but the future of resolving document histories is uncertain and method-dependent. Only `version_time`/`version_datetime` would be supported if we implement this right now, as we do not have document versioning.
- This does not alter the `resolution` module; see issue #532.

### Why is this different from the outcome of #422?

As mentioned in #493, it was decided after internal discussions to avoid implementing a compliant resolver API as it is more limiting than the functionality we expose through `identity.rs` directly. Many of its features like which encoding to accept, for example, are centred around web/API resolution rather than the strongly-typed framework we provide. Universal resolver implementations can still use `identity.rs` to implement such a resolver quite easily.

### Added

- Add `Resolver`, `WasmResolver`.
- Add `ResolverBuilder`

### Removed

- Remove `ClientMap`.

# Open Questions

1. Should we revert to a `ClientMap`-like interface where we expose helper functions for publishing across different networks?
2. Should we expose `ResolverBuilder` to the Wasm bindings in order to remove `set_client` and `remove_client` from `Resolver`?

Fixes issue #493 

## Type of change

- [ ] Documentation Fix

Several VC examples have been updated to use the `Resolver` but only in Rust. Will be more prevalent after #599.

## Change checklist


identity-iota/src/tangle/resolver.rs
```diff
@@ -0,0 +1,220 @@
...
+        ClientOrBuilder::Builder(builder) => ClientRc::new(builder.build().await?),
+      };
+      client_map.insert(network_name, client);
+    }
+
+    Ok(Resolver { client_map })
+  }
+}
+
+#[async_trait::async_trait(? Send)]
```
PhilippGackstatter, line 215:
```suggestion
#[async_trait::async_trait(?Send)]
```

Unintended space?

identity-iota/src/tangle/resolver.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// A `Resolver` supports resolving DID Documents across different Tangle networks using
+/// multiple [`Clients`][Client].
+///
+/// Also provides convenience functions for resolving DID Documents associated with
+/// verifiable [`Credentials`][Credential] and [`Presentations`][Presentation].
+#[derive(Debug)]
+pub struct Resolver {
+  // TODO: is DashMap really necessary here?
+  //  We can avoid mutation if we disallow adding/removing Clients post-creation.
```
PhilippGackstatter, line 43: Makes a lot of sense to me to setup a resolver once at build-time and then disallow mutation afterwards. We don't have to commit to allowing mutability now, but if the need arises later, adding it is non-breaking.

PhilippGackstatter (approved):
Looks good to me!

I think the `DashMap` can be replaced with a `HashMap`. The resolver should be useful enough without post-creation mutability.

identity-iota/src/tangle/resolver.rs
cycraig: Sorry, Intellij formatter keeps adding a space there.

identity-iota/src/tangle/resolver.rs
cycraig: Cool, I guess we need the `ResolverBuilder` in the Wasm bindings then? It was not ported since there is no good way to avoid the builder pattern in Wasm in this case, but it would be the only option if we remove the mutators.

PhilippGackstatter: Can we not take a list of `ClientOrBuilder`s in `WasmResolver::new`, or even just `Client`s? Or is that the not-good-way you meant :smile:?

```ts
const resolver = new Resolver([
      Client.fromConfig(Config.fromNetwork(Network.mainnet())),
      Client.fromConfig(Config.fromNetwork(Network.devnet())),
  ]);
```

Is the problem that someone could provide two clients with the same network name?

identity-iota/src/tangle/resolver.rs
```diff
@@ -0,0 +1,220 @@
...
+
+  /// Removes the [`Client`] corresponding to the given [`NetworkName`] and returns it
+  /// if one exists.
+  // TODO: remove this?
+  pub fn remove_client(&self, network_name: &NetworkName) -> Option<ClientRc> {
+    self.client_map.remove(network_name).map(|(_, client)| client)
+  }
+
+  /// Returns the [`Client`] corresponding to the given [`NetworkName`] if one exists.
+  pub fn get_client(&self, network_name: &NetworkName) -> Option<ClientRc> {
```
olivereanderson, line 81: Consider the following scenario:  
Crate x: depends on identity-iota without the `arc` feature 
in this crate there is a function 
```
pub fn get_client() -> Option<Rc<Client>> {
  let resolver: Resolver = .. ;
  resolver.get_client()
}
```
Crate y: depends on identity-iota with the `arc` feature 
Crate z: depends directly on Crate x and has a function 
`fn foo(client: Rc<Client>); `
The following code works in crate z: 
```
let client = crate_x::get_client();
foo(client);
```
but if crate y ever becomes part of crate z's dependency tree (possibly via transitive dependencies) I think crate `z`will no longer compile? 

This is probably the reason why the `im` crate comes in two flavours `im` and `im-rc` (See https://docs.rs/im/latest/im/#thread-safety) instead of changing the pointer types via feature flags. 

We might want to consider using the [`archery crate`](https://docs.rs/archery/latest/archery/) and `Resolver` generic over the pointer type. Then the `new` constructor in the `WasmResolver` can use the `Rc` variant and in the Rust library we either remove `new` and require construction through the builder or use an `Arc` there.

PhilippGackstatter: > Dependencies are resolved with the union of all features enabled on them. For example, if one package depends on the [im](https://crates.io/crates/im) package with the serde dependency enabled and another package depends on it with the rayon dependency enabled, then im will be built with both features enabled, and the serde and rayon crates will be included in the resolve graph. If no packages depend on im with those features, then those optional dependencies will be ignored, and they will not affect resolution. https://doc.rust-lang.org/cargo/reference/resolver.html#features

I guess you're right that if our crate is a dependency of two other crates, and one enables `arc` and the other doesn't, there will be a conflict.

The `archery` crate looks interesting.

cycraig: Great point, I agree that putting `arc` as a feature isn't ideal since it's not additive and can lead to the situation you described.

I'll try `archery` for now since we would want to use the same pattern to fix the `Account` using `Arc` instead of `Rc` in Wasm.

identity-iota/src/tangle/resolver.rs
cycraig: As far as I know there is no way to take `Vec<WasmClient>` as a parameter (`Vec<T>` is not supported in general), and if I recall correctly we cannot cast `JsValue` to one of our exported structs, and since `Client` and `ClientBuilder` are not serializable (for good reason), we cannot use a duck-typed Typescript interface either.

I'll try again just to be sure though.

PhilippGackstatter: Ah, yeah, my bad, I was thinking along the lines of the `serde` solution, but of course that doesn't work with `Client`s. Perhaps a map would work alternatively, where the keys are network names and the values `Client`s. Using [`Reflect::own_keys`](https://docs.rs/js-sys/latest/js_sys/Reflect/fn.own_keys.html) we could iterate the keys and retrieve the clients. That would even prevent duplication.

identity-iota/src/tangle/resolver.rs
```diff
@@ -0,0 +1,220 @@
...
+  pub async fn resolve_presentation_issuers(&self, presentation: &Presentation) -> Result<Vec<ResolvedIotaDocument>> {
+    // Extract unique issuers.
+    let mut issuers: HashSet<IotaDID> = HashSet::new();
+    for credential in presentation.verifiable_credential.iter() {
+      let issuer: IotaDID = IotaDID::parse(credential.issuer.url().as_str())?;
+      issuers.insert(issuer);
+    }
+
+    // Resolve issuers concurrently.
+    futures::future::try_join_all(issuers.iter().map(|issuer| self.resolve(issuer)).collect::<Vec<_>>()).await
```
olivereanderson, line 144:
```
    // Extract unique issuers.
    let issuers: Result<HashSet<IotaDID>> = presentation
      .verifiable_credential
      .iter()
      .map(|credential| IotaDID::parse(credential.issuer.url().as_str()))
      .collect();

    // Resolve issuers concurrently.
    futures::future::try_join_all(issuers?.iter().map(|issuer| self.resolve(issuer)).collect::<Vec<_>>()).await
```

identity-iota/src/tangle/resolver.rs
cycraig: I think that returns the keys of the map/object, not the values (`Clients`)? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/ownKeys#using_reflect.ownkeys

identity-iota/Cargo.toml
```diff
@@ -50,7 +50,10 @@ features = ["blake2b"]
 tokio = { version = "1.15", default-features = false, features = ["macros"] }
 
 [features]
-default = ["async"]
+default = ["arc", "async"]
+
+# Changes smart pointers in Resolver from Rc to Arc.
```
olivereanderson, line 55: Not sure if this should be done with feature flags. See [my comment to `get_client`](https://github.com/iotaledger/identity.rs/pull/594/files#r812729822).

identity-iota/src/tangle/resolver.rs
PhilippGackstatter: Yes, this would actually be pretty ugly typing-wise, as we'd have to follow-up with `Reflect::get` which only returns a `JsValue` as you pointed out, and we'd have to downcast that to `Client` similar to how WasmStorage does that. That should be avoided if possible, though. Exposing the builder seems like the best option after all. Sorry for the half-baked ideas.

cycraig: Ideally all this would be avoided if `wasm-bindgen` supported `dyn_ref`/`dyn_into` on exported types, but it will probably never happen: https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen/trait.JsCast.html#method.dyn_into

Wow that downcasting workaround is scary. I mean it works but there's got to be a better way than relying on the constructor name: https://github.com/rustwasm/wasm-bindgen/issues/2231#issuecomment-656293288

bindings/wasm/src/tangle/resolver.rs
```diff
@@ -0,0 +1,231 @@
...
+  pub fn resolve_presentation_issuers(&self, presentation: &WasmPresentation) -> Result<PromiseArrayResolvedDocument> {
+    // TODO: reimplemented function to avoid cloning the entire presentation.
+    //       Would be solved with Rc internal representation, pending memory leak discussions.
+
+    // Extract unique issuers.
+    let mut issuers: HashSet<IotaDID> = HashSet::new();
+    for credential in presentation.0.verifiable_credential.iter() {
+      let issuer: IotaDID = IotaDID::parse(credential.issuer.url().as_str()).wasm_result()?;
+      issuers.insert(issuer);
+    }
```
olivereanderson, line 170: [This part can be simplified in the same way as described in this comment](https://github.com/iotaledger/identity.rs/pull/594/files#r812751853) also applies here.

identity-iota/src/tangle/resolver.rs
cycraig: So the problem with `archery` I've found is that it does not support converting from `Arc`/`Rc` to their `SharedPointer`, nor vice-versa (at least as far as I can tell).

This is a problem because we explicitly use `Rc<Client>` in `WasmClient` rather than `SharedPointer<Client, RcK>` (which the `Resolver` uses), so it's not possible to convert the `Resolver` entries into `WasmClient`.

E.g.
```rust 
#[wasm_bindgen(js_name = getClient)]
pub fn get_client(&self, network_name: String) -> Option<WasmClient> {
  let network_name: NetworkName = NetworkName::try_from(network_name).ok()?;
  let client: SharedPointer<Client, RcK> = self.0.get_client(&network_name).cloned()?;
  let rc_client: Rc<Client> = client.into(); // not possible with any safe function archery exposes.
  Some(WasmClient { client: rc_client })
}
```

The reverse also appears not possible.

E.g.
```rust
#[wasm_bindgen]
pub fn client(mut self, client: &WasmClient) -> WasmResolverBuilder {
  let client: Rc<Client> = client.client.clone();
  let shared_client: SharedPointer<Client, RcK> = client.into(); // not possible with any safe function archery exposes.
  self.0 = self.0.client(shared_client);
  self
}
```

identity-iota/src/tangle/resolver.rs
cycraig: Exposed `WasmResolverBuilder`. Using it looks like this in Javascript:

```ts
const client = new Client();

const resolver = await Resolver
        .builder()
        .client(client)
        .clientConfig(Config.fromNetwork(Network.devnet()))
        .build();
```

identity-iota/src/tangle/resolver.rs
cycraig: One solution would be to use `SharedPointer<Client, RcK>` exclusively in the Wasm bindings though, if we want to rely on `archery` completely? 

I don't think this is very friendly to Rust developers though, since I imagine they might want something like a base `Arc<Client>` eventually.

olivereanderson:
```
let shared_client: SharedPointer<Client, RcK> = client.into(); // not possible with any safe function archery exposes.
```

Can't you use https://docs.rs/archery/latest/archery/shared_pointer/struct.SharedPointer.html#method.new ?

cycraig: `SharedPointer::new` takes `T`, not `Rc<T>`.

olivereanderson: Can we instead of using `archery` just use the sealed trait pattern like so:

```
mod private {
  use super::Client;
  pub(super) trait Sealed {}
  impl Sealed for std::sync::Arc<Client> {}
  impl Sealed for std::rc::Rc<Client> {}
}

```
pub struct Resolver<T: private::Sealed> {
  // TODO: is DashMap really necessary here?
  //  We can avoid mutation if we disallow adding/removing Clients post-creation.
  client_map: DashMap<NetworkName, T>,
}
```
?

Comment (cycraig) [reply identity-iota/src/tangle/resolver.rs]: Yeah there's two approaches I'm trying to hack together now. 

1. Your suggestion of using `Arc<Client>`/`Rc<Client>` as the generic itself. 
2. A trait with an associated type for `Rc` or `Arc`. https://github.com/orium/rpds/issues/7#issuecomment-362635901 , https://joshlf.com/post/2018/10/18/rust-higher-kinded-types-already/#comment-4160863400

Comment (cycraig) [reply identity-iota/src/tangle/resolver.rs]: This is what I have currently. It works and doesn't require any odd hacks. 

We can restrict it using a sealed trait to prevent unexpected entries for `T` from dependents, but I don't think it's too bad as-is?

```rust
pub struct Resolver<T = Arc<Client>>
where
  T: Clone + AsRef<Client> + From<Client>,
{
  client_map: HashMap<NetworkName, T>,
}
```

```rust
// Developers using it in Rust will default to Resolver<Arc<Client>.
let resolver: Resolver = Resolver::new().await?; 

// Changing to Rc is a bit verbose but not bad.
let resolver: Resolver<Rc<Client>> = Resolver::<Rc<Client>>::new().await?;
```

Using a generic trait with an associated type of `Rc` or `Arc` for a generic `T` (rather than just `Client`) is another option but unnecessary in our case since we only require it for `Client`, in my opinion.

PhilippGackstatter: I think that's perfect :+1:

cycraig: Actually, I think you're right that a sealed trait might be better so we can change it without affecting semantic versioning.

cycraig: Thanks for all the feedback. I'll leave it as-is.

olivereanderson (approved):
Looks great now. Thanks for the changes.

PR #597, opened 2022-01-17
Feat/wasm storage

HenriqueNogara:
# Description of change
Expose `Storage` trait in the bindings.

## Type of change
- [ ] Documentation Fix

## Change checklist


bindings/wasm/src/account/chain_state.rs
```diff
@@ -0,0 +1,67 @@
...
+  #[wasm_bindgen(js_name = setLastIntegrationMessageId)]
+  pub fn set_last_integration_message_id(&mut self, message: WasmMessageId) {
+    self.0.set_last_integration_message_id(message.into())
+  }
+
+  /// Sets the last diff message id.
+  #[wasm_bindgen(js_name = setLastDiffMessageId)]
+  pub fn set_last_diff_message_id(&mut self, message: WasmMessageId) {
+    self.0.set_last_diff_message_id(message.into())
+  }
```
cycraig, line 48: Discussion: we have previously used `string` wherever `MessageId` is needed in the bindings. Should we embrace exposing 1-1 types everywhere and thus update existing instances with `WasmMessageId` or continue using `string`? I do think we need to be consistent and not have `string` in some places and `WasmMessageId` in others.

bindings/wasm/src/account/crypto.rs
```diff
@@ -0,0 +1,24 @@
...
+#[wasm_bindgen(js_name = EncryptionKey, inspectable)]
+pub struct WasmEncryptionKey(EncryptionKey_);
+
+#[wasm_bindgen(js_class = EncryptionKey)]
+impl WasmEncryptionKey {
+  #[wasm_bindgen(js_name = deriveEncryptionKey)]
+  pub fn derive_encryption_key(password: &str) -> WasmEncryptionKey {
+    WasmEncryptionKey(derive_encryption_key_(password))
+  }
+}
```
cycraig, line 18: The internal type of `EncryptionKey` is simply `[u8; 32]`. I think we should consider just using `Uint8Array` or add a converter function? How would a Javascript `Storage` implementation actually use this as a password, can they access the bytes through `inspectable`?

bindings/wasm/src/account/generation.rs
```diff
@@ -0,0 +1,78 @@
...
+use wasm_bindgen::prelude::*;
+
+use identity::account::Generation as Generation_;
+
+use crate::error::Result;
+use crate::error::WasmResult;
+
+#[wasm_bindgen(js_name = Generation, inspectable)]
+#[derive(Clone, Copy, Default, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+pub struct WasmGeneration(pub(crate) Generation_);
```
cycraig, line 13: Aside: do we actually need all the derived traits on Wasm wrappers? I don't think we actually use them anywhere so this could just be code bloat?

bindings/wasm/src/account/identity_state.rs
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  // ===========================================================================
+  // Document State
+  // ===========================================================================
+
+  #[wasm_bindgen(getter)]
+  pub fn document(&self) -> WasmDocument {
+    self.0.document().clone().into()
+  }
```
cycraig, line 66: As discussed in #603 I don't think this should have the getter annotation, and it should be indicated in a doc-comment that this clones the data.

bindings/wasm/src/core/fragment.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use wasm_bindgen::prelude::*;
+
+use identity::core::Fragment as Fragment_;
+
+#[wasm_bindgen(js_name = Fragment, inspectable)]
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+pub struct WasmFragment(pub(crate) Fragment_);
```
cycraig, line 10: `WasmFragment` is another instance where we have typically used `strings` elsewhere in the Wasm bindings for convenience. We should evaluate whether it's worth exporting this at all or just having fallible functions which take `string` and attempt to convert it to the required type.

bindings/wasm/src/account/traits.rs
```diff
@@ -0,0 +1,222 @@
...
+use crate::account::WasmIdentityState;
+use crate::account::WasmKeyLocation;
+use crate::did::WasmDID;
+use crate::error::JsValueResult;
+
+#[wasm_bindgen]
+extern "C" {
+  pub type WasmStorage;
+  #[wasm_bindgen(typescript_type = "Promise<()>")]
+  pub type PromiseUnit;
```
cycraig, line 37: Is `Promise<()>` the correct equivalent in Typescript? Should it not be something like `Promise<void>`?

bindings/wasm/src/account/traits.rs
```diff
@@ -0,0 +1,222 @@
...
+  #[wasm_bindgen(typescript_type = "Promise<Signature>")]
+  pub type PromiseSignature;
+  #[wasm_bindgen(typescript_type = "Promise<bool>")]
+  pub type PromiseBool;
+  #[wasm_bindgen(typescript_type = "Promise<Option<Generation>>")]
+  pub type PromiseOptionGeneration;
+  #[wasm_bindgen(typescript_type = "Promise<Option<ChainState>>")]
+  pub type PromiseOptionChainState;
+  #[wasm_bindgen(typescript_type = "Promise<Option<IdentityState>>")]
+  pub type PromiseOptionIdentityState;
```
cycraig, line 51: Nit: maybe move the `Promise` definitions to a separate `extern "C"` block to make it easier to see the `WasmStorage` definitions.

bindings/wasm/src/account/traits.rs
```diff
@@ -0,0 +1,222 @@
...
+
+  #[wasm_bindgen(method)]
+  pub fn set_password(this: &WasmStorage, password: WasmEncryptionKey) -> PromiseUnit;
```
cycraig, line 54: I'm still unclear on how an implementor would use `WasmEncryptionKey`. Is it worth making a proper e.g. `JavascriptMemStore` implementation in Javascript/Typescript to explore whether all the exported types are actually usable?

bindings/wasm/src/account/traits.rs
```diff
@@ -0,0 +1,222 @@
...
+  #[wasm_bindgen(method)]
+  pub fn flush_changes(this: &WasmStorage) -> PromiseUnit;
+  #[wasm_bindgen(method)]
```
cycraig, line 57: All the `Storage` functions are fallible right? I think we need to use the `catch` attribute to manage exceptions from Javascript: https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/catch.html

bindings/wasm/src/error.rs
```diff
@@ -110,3 +114,53 @@ impl From<identity::iota::BeeMessageError> for WasmError<'_> {
...
+
+impl_from_js_value!(
+  identity::account::DIDLease,
+  identity::crypto::PublicKey,
+  identity::account::Signature,
+  Option<identity::account::Generation>,
+  Option<identity::account::ChainState>,
+  Option<identity::account::IdentityState>,
+  bool
+);
```
cycraig, line 153: Hmm, it's really annoying that we have to deserialize through `JsValue` as a result of `js_sys::Promise`. I don't see a way around this though...

cycraig (commented):
I am still super impressed that you managed to port the `Storage` trait. 

Preliminary comments below (apologies for the noisy comments). Primarily questions around some of the types, whether they are necessary, and how implementors would use them.

bindings/wasm/src/account/chain_state.rs
HenriqueNogara: I agree about being consistent. I know sometimes it can be annoying, but I would prefer if we could have 1-1 types. That way we can make it more type-safe I guess.

identity-account/src/storage/stronghold.rs
```diff
@@ -83,7 +84,8 @@ impl Stronghold {
   }
 }
 
-#[async_trait::async_trait]
+#[cfg_attr(feature = "wasm", async_trait(?Send))]
+#[cfg_attr(not(feature = "wasm"), async_trait)]
```
PhilippGackstatter, line 88: We won't be exposing `Stronghold` to Wasm anyway, so we can actually remove the `cfg_attr`s completely here and go with `async_trait` directly.

bindings/wasm/src/account/traits.rs
```diff
@@ -0,0 +1,223 @@
...
+  /// Attempt to obtain the exclusive permission to modify the given `did`.
+  /// The caller is expected to make no more modifications after the lease has been dropped.
+  /// Returns an [`IdentityInUse`][crate::Error::IdentityInUse] error if already leased.
+  async fn lease_did(&self, did: &IotaDID) -> AccountResult<DIDLease> {
+    let promise: Promise = Promise::resolve(&self.lease_did(did.clone().into()));
+    let result: JsValueResult = JsFuture::from(promise).await.into();
+    // workaround due to problems deserializing `DIDLease`, even with serde "rc" feature
+    let account_result: AccountResult<bool> = result.into();
+    account_result.map(|value| DIDLease::from(AtomicBool::from(value)))
+  }
```
PhilippGackstatter, line 122: It's not necessary to serialize the lease because it doesn't need to be stored, it just needs to exist in memory. Using just a bool won't work correctly, because while in JS you can set `"did:iota:123...": true` to store that it's in use, it's never going to be set back to `false` when the lease is dropped in Rust. Something like that should be all that's needed:

```rust
#[wasm_bindgen(js_name = DIDLease)]
pub struct WasmDIDLease(pub(crate) DIDLease);

#[wasm_bindgen]
impl WasmDIDLease {
  #[wasm_bindgen(constructor)]
  pub fn new() -> Self {
    WasmDIDLease(DIDLease::new())
  }

  #[wasm_bindgen(constructor)]
  pub fn load(&self) -> bool {
    self.0.load()
  }

  #[wasm_bindgen(constructor)]
  pub fn store(&self, value: bool) {
    self.0.store(value)
  }
}
```

That way, when the lease is dropped, the inner `AtomicBool` is set to false which should enable the correct behavior. But we need to make sure the `WasmDIDLease` isn't dropped to early somewhere. I'm probably missing some other subtleties, but with tests or an example we should see if it works.

identity-core/src/crypto/key/key.rs
```diff
@@ -22,11 +22,11 @@ mod private {
...
 
 // A marker type for the `Public` components of an asymmetric cryptographic key.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Deserialize)]
 pub enum Public {}
 
 // A marker type for the `Private` components of an asymmetric cryptographic key.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Deserialize)]
 pub enum Private {}
```
PhilippGackstatter, line 30: Would an alternative to deriving `Deserialize` be to "serialize" manually to and from bytes with `Key::as_ref` and `Key::from`, since keys are just wrappers around bytes? This is not a request to change anything, just a question whether `Deserialize` has inherent benefits?

PhilippGackstatter (changes requested):
Agree with Craig, really cool that this actually all compiles now! It's a much better solution than using anything with `unsafe`.

I think some methods on some types aren't really needed because they just need to be serialized on the Wasm side, so they can be stored, but aren't actually used when implemeting a storage.

bindings/wasm/src/account/chain_state.rs
```diff
@@ -0,0 +1,67 @@
...
+  pub fn is_new_identity(&self) -> bool {
+    self.0.is_new_identity()
+  }
+}
+
+impl Default for WasmChainState {
+  fn default() -> Self {
+    Self::new()
+  }
+}
```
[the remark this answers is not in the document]
HenriqueNogara: I only added because clippy issued a warning: 
```
you should consider adding a `Default` implementation for `WasmChainState`
```

bindings/wasm/src/account/identity_state.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::did::WasmMethodType;
+use crate::error::Result;
+use crate::error::WasmResult;
+
+#[wasm_bindgen(js_name = IdentityState, inspectable)]
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct WasmIdentityState(pub(crate) IdentityState_);
+
+#[wasm_bindgen(js_class = IdentityState)]
+impl WasmIdentityState {
```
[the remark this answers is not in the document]
HenriqueNogara: Yes, you are right

bindings/wasm/src/account/traits.rs
HenriqueNogara: I'll add an example to double check if we need all of this

identity-core/src/crypto/key/key.rs
HenriqueNogara: If I understood correctly, `js_value.into_serde()` wouldn't work

HenriqueNogara: We might not need this afterall

bindings/wasm/src/error.rs
```diff
@@ -110,3 +114,53 @@ impl From<identity::iota::BeeMessageError> for WasmError<'_> {
...
+  ( $($t:ty),* ) => {
+    $(impl From<JsValueResult> for AccountResult<$t> {
+      fn from(result: JsValueResult) -> Self {
+        result
+          .0
+          .map_err(|js_value| AccountError::PromiseError(js_value.as_string().unwrap_or_default()))
+          .and_then(|js_value| {
+            js_value
+              .into_serde()
+              .map_err(|e| AccountError::InvalidJsValue(e.to_string()))
```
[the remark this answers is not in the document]
HenriqueNogara: Moving to a more generic name is fine by me. Also think a fatal error would be appropriate

eike-hass:
Can we expose a Typescript interface (or an abstract class) for the storage interface so we could do `class MemStore implements/extends Storage { ...`

eike-hass:
> Can we expose a Typescript interface (or an abstract class) for the storage interface so we could do `class MemStore implements/extends Storage { ...` ?

maybe even automatically with something like https://github.com/Aleph-Alpha/ts-rs?

eike-hass:
I now this still WIP but do we want the examples to run in nightly action as well?

eike-hass:
> > > Can we expose a Typescript interface (or an abstract class) for the storage interface so we could do `class MemStore implements/extends Storage { ...` ?
> > 
> > 
> > maybe even automatically with something like https://github.com/Aleph-Alpha/ts-rs?
> 
> Would something like this suffice?
> 
> ```
> #[wasm_bindgen(typescript_custom_section)]
> const STORAGE: &'static str = r#"
> /** All methods an object must implement to be used as an account storage. */
> interface Storage {
>   setPassword: (encryptionKey: Uint8Array) => Promise<void>;
>   flushChanges: () => Promise<void>;
>   leaseDid: (did: DID) => Promise<DIDLease>;
>   keyNew: (did: DID, keyLocation: KeyLocation) => Promise<string>;
>   keyInsert: (did: DID, keyLocation: KeyLocation, privateKey: string) => Promise<string>;
>   keyExists: (did: DID, keyLocation: KeyLocation) => Promise<boolean>;
>   keyGet: (did: DID, keyLocation: KeyLocation) => Promise<string>;
>   keyDel: (did: DID, keyLocation: KeyLocation) => Promise<void>;
>   keySign: (did: DID, keyLocation: KeyLocation, data: Uint8Array) => Promise<Signature>;
>   chainState: (did: DID) => Promise<ChainState>;
>   setChainState: (did: DID, chainState: ChainState) => Promise<void>;
>   state: (did: DID) => Promise<IdentityState>;
>   setState: (did: DID, identityState: IdentityState) => Promise<void>;
>   purge: (did: DID) => Promise<void>;
>   publishedGeneration: (did: DID) => Promise<Generation>;
>   setPublishedGeneration: (did: DID, generation: Generation) => Promise<void>;
> }"#;
> ```

That looks very promising! I would assume we can remove basically all the type annotation from `memory_storage.ts`.
One more thought on this: Where do we plan to document what those methods are supposed achieve? In Rust we have e.g.:
```
/// Returns `true` if a keypair exists at the specified `location`.
  async fn key_exists(&self, did: &IotaDID, location: &KeyLocation) -> AccountResult<bool> {
```
Do we replicate that comment for ts? Or do we centrally (in the wiki) describe how to implement storage?

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -0,0 +1,307 @@
...
+   * The caller is expected to make no more modifications after the lease has been dropped.
+   * Returns an error if already leased.
+  */
+  leaseDid: (did: DID) => Promise<DIDLease>;
+
+  /** Creates a new keypair at the specified `location`.*/
+  keyNew: (did: DID, keyLocation: KeyLocation) => Promise<string>;
+
+  /** Inserts a private key at the specified `location`.*/
+  keyInsert: (did: DID, keyLocation: KeyLocation, privateKey: string) => Promise<string>;
```
PhilippGackstatter, line 255: - All the places that take a key as a string or return one, should document that the (expected) format is base58. 
- The documentation of the storage trait in Rust is also lacking a bit, we should perhaps fix that as well. `key_*` functions that are expected to return a public key should state so in the documentation. This is particularly important for the JS version where a `string` is a less meaningful return type than `PublicKey` in Rust.

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -0,0 +1,307 @@
...
+  purge: (did: DID) => Promise<void>;
+
+  /** Returns the last generation that has been published to the tangle for the given `did`.*/
+  publishedGeneration: (did: DID) => Promise<Generation>;
+
+  /** Sets the last generation that has been published to the tangle for the given `did`.*/
+  setPublishedGeneration: (did: DID, generation: Generation) => Promise<void>;
+}"#;
+
+pub fn downcast_js_value<T: FromWasmAbi<Abi = u32>>(js_value: JsValue, classname: &str) -> Result<T, AccountError> {
```
PhilippGackstatter, line 291:
```suggestion
// from https://github.com/rustwasm/wasm-bindgen/issues/2231#issuecomment-656293288
pub fn downcast_js_value<T: FromWasmAbi<Abi = u32>>(js_value: JsValue, classname: &str) -> Result<T, AccountError> {
```

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -0,0 +1,307 @@
...
+  keyInsert: (did: DID, keyLocation: KeyLocation, privateKey: string) => Promise<string>;
+
+  /** Returns `true` if a keypair exists at the specified `location`.*/
+  keyExists: (did: DID, keyLocation: KeyLocation) => Promise<boolean>;
+
+  /** Retrieves the public key at the specified `location`.*/
+  keyGet: (did: DID, keyLocation: KeyLocation) => Promise<string>;
+
+  /** Deletes the keypair specified by `location`.*/
+  keyDel: (did: DID, keyLocation: KeyLocation) => Promise<void>;
```
PhilippGackstatter, line 264: Is `keyDel` supposed to be idempotent? This came up when I looked over the memory_storage.ts implementation, which behaved slightly differently from the Rust `MemStore`. Idempotency seems like a good idea to me here, and we should state that in the doc as well, then. That would also require changes to all implementations. Feel free to leave this for a later PR, in which case we should open an issue, so it's not forgotten about. I can open the issue if there is agreement for that.

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -0,0 +1,307 @@
...
+  setState: (did: DID, identityState: IdentityState) => Promise<void>;
+
+  /** Removes the keys and any state for the identity specified by `did`.*/
+  purge: (did: DID) => Promise<void>;
+
+  /** Returns the last generation that has been published to the tangle for the given `did`.*/
+  publishedGeneration: (did: DID) => Promise<Generation>;
+
+  /** Sets the last generation that has been published to the tangle for the given `did`.*/
+  setPublishedGeneration: (did: DID, generation: Generation) => Promise<void>;
```
PhilippGackstatter, line 288: I assume this will be removed in the napi or epic branch?

PhilippGackstatter (changes requested):
Thanks for the memory store implementation in TS!

I still have two primary concerns.

- There are still a lot of types that seem unnecessarily exposed. On the flipside, I'm concerned we might be missing places where the types that appear in the storage trait do not have (de)serialization functions exposed. Ideally, we would verify that these types can all be (de)serialized in the memory_storage.ts implementation, by turning it into a file store, or something like that. In-memory by itself also seems fine, as long as all types go through serialization at some point.

- The other bigger concern is around documentation of functions, mostly relating to stating the base58 requirement, and if we actually want to keep that, since it requires lots of encoding from implementors, while we could just as easily expose these things as bytes.

bindings/wasm/examples-account/src/memory_storage.ts
```diff
@@ -0,0 +1,159 @@
...
+        this._vaults.delete(did);
+    }
+    
+    public async publishedGeneration(did: DID): Promise<Generation> {
+        return this._publishedGenerations.get(did)
+    }
+    
+    public async setPublishedGeneration(did: DID, generation: Generation): Promise<void> {
+        this._publishedGenerations.set(did, generation);
+    }
```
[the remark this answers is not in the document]
HenriqueNogara: handled on #660

bindings/wasm/src/account/types/generation.rs
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+impl From<WasmGeneration> for Generation {
+  fn from(wasm_generation: WasmGeneration) -> Self {
+    wasm_generation.0
+  }
+}
+
+impl Default for WasmGeneration {
```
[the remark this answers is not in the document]
HenriqueNogara: handled on #660

bindings/wasm/src/account/storage/traits.rs
HenriqueNogara: Yes. Handled on #660

bindings/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
eike-hass, line 1: This file shouldn't exist.

bindings/wasm/examples-account/src/memory_storage.ts
```diff
@@ -0,0 +1,159 @@
...
+    public async keyGet(did: DID, keyLocation: KeyLocation): Promise<string> {
+        if (this.vaults.has(did)) {
+            let vault: Map<KeyLocation, KeyPair> = this.vaults.get(did);
+            if (vault.has(keyLocation)) {
+                let keyPair: KeyPair = vault.get(keyLocation);
+                return keyPair.public
+            }
+            throw 'Key location not found'
+        }
+        throw 'DID not found'
```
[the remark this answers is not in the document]
eike-hass: yes, @PhilippGackstatter is right. It should be `throw new Error('Key location not found')`.

bindings/wasm/examples-account/src/memory_storage.ts
```diff
@@ -0,0 +1,159 @@
...
+        let vault: Map<KeyLocation, KeyPair> = this.vaults.get(did);
+        if (!vault.has(keyLocation)) {
+            throw 'Key location not found'
+        }
+        let keyPair: KeyPair = vault.get(keyLocation);
+        if (keyLocation.method !== MethodType.Ed25519VerificationKey2018()) {
+            throw 'Unsuported Method'
+        }
+        let signature: Uint8Array = Ed25519.sign(data, keyPair.private);
+        return new Signature(keyPair.public, signature)
```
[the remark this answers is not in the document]
eike-hass: I believe exposing utility functions for crypto, de/encoding etc. through Wasm is valuable.
I expect the outputs to atomic (as in a string) right? So they can be replaced with custom implementation, but we can supply blessed algos. What we should be vary of is forcing bound Wasm structures to on users of the interface IMHO, but if we values are atomic, that shouldn't be a problem.

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -267,19 +269,20 @@ interface Storage {
...
 }"#;
 
+// from https://github.com/rustwasm/wasm-bindgen/issues/2231#issuecomment-656293288
 #[allow(dead_code)]
 pub fn downcast_js_value<T: FromWasmAbi<Abi = u32>>(js_value: JsValue, classname: &str) -> Result<T, AccountError> {
   let constructor_name: js_sys::JsString = Object::get_prototype_of(&js_value).constructor().name();
   if constructor_name == classname {
     let ptr: JsValue = Reflect::get(&js_value, &JsValue::from_str("ptr"))
-      .map_err(|e| AccountError::PromiseError(format!("Invalid value: {:?} - Get ptr error: {:?}", js_value, e)))?;
+      .map_err(|e| AccountError::JsError(format!("Invalid value: {:?} - Get ptr error: {:?}", js_value, e)))?;
```
PhilippGackstatter, line 278:
```suggestion
      .map_err(|e| AccountError::JsError(format!("unable to retrieve `ptr` property: {:?}", e)))?;
```

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -267,19 +269,20 @@ interface Storage {
...
     let ptr_u32: u32 = ptr
       .as_f64()
-      .ok_or_else(|| AccountError::PromiseError(format!("Invalid value: {:?} - Casting error", js_value)))?
+      .ok_or_else(|| AccountError::JsError(format!("Invalid value: {:?} - Casting error", js_value)))?
```
PhilippGackstatter, line 281:
```suggestion
      .ok_or_else(|| AccountError::JsError(format!("unable to read `ptr` property of `JsValue` as f64")))?
```

Sorry, I missed these errors last time. I think they could be clearer.

bindings/wasm/src/core/ed25519.rs
```diff
@@ -16,6 +16,7 @@ pub struct WasmEd25519(pub(crate) Ed25519);
 
 #[wasm_bindgen(js_class = Ed25519)]
 impl WasmEd25519 {
+  /// Signs the given message with a base58 encoded string.
```
PhilippGackstatter, line 19:
```suggestion
  /// Signs the given `message` with a base58 encoded `key`.
```

bindings/wasm/src/tangle/message_did.rs
```diff
@@ -14,15 +12,16 @@ pub struct WasmMessageId(pub(crate) MessageId);
 
 #[wasm_bindgen(js_class = MessageId)]
 impl WasmMessageId {
```
PhilippGackstatter, line 14: This type seems to be unused.

PhilippGackstatter (commented):
A few minor notes on the latest commits, will do another pass in a bit.

bindings/wasm/examples-account/src/memory_storage.ts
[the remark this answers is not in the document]
PhilippGackstatter: For crypto operations, it seems valuable to me to have different implementations and ensure they are consistent. And not exposing these types encourages users to look for other libraries. Of course, the downside is that it also puts more of a burden on users. However, with each binding, I hope we could find one library that is a valid replacement. So while I see what you mean with the exposed types having a value to users, I'm still leaning more towards not exposing them.

PhilippGackstatter (approved):
This looks good to me now. Thank you very much for addressing all the comments! This is a really solid Rust "trait export" to Wasm, imho.

I honestly may have missed a thing or two because of the split in epic branch and napi stronghold, but since this goes through more reviews in those PRs I believe that should be okay.

bindings/wasm/examples-account/src/memory_storage.ts
```diff
@@ -0,0 +1,126 @@
...
+    constructor() {
+        this._publishedGenerations = new Map();
+        this._chainStates = new Map();
+        this._states = new Map();
+        this._vaults = new Map();
+    }
+
+    public async setPassword(_encryptionKey: Uint8Array): Promise<void> {}
+
+    public async flushChanges(): Promise<void> {}
```
cycraig, line 21: We might need a comment here explaining that these functions are not necessary here but are needed for other `Storage` implementations. I.e. it's fine to have no-op functions?

Aside: still not sure if we're going to keep `setPassword`, since I'm hoping that `Stronghold` will take over password management. Basically authentication should be part of the implementation rather than the interface, ideally.

identity-account/src/error.rs
```diff
@@ -76,6 +76,12 @@ pub enum Error {
   InvalidPrivateKey(String),
   #[error("method missing fragment")]
   MethodMissingFragment,
+  #[cfg(feature = "wasm")]
+  #[error("JsValue serialization error: {0}")]
+  SerializationError(String),
+  #[cfg(feature = "wasm")]
+  #[error("javascript function threw an exception: {0}")]
+  JsError(String),
```
cycraig, line 84: Regarding errors for the `Storage` interface, I think it's a bit unwieldy to restrict implementers to the top-level `AccountError`.

Maybe we can have a sub-error enum for the `Storage` and map them to a single `AccountError` internally? This would be similar to the `Error` in the diff crate, where each variant maps to a function.

```rust
#[derive(Debug, thiserror::Error, strum::IntoStaticStr)]
pub enum Error {
  #[error("Diff Error: {0}")]
  DiffError(String),
  #[error("Merge Error: {0}")]
  MergeError(String),
  #[error("Conversion Error: {0}")]
  ConversionError(String),
}
```

We could use strings too or retain the error `source` with e.g. `Box<dyn Error>` to preserve reporting, as mentioned in #636.

This would basically eliminate the need for these feature-gated `JsError`/`SerializationError` variants since we could define errors in the Wasm bindings themselves that are bubbled up to the correct variant with no loss of context (hopefully).

This would benefit future bindings and `Storage` implementers using the core Rust library too.

cc @olivereanderson

bindings/wasm/examples-account/src/memory_storage.ts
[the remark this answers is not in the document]
cycraig: This discussion is continued in the related issue https://github.com/iotaledger/identity.rs/issues/669#issuecomment-1047842537

bindings/wasm/src/core/secret_key.rs
```diff
@@ -0,0 +1,38 @@
...
+    Ok(WasmPrivateKey(ed25519::SecretKey::from_bytes(private_key_bytes)))
+  }
+
+  /// Returns a base58 encoded string that represents the PublicKey.
+  #[wasm_bindgen(js_name = publicKey)]
+  pub fn public_key(&self) -> String {
+    let public: ed25519::PublicKey = self.0.public_key();
+    let public_key: PublicKey = public.to_bytes().to_vec().into();
+    encode_b58(&public_key)
+  }
```
cycraig, line 37: Mentioned in #669 that we should probably move to `UInt8Array` rather than Base58-BTC encoded strings but that's a larger refactor for the Wasm bindings. We should really use the same types everywhere though, so `WasmPrivateKey` in `WasmDocument` and `WasmKeyPair` too.

bindings/wasm/src/error.rs
```diff
@@ -130,3 +132,52 @@ impl From<UpdateError> for WasmError<'_> {
     }
   }
 }
+
+/// Convenience struct to convert Result<JsValue, JsValue> to an AccountResult<_, AccountError>
+pub struct JsValueResult(pub(crate) Result<JsValue>);
```
cycraig, line 137: See the previous comment on possibly refactoring `Storage` errors, it might make `JsValueResult` unnecessary since errors would be mapped to e.g. `StorageError` variants internally?

identity-account/src/types/signature.rs
```diff
@@ -1,10 +1,10 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
 use identity_core::crypto::PublicKey;
 
 /// A digital signature and associated public key.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Deserialize)]
 pub struct Signature {
   pub(crate) pkey: PublicKey,
   pub(crate) data: Vec<u8>,
```
cycraig, line 10: I think I commented about this on the other PR but do we need `Signature` to contain the `PublicKey` as well? I don't think it's used. Is it needed for future signature schemes?

Not related to the changes in this PR but it is about the `Storage` implementation.

cycraig (commented):
I assume things like `DIDLease` and `publishedGeneration` have been removed in the other PR.

Otherwise it seems good overall, few general comments.

bindings/wasm/examples-account/src/memory_storage.ts
PhilippGackstatter: Fully agree `setPassword` should be removed, but perhaps in a follow-up PR? The changes required are minimal, I think, so it could be done here as well. I'm fine either way. Passwords should be passed to storage constructors and not be required post-creation anymore, which will, for example, be the case with the upcoming stronghold refactor. Some storages might not even need a password, so it shouldn't be part of the interface.

identity-account/src/error.rs
PhilippGackstatter: There definitely needs to be a separate `StorageError`, and I really like your suggestion with `Box<dyn Error>`. Still, I think this needs a good amount of thought put into it to provide a reasonably abstract error that can accommodate not just stronghold and memory stores, which is why I would suggest doing this in a separate PR and leave the current approach as an acceptable temporary solution. I'd be happy to open an issue for it.

identity-account/src/types/signature.rs
PhilippGackstatter: No we don't need it. In fact, this is unused in the account, too. The account `Signature` type could need a refactoring to simply return either a `Vec<u8>` or a newtype wrapper around it. The `pkey` is simply [discarded in `RemoteSign`](https://github.com/iotaledger/identity.rs/blob/267574cd0a3a4edfaa7876ccdae7151d9de1e6b6/identity-account/src/crypto/remote.rs#L94). Could be part of a number of storage refactors mentioned before (redesign errors, remove `set_password`, refactor `Signature`)?

bindings/wasm/examples-account/src/memory_storage.ts
cycraig: I'm fine with any changes being added to an issue or appended to #669, as long as there's something to keep track of them.

identity-account/src/error.rs
cycraig: Sure, it can be added to #669 or its own issue. I wouldn't merge this without having a tracking issue for what's left, however.

PR #602, opened 2022-01-18
improved OneOrMany

olivereanderson:
# Description of change
This PR consists of a couple of small improvements to `OneOrMany`: 

- `OneOrMany<T>` now implements `FromIterator<T>` 
- The `push` method now behaves arguably more intuitively in the edge case of `OneOrMany::Many(vec![])`. 

To elaborate further on the latter. The following test did not pass before this PR: 
```rust 
   let mut collection = OneOrMany::Many(Vec::<u32>::new());
    collection.push(42);
    assert_eq!(collection, OneOrMany::One(42));
```
This is not a bug, but most would probably expect the `One` variant to be matched when there is exactly *one element* in the collection. Our change should not affect the behaviour in any other cases. 

## Type of change

- [ ] Documentation Fix

## Change checklist

- [] I have made corresponding changes to the documentation

### DIFF
--- a/identity-core/src/common/one_or_many.rs
+++ b/identity-core/src/common/one_or_many.rs
@@ -74,7 +74,11 @@ impl<T> OneOrMany<T> {
         Self::Many(_) => unreachable!(),
       },
       Self::Many(ref mut inner) => {
-        inner.push(value);
+        if inner.is_empty() {
+          *self = Self::One(value);
+        } else {
+          inner.push(value);
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


cycraig:
Does `OneOrMany` currently allow instantiating an empty collection by using the `OneOrMany::Many` variant directly? Does this mean we allow deserializing invalid objects, such as a `Presentation` with no `Credentials`?

olivereanderson:
> Does `OneOrMany` currently allow instantiating an empty collection by using the `OneOrMany::Many` variant directly? Does this mean we allow deserializing invalid objects, such as a `Presentation` with no `Credentials`?

Yes, it is possible to do that.

PhilippGackstatter (approved):
Looks good to me.

The name of the type implies that it has `len > 0` so perhaps it makes sense to enforce that in a future PR, if it currently allows invalid deserializations.

cycraig (approved):
Looks good to me! Thanks for the tests.

I don't think we use the `Chore` label for PRs anymore, just issues.

I think restricting `OneOrMany` to reject empty collections would be quite a large refactor since it would require bubbling-up errors and replacing `From` with `TryFrom` in quite a few places. I also think we need to keep a custom implementation rather than replacing it with `NonEmpty` for instance as the special serialization for the `One` variant is important.

PR #598, opened 2022-01-17
Remove JSON string escaping in diff messages

cycraig:
# Description of change
Removes JSON string escaping in `DiffMessage::diff` and `ServiceEndpoint` diffs. This reduces the size of `DiffMessages` on the Tangle by ~6%. This reduction has diminishing returns for larger diff messages, going to ~1% for 2KB and <1% for 3KB+, but considering we expect most diffs to be <1KB this would still have an impact.

**WARNING:** this is a breaking change and will invalidate all prior diff chain updates. Advise developers using `0.5.0-dev.1`/`0.5.0-dev.2` to publish an integration update with all diffs applied to prevent unexpected changes.

### Advantages
- Reduces the size of Tangle messages slightly, reducing data transmission and long-term storage costs.
- No unescaping pre-processing required in the identity-resolver on the explorer website.

### Disadvantages
- Requires deserializing the `IotaDiffDocument` up-front for _every_ `DiffMessage` processed when resolving a DID (rather than only after the signature is validated). This could slow down resolution when there are many conflicting or spam messages on the diff chain but the effect is capped since Tangle message sizes have a reasonable upper-limit. Benchmarking (see section below after the example) shows this effect is significant but only in malicious cases since deserialization is always performed anyway for valid `DiffMessages`. Would not be an issue with planned spam solutions.

### Example:

A medium-sized `DiffMessage` which adds a service and a verification method.

| # Bytes           | JSON | Brotli+JSON |
|---------------------|------|-------------|
| Escaped (current)   | 967  | 562         |
| Unescaped (this PR) | 898  | 526         |
| **Change**              | -69  | -36         |
| **Percent Change**      | 7.1% | 6.4%        |

1. Current: stores an escaped JSON string as the `diff` (formatted for display).

```json
{
  "id": "did:iota:3ZW4YC6jhPBzBE5devGRnPbrhkJJWyxT1mXRRp11hLVX",
[...]

PR #605, opened 2022-01-20
Fix dependent diff updates being rejected

### LINKED ISSUE #593: [Bug] DiffChain resolution discards certain valid updates as spam
## Bug description

There is an issue in the diff chain resolution code. The following example reproduces the issue (thanks @abdulmth for providing the initial example):

```rust
use std::path::PathBuf;

use identity::account::Account;
use identity::account::AccountStorage;
use identity::account::IdentitySetup;
use identity::account::Result;
use identity::core::Url;

#[tokio::main]
  let password: String = "my-password".into();

  let mut account: Account = Account::builder()
    .storage(AccountStorage::Stronghold(stronghold_path, Some(password), None))
    .create_identity(IdentitySetup::default().network("dev").unwrap())
    .await?;

  account
    .update_identity()
    .create_service()
    .fragment("my-service-1")
    .type_("MyCustomService")
    .endpoint(Url::parse("https://example.com")?)
    .apply()
    .await?;

  account
    .update_identity()
    .delete_service()
    .fragment("my-service-1")
    .apply()
    .await?;

  println!(
    "resolved document: {:#?}",
    account.resolve_identity().await.unwrap()
  );

  Ok(())
```

## Language version

- Language: Rust
- Version:1.56.1

## IOTA Identity version

- Version (version number, commit, or branch): Latest current dev `cc0089f2c2ec6b639c34527de74d01df8dded78e`

## Expected behaviour

Print a document which does not contain any services.

## Actual behaviour

Prints a document which still contains the `my-service-1` that was added, even though we deleted the service and would expect the services to be empty.

## Fix

The problem in the resolution is most likely here:

https://github.com/iotaledger/identity.rs/blob/cc0089f2c2ec6b639c34527de74d01df8dded78e/identity-iota/src/chain/diff_chain.rs#L84-L91

In the above example, the two service updates are published on the same diff chain. When resolving, it is checked whether the service-remove update is a valid addition. This is done agains the original `integration_document` rather than the one that contains the service and so it fails with `DiffError(MergeError("Unable to pop value"))` - the vector of services is still empty.

A fix would be to verify the signature on the diff message and everything that `check_valid_addition` currently does, except for the `merge`. (Signature verification as an initial filtering step might also be enough). Then sorting by milestone, trying to merge and taking the result of the first successful `merge` as the `integration_document` for the next iteration. (Paraphrasing @cycraig here, thanks!).

cycraig:
# Description of change

Fixes the `DiffChain` rejecting `DiffMessages` that update document sections added in previous diffs. From the example given in #593: adding a service in a diff update then editing or removing it in another diff update will result in the latter update being rejected/discarded by the `DiffChain` during resolution.

Thank you to @PhilippGackstatter and @abdulmth for identifying and investigating the issue!

This also flattens the properties of `DocumentMetadata` serialization (this ensures unhandled fields are captured during deserialization), which may lead to breaking changes if any existing user-defined custom fields conflict with hardcoded field names.

**WARNING:** this may result in resolutions of IOTA DID Documents that contain diff updates being different after this is merged.

Fixes #593 

## Type of change

- [ ] Documentation Fix

Existing and new tests pass, including examples.

## Change checklist

### DIFF
--- a/identity-iota/src/chain/diff_chain.rs
+++ b/identity-iota/src/chain/diff_chain.rs
@@ -1,22 +1,20 @@
-// Copyright 2020-2021 IOTA Stiftung
+// Copyright 2020-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use core::fmt::Display;
 use core::fmt::Formatter;
-
 use core::slice::Iter;
 
 use serde;
 use serde::Deserialize;
 use serde::Serialize;
 
-use crate::chain::IntegrationChain;
 use identity_core::convert::FmtJson;
 
 use crate::chain::milestone::sort_by_milestone;
+use crate::chain::IntegrationChain;
 use crate::did::IotaDID;
 use crate::diff::DiffMessage;
-use crate::document::IotaDocument;
 use crate::document::ResolvedIotaDocument;
 use crate::error::Error;
 use crate::error::Result;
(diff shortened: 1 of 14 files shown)


identity-iota/src/chain/diff_chain.rs
```diff
@@ -76,23 +74,40 @@ impl DiffChain {
...
+          // signing verification methods and reject invalid changes to non-existent sections.
+          match Self::try_merge(&diff, &current_document) {
+            Ok(merged_document) => Some((diff, merged_document)),
+            _ => None,
+          })
+        .next()
+      {
+        // Update the document for the next diff to allow updating sections added by previous diffs.
+        current_document = merged_document;
+        this.push_unchecked(diff); // checked by check_valid_addition and try_merge above
```
PhilippGackstatter, line 110:
```suggestion
        // checked by verify_diff and try_merge above
        this.push_unchecked(diff);
```

PhilippGackstatter (commented):
Unfortunately, I can't confirm it's working on my side. The example from #593 is printing the expected result, but that seems to be a coincidence. No matter how many diffs I publish in `try_from_index_with_document`,

```rust
index.get(
      this
        .current_message_id()
        .unwrap_or_else(|| current_document.message_id()),
    ).unwrap().len()
```

 is always `1`.

For instance, if I run an example that adds `service-1`, deletes `service-1` and adds `service-2`, the `service` property is still empty when being resolved, even though I confirmed three diffs are being published.

cycraig:
Thanks for catching that! Turns out there's an entirely unrelated bug to what this PR fixes and I tested insufficiently (multiple times).

All `DiffMessages` fail validation after deserialization now because of the change in #598. It appears that updates encoded in `DiffIotaDocument` change after (de)serialization to JSON, which means the signature validation fails since when it's serialized a second time the JSON changes. I'll open another PR to address this...

cycraig:
Opened #611 to fix the deserialisation bug, will re-evaluate this PR once that's merged.

cycraig:
Seems to be fixed, I manually checked the examples this time!

identity-iota/src/document/iota_document.rs
```diff
@@ -311,6 +311,7 @@ impl IotaDocument {
   }
 
   /// Remove a [`Service`] identified by the given [`IotaDIDUrl`] from the document.
+  // TODO: return an error or bool if no service was removed?
```
PhilippGackstatter, line 314: I agree, then we can remove this:

https://github.com/iotaledger/identity.rs/blob/630a61e9c9aa25273b3cf5718844b3e563bd28c3/identity-account/src/updates/update.rs#L264-L267

Doesn't need to be in this PR.

PhilippGackstatter (approved):
Looks good to me, can confirm my example is also working now :ok_hand:! Thank you very much for fixing this!

HenriqueNogara (approved):

PR #606, opened 2022-01-20
Remove `DocumentVerifier`

olivereanderson:
# Description of change
Removes `DocumentVerifier` and moves its verification methods `verify`, `do_verify` to  `CoreDocument` essentially as: 
```
pub fn verify_data<X>(&self, data: &X, options: &VerifierOptions) -> Result<()>;
pub fn do_verify<X>(method: &VerificationMethod<U>, data: &X) -> Result<()>; 
```

Additionally `IotaDocument::verify_data` now borrows  `VerifierOptions` instead of consuming it. More precisely the new function signature is now: 
```
pub fn verify_data<X>(&self, data: &X, options: &VerifierOptions) -> Result<()>; 
```
in place of 
```
pub fn verify_data<X>(&self, data: &X, options: VerifierOptions) -> Result<()>; 
```
This is arguably more conceptual as it should be enough to "read" the verifier options in order to verify data and it allows for optimizations such as mutating the options in place between verifications. 

## Type of change

- [ ] Documentation Fix

`cargo test` and `npm run build && npm run build:examples && npm run test:node && npm run test:browser` run locally. 

## Change checklist


identity-did/src/verifiable/document_verifier.rs
```diff
@@ -36,54 +38,87 @@ use crate::Result;
...
-    self.options = self.options.method_scope(method_scope);
+  pub fn method_scope(self, method_scope: MethodScope) -> DocumentVerifier<'base, 'static, T, U, V> {
+    /*self.options = self.options.method_scope(method_scope);
     self
+    */
+    DocumentVerifier {
+      document: self.document,
+      options: Cow::Owned(self.options.into_owned().method_scope(method_scope)),
+    }
   }
```
cycraig, line 84: Maybe instead of all these setters on `DocumentVerifier` we can just leave it as an immutable borrow field and allow overwriting that borrow as a whole with `DocumentVerifier::set_options`? I don't imagine many developers will use `DocumentVerifier` over `IotaDocument::verify_data`.

The conversion of `Cow` to owned is unexpected in my opinion.

olivereanderson: Thanks for your early feedback :+1:  

> The conversion of Cow to owned is unexpected in my opinion.

The fact that it must convert the `Cow`to owned is somewhat reflected in the function signature (the return type has a `'static` lifetime in the second generic parameter). To make this even more obvious we could introduce a sealed trait 
`trait OptionsState` that is implemented for two zero sized types `Borrowed` and `Owned`. Then `DocumentVerifier` is changed to  
```
DocumentVerifier<'base, 'options, State: VerifierOptionState, T = Object, U = Object, V = Object,>
```
and then there is an implementation 
```
impl DocumentVerifier<'base, 'options, Borrowed, T, U, V> {
     fn into_owned(self) -> DocumentVerifier<'base, 'static, Owned, T, U, V>; 
}
```
and the methods that change the `options` are only implemented on `DocumentVerifier<'base, 'static, Owned, T, U, V>`. This would make the fact that one needs to clone if one wants to manipulate the verifier options directly on `DocumentVerifier` very obvious. The question is whether it is worth the extra code? Otherwise it might be OK to not let `DocumentVerifier` mutate the individual fields of its verifier options as you mentioned.

HenriqueNogara (approved):

identity-did/src/verifiable/document_verifier.rs
olivereanderson: Decided to refactor `DocumentVerifier` to take the verifier options in the `verify` method instead of having them as a field in the struct.

cycraig (commented):
Thanks for the changes. I'm still hesitant to approve because while the previous approach had the disadvantage of unexpected `Cow`/cloning behaviour internally, removing the mutability of the options from `DocumentVerifier` essentially makes it useless (since its purpose was mainly ergonomics using the builder pattern), unlike `DocumentSigner` which still retains some options separate from `SigantureOptions`. 

If we were to proceed with this PR I would advocate for removing `DocumentVerifier` completely and replace it with a `Verifier`/`SignatureVerifier` trait implemented on `CoreDocument`, `IotaDocument`, and maybe `Account`.

Apologies for being difficult and I know I'm the one that requested the latest changes to your initial approach; I just want to minimise breaking API changes, since it looks like if this is merged as-is we might want/need to change it again later.

PhilippGackstatter (commented):
I really like the simplification of passing the options as a parameter as opposed to adding a second lifetime to `DocumentVerifier` and using `Cow`.

olivereanderson:
> I would advocate we remove `DocumentVerifier` completely and replace it with a `Verifier`/`SignatureVerifier` trait implemented on `CoreDocument`, `IotaDocument`, and maybe `Account`.

Interesting idea. How would such a trait fir with traits we already have such as: https://github.com/iotaledger/identity.rs/blob/dev/identity-core/src/crypto/signature/traits/core.rs#L82? 

Also which functions/methods would/might impose the new `Verifier/Signature` verifier trait as trait bounds?

cycraig:
> How would such a trait fit with the traits we already have such as: https://github.com/iotaledger/identity.rs/blob/dev/identity-core/src/crypto/signature/traits/core.rs#L82?
> Also which functions/methods would/might impose the new Verifier/SignatureVerifier trait as trait bounds?

I imagine the `DocumentVerifier` trait to be user-facing, so it just exposes e.g. `verify_data`/`verify_document` since we don't have internal uses of the `DocumentVerifier` struct other than in tests and examples as far as I know. The other option would be just to leave them as associated methods on `CoreDocument`, `IotaDocument` and `Account`. The reason I initially suggested a trait was in case a developer wanted to take e.g. `dyn DocumentVerifier` somewhere but in reality I think they would just pass the document around or an `Arc` of the `Account` for instance, so it might not be useful.

I'm in favour of just leaving it as an associated method then, any other options or opinions? Leaving everything as-is is also an option.

PhilippGackstatter:
What about moving the method and having it as `CoreDocument::verify_data`, then also exposing it as `IotaDocument::verify_data`? In theory the trait sounds nice, but if the only advantage of it is to have `account.verify_data(...)` instead of `account.document().verify_data(...)` that seems like too small of a reason to introduce a trait. Unless we expect to also implement it on more types in the future?

cycraig:
> [...] that seems like too small of a reason to introduce a trait. Unless we expect to also implement it on more types in the future?

A trait might be useful if we need interoperability between future DID method documents, but that assumes they will be implemented in different structs.

> What about moving the method and having it as CoreDocument::verify_data, then also exposing it as IotaDocument::verify_data [...]

It looks like we're all leaning towards this option.

identity-iota/src/document/iota_document.rs
```diff
@@ -551,12 +543,10 @@ impl IotaDocument {
...
   pub fn verify_document(signed: &IotaDocument, signer: &IotaDocument) -> Result<()> {
     // Ensure signing method is allowed to sign document updates.
-    signer
-      .verifier()
+    let options = VerifierOptions::default()
       .method_scope(MethodScope::capability_invocation())
-      .method_type(Self::UPDATE_METHOD_TYPES.to_vec())
-      .verify(signed)
-      .map_err(Into::into)
+      .method_type(Self::UPDATE_METHOD_TYPES.to_vec());
```
cycraig, line 548: Aside: it would be nice to have these options static. I tried making the functions in `VerifierOptions` const but ran into the lack of the `const_precise_live_drops` feature in stable to handle dropping the `MethodType` `Vec`. Other option is `lazy_static` but probably not worth it in this instance.

cycraig (approved):
Looks good to me, thanks for all the changes. I like that `WasmDocument::verify_data` borrows `WasmVerifierOptions` now, it's always going to be confusing to Javscript developers when our API consumes an object and nulls its pointer.

HenriqueNogara (approved):

identity-did/src/document/core_document.rs
```diff
@@ -565,6 +586,125 @@ impl<T, U, V> CoreDocument<T, U, V> {
...
+
+  /// Verifies the signature of the provided data matches the public key data from the given
+  /// verification method.
+  ///
+  /// # Errors
+  ///
+  /// Fails if an unsupported verification method is used, data
+  /// serialization fails, or the verification operation fails.
+  pub fn do_verify<X>(method: &VerificationMethod<U>, data: &X) -> Result<()>
+  where
```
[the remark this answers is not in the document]
olivereanderson: I made it private for now and will open up an issue on where and how we make this functionality public. In this way we can make it public in a non-breaking way in a later PR.

identity-iota/src/document/iota_document.rs
olivereanderson: Feel free to create an issue for this and we can look into it later, when/if `const_precise_live_drops` lands.

PR #611, opened 2022-01-21
Fix diff properties (de)serialization

cycraig:
# Description of change
Fixes a bug introduced in #598 where `DiffMessages` from the Tangle no longer match the expected signature due to serialization differences, causing verification in `DiffChain` to fail and any diffs that add a `Service` or `VerificationMethod` to be rejected.
Before #598, the `DiffIotaDocument` was serialised to a JSON string first for the `DiffMessage::diff` field. This meant the signature verification of `DiffMessage` was not affected, since `DiffIotaDocument` was not deserialised then serialised again during verification.

The root cause is due to how `properties` in `Service` and `VerificationMethod` are handled in their `ServiceDiff::into_diff` and `DiffMethod::into_diff` implementations: empty `properties` (of type `Object` i.e. `BTreeMap`) would be diffed to `Some(DiffHashMap([]))`. Internally `DiffHashMap([])` (which is `None`) skips serialization, but wrapping it in `Some` means it serializes to `"properties": null` instead. Deserializing `"properties": null` results in `None` instead of `Some(DiffHashMap([]))`, so the JSON serialization changes after deserialization, which means the signature verification (which internally serializes objects to canonical JSON) fails. 

This scenario only occurs for types with this bug when wrapped in a collection (like `OrderedSet`s of `Service`s and `VerificationMethod`s in `CoreDocument`) with `DiffVec`, and then only for additions (which use `Diff::into_diff` and `Diff::from_diff`) since other changes like editing or removing a collection entry do not use `Diff:into_diff`.

It would be worth investigating if it's possible to remove `Diff::into_diff` and `Diff::from_diff` by refactoring `DiffVec` (specifically `InnerVec`) to store the actual type `T` instead of converting between its `<T as Diff>::Type`. However: I'm unsure if that's possible, how long it would take, or what other bugs that would introduce.

Multiple unit tests have been introduced to prevent regressions but we do not have coverage for every possible diff of every field.

### Example

Serialising a `DiffService` with a default/empty `properties` field to JSON, then serializing its deserialisation.

#### 1. Before (current) serialisation changes.

Initial serialisation:
```json
{"id":"did:example:1234","type_":"test_service","service_endpoint":"did:service:1234","properties":null}
```
Serialisation after deserialisation:
```json
{"id":"did:example:1234","type_":"test_service","service_endpoint":"did:service:1234"}
```

#### 2. After (this PR) serialisation is consistent.

Initial serialisation:
```json
{"id":"did:example:1234","type_":"test_service","service_endpoint":"did:service:1234"}
```
Serialisation after deserialisation:
```json
{"id":"did:example:1234","type_":"test_service","service_endpoint":"did:service:1234"}
```

Aside: this also removes some unnecessary clones from `Diff::into_diff` implementations.

NOTE: the fix in #605 is still required for diffs to work as expected. This PR fixes adding structs in diffs, #605 fixes editing or removing structs added in previous diffs.

## Type of change

- [ ] Documentation Fix

Examples were hand-checked to work as-expected (but still suffer from the bug to be fixed by #605). New unit and integration tests were added.

All tests and examples pass locally.

## Change checklist

### DIFF
--- a/bindings/wasm/examples/src/diff_chain.js
+++ b/bindings/wasm/examples/src/diff_chain.js
@@ -1,7 +1,7 @@
 // Copyright 2020-2021 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import {Client, Config, Document, Service} from '@iota/identity-wasm';
+import {Client, Config, Document, Service, Timestamp} from '@iota/identity-wasm';
 import {createIdentity} from "./create_did";
 import {logExplorerUrl, logResolverUrl} from "./utils";
 
@@ -29,11 +29,12 @@ async function createDiff(clientConfig) {
 
     // Add a Service
     let serviceJSON = {
-        id: doc.id + "#new-linked-domain",
+        id: doc.id + "#linked-domain-1",
         type: "LinkedDomains",
         serviceEndpoint: "https://identity.iota.org",
     };
     updatedDoc.insertService(Service.fromJSON(serviceJSON));
+    updatedDoc.metadataUpdated = Timestamp.nowUTC();
     console.log(updatedDoc);
 
     // Create diff
--- a/documentation/docs/specs/did/iota_did_method_spec.md
+++ b/documentation/docs/specs/did/iota_did_method_spec.md
@@ -121,23 +121,23 @@ Example of an Integration DID Message:
 ```json
 {
   "doc": {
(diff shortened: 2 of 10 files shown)


olivereanderson (approved):
Looks good to me!

PhilippGackstatter (approved):
Thanks for fixing it so quickly, looks good!

PR #616, opened 2022-01-24
Add union type parameters

cycraig:
# Description of change

This adds the `UMethodQuery` and `UDID` (naming convention open to change) Typescript union types. This pattern allows us to create analogues for generic types in function parameters, such as `MethodQuery` which takes either an `IotaDIDUrl` or `String`. This is to enhance the developer-friendliness of our Wasm API by removing unnecessary `toString` invocations and allowing pseudo-generic, polymorphic function parameters.

~~`export type UMethodQuery = DIDUrl | string;`~~
~~`export type UDID = DID | string;`~~

Edit: this PR no longer exports the union types to Typescript, instead it injects the union into the function parameter definitions directly.

E.g.
```ts
signSelf(key_pair: KeyPair, method_query: DIDUrl | string): void;
```

With this change, all of the following work:
```ts
doc.signSelf("did:iota:example#sign-0");                // CURRENT
doc.signSelf(doc.defaultSigningMethod().id.toString()); // CURRENT
doc.signSelf(doc.defaultSigningMethod().id);            // NEW

doc.resolveMethod("did:iota:example#sign-0");                // CURRENT
doc.resolveMethod(doc.defaultSigningMethod().id.toString()); // CURRENT
doc.resolveMethod(doc.defaultSigningMethod().id);            // NEW

let resolved = await client.resolve("did:iota:example"); // CURRENT
let resolved = await client.resolve(doc.id.toString());  // CURRENT
let resolved = await client.resolve(doc.id);             // NEW

let history = await client.resolveHistory("did:iota:example"); // CURRENT
let history = await client.resolveHistory(doc.id.toString());  // CURRENT
let history = await client.resolveHistory(doc.id);             // NEW
```

The alternative for `Client::resolve` and `Client::resolveHistory` would be to enforce taking in a `WasmDID` only, but error handling is more unwieldy than in Rust and users may extract DIDs as strings from JSON objects like credentials. The proposed pattern would still be useful for `UMethodQuery` which matches the behaviour in Rust. Note this is only possible when all types in the union serialise to the same representation, since we use `serde` to handle the conversion. Different union types are possible but would require an intermediary untagged enum to handle deserialisation with this method.

### Added

- `UDID` union type to take either `DID` or a `string` as a parameter.
- `UMethodQuery` union type to take either `DIDUrl` or a `string` as a parameter.
- `WasmNetwork::name` getter.

### Changed

- `Document::resolveMethod` parameter changed from `string` to `UMethodQuery`.
- `Document::revokeMerkleKey` parameter changed from `string` to `UMethodQuery`.
- `Document::signSelf` parameter changed from `string` to `UMethodQuery`.
- `Document::diff` parameter changed from `string` to `UMethodQuery`.
- `Client::resolve` parameter changed from `string` to `UDID`.
- `Client::resolveHistory` parameter changed from `string` to `UDID`.
- `WasmDID::network` is no longer a getter, just a regular function.
- `WasmDID` JSON representation changed to match `IotaDID`.
- `WasmDIDUrl` JSON representation changed to match `IotaDID`.

## Type of change

- [ ] Documentation Fix

Wasm tests and examples pass locally.

## Change checklist

### DIFF
--- a/bindings/wasm/docs/api-reference.md
+++ b/bindings/wasm/docs/api-reference.md
@@ -180,7 +180,7 @@ Fetch the DID document specified by the given `DID`.
 
 | Param | Type |
 | --- | --- |
-| did | <code>UDID</code> | 
+| did | [<code>DID</code>](#DID) \| <code>string</code> | 
 
 <a name="Client+resolveHistory"></a>
 
@@ -191,7 +191,7 @@ Returns the message history of the given DID.
 
 | Param | Type |
 | --- | --- |
-| did | <code>UDID</code> | 
+| did | [<code>DID</code>](#DID) \| <code>string</code> | 
 
 <a name="Client+resolveDiffHistory"></a>
 
@@ -1059,7 +1059,7 @@ Throws an error if the method is not found.
 
 | Param | Type |
 | --- | --- |
-| query | <code>UMethodQuery</code> | 
+| query | [<code>DIDUrl</code>](#DIDUrl) \| <code>string</code> | 
 
 <a name="Document+revokeMerkleKey"></a>
 
@@ -1068,7 +1068,7 @@ Throws an error if the method is not found.
 
 | Param | Type |
 | --- | --- |
-| query | <code>UMethodQuery</code> | 
+| query | [<code>DIDUrl</code>](#DIDUrl) \| <code>string</code> | 
 | index | <code>number</code> | 
 
 <a name="Document+signSelf"></a>
@@ -1086,7 +1086,7 @@ verification method. See `Document::verifySelfSigned`.
 | Param | Type |
 | --- | --- |
 | key_pair | [<code>KeyPair</code>](#KeyPair) | 
-| method_query | <code>UMethodQuery</code> | 
+| method_query | [<code>DIDUrl</code>](#DIDUrl) \| <code>string</code> | 
 
 <a name="Document+signCredential"></a>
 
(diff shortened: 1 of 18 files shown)


olivereanderson (approved):
Looks good to me!

abdulmth:
The only problem I see with this change is that the type in the method signature became harder to understand.
for example:
```javascript 
signSelf(key_pair: KeyPair, method_query: UMethodQuery): void
```
Developers must go to the method declaration to see what `UMethodQuery` actually includes, and that's a string can be passed as an argument.
In a perfect world, the method should look like: 
```javascript 
signSelf(key_pair: KeyPair, method_query: DIDUrl | string): void
```
I'm not sure removing `toString()` is worth the complexity the new method signature has. But also having polymorphic arguments is an improvement.

cycraig:
> The only problem I see with this change is that the type in the method signature became harder to understand <...>
>
> ```js
> signSelf(key_pair: KeyPair, method_query: DIDUrl | string): void
> ```

Turns out `wasm-bindgen` does actually support generating such a function signature via `typescript_type`! https://github.com/rustwasm/wasm-bindgen/issues/2155

I forgot about the above since I was just looking at their guide: https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/typescript_type.html

```rust
/// Duck-typed union to pass either a string or WasmDID as a parameter.
#[wasm_bindgen]
extern "C" {
  #[wasm_bindgen(typescript_type = "DID | string")]
  pub type UWasmDID;
}

/// Duck-typed union to pass either a string or WasmDIDUrl as a parameter.
#[wasm_bindgen]
extern "C" {
  #[wasm_bindgen(typescript_type = "DIDUrl | string")]
  pub type UWasmMethodQuery;
}
```

With the revised code the Typescript function signature is as-expected:

```ts
* @param {KeyPair} key_pair
* @param {DIDUrl | string} method_query
*/
  signSelf(key_pair: KeyPair, method_query: DIDUrl | string): void;
```

Thanks for the catch!

eike-hass (approved):
big improvement in my opinion

PR #624, opened 2022-01-25
Disallow (de)serializing empty `OneOrMany`

olivereanderson:
# Description of change
This PR ensures that an attempt to deserialize the string "[]" to an empty `OneOrMany` collection fails. Similarly serializing such empty collections is now also not possible.  This change makes it harder to produce non-spec compliant verifiable credentials. 

Note that in order to serialize structs that allow empty `OneOrMany` fields one will have to apply the  `#[serde(default,  skip_serializing_if = "OneOrMany::is_empty")]` attribute on those fields after this PR. 

## Type of change

- [ ] Documentation Fix

runs locally under `cargo test` and `npm run build && npm run build:examples && npm run test:node && npm run test:browser`

## Change checklist

### DIFF
--- a/identity-core/src/common/one_or_many.rs
+++ b/identity-core/src/common/one_or_many.rs
@@ -10,14 +10,56 @@ use core::mem::replace;
 use core::ops::Deref;
 use core::slice::from_ref;
 
+use serde::de::Deserializer;
+use serde::Deserialize;
+use serde::Serialize;
+use serde::Serializer;
+
 /// A generic container that stores exactly one or many (0+) values of a given type.
+///
+/// # Serialization
+/// An attempt to serialize an empty `OneOrMany` collection will fail. Similarly one cannot construct an empty
+/// collection by means of deserialization.
 #[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
 #[serde(untagged)]
 pub enum OneOrMany<T> {
   /// A single instance of `T`.
   One(T),
   /// Multiple (zero or more) instances of `T`.
-  Many(Vec<T>),
+  Many(
+    #[serde(deserialize_with = "non_empty_vec_deserializer")]
+    #[serde(serialize_with = "non_empty_vec_serializer")]
+    Vec<T>,
+  ),
+}
+
+fn non_empty_vec_deserializer<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
+where
+  D: Deserializer<'de>,
+  T: Deserialize<'de>,
+{
+  let v = Vec::<T>::deserialize(deserializer)?;
+  if v.is_empty() {
+    Err(serde::de::Error::custom(
+      "deserialization failed: an empty OneOrMany collection cannot be deserialized",
+    ))
+  } else {
+    Ok(v)
+  }
+}
+
+fn non_empty_vec_serializer<T, S>(v: &[T], serializer: S) -> std::result::Result<S::Ok, S::Error>
+where
+  S: Serializer,
+  T: Serialize,
+{
+  if v.is_empty() {
+    Err(serde::ser::Error::custom(
+      "serialization failed: an empty OneOrMany collection cannot be serialized",
+    ))
+  } else {
+    v.serialize(serializer)
+  }
 }
 
 impl<T> OneOrMany<T> {
@@ -216,6 +258,8 @@ impl<'a, T> Iterator for OneOrManyIter<'a, T> {
 
 #[cfg(test)]
 mod tests {
+  use crate::convert::ToJson;
+
   use super::*;
 
   #[test]
@@ -268,5 +312,30 @@ mod tests {
     let mut collection = OneOrMany::Many(v);
     collection.push(42);
     assert_eq!(collection, OneOrMany::Many((0..=42).collect()));
+  }
+
+  #[test]
+  fn serialize_empty_collection_fails() {
+    let empty_collection = OneOrMany::Many(Vec::<u32>::new());
+    assert!(empty_collection.to_json().is_err());
+  }
+
+  #[test]
+  fn deserialize_empty_collection_fails() {
+    let value_str = "[]";
+    let deserialized: Result<OneOrMany<u32>, serde_json::Error> = serde_json::from_str(value_str);
+    assert!(deserialized.is_err());
+  }
+
+  #[test]
+  fn valid_serialize_deserialize_roundtrips() {
+    let one = OneOrMany::One(1);
+    let many = OneOrMany::Many(vec![1, 2, 3]);
+    let one_serialized = one.to_json().unwrap();
+    let many_serialized = many.to_json().unwrap();
+    let one_deserialized: OneOrMany<i32> = serde_json::from_str(&one_serialized).unwrap();
+    let many_deserialized: OneOrMany<i32> = serde_json::from_str(&many_serialized).unwrap();
+    assert_eq!(one, one_deserialized);
+    assert_eq!(many, many_deserialized);
   }
 }


identity-core/src/common/one_or_many.rs
```diff
@@ -10,14 +10,56 @@ use core::mem::replace;
...
 pub enum OneOrMany<T> {
   /// A single instance of `T`.
   One(T),
   /// Multiple (zero or more) instances of `T`.
-  Many(Vec<T>),
+  Many(
+    #[serde(deserialize_with = "non_empty_vec_deserializer")]
+    #[serde(serialize_with = "non_empty_vec_serializer")]
+    Vec<T>,
+  ),
```
cycraig, line 33:
```suggestion
  Many(
    #[serde(serialize_with = "non_empty_vec_serializer", deserialize_with = "non_empty_vec_deserializer")]
    Vec<T>,
  ),
```

identity-core/src/common/one_or_many.rs
```diff
@@ -10,14 +10,56 @@ use core::mem::replace;
...
+  if v.is_empty() {
+    Err(serde::de::Error::custom(
+      "deserialization failed: an empty OneOrMany collection cannot be deserialized",
+    ))
+  } else {
+    Ok(v)
+  }
+}
+
+fn non_empty_vec_serializer<T, S>(v: &[T], serializer: S) -> std::result::Result<S::Ok, S::Error>
```
cycraig, line 51: Opinion: maybe `many_serializer` would be more concise?

cycraig (changes requested):
This introduces an invariant on `OneOrMany` that is only enforced when (de)serialized, which doesn't match the type's behaviour since one can easily create an empty `OneOrMany::Many` variant.

This also breaks valid `Credential` and `Presentation` instances since many of the fields on those structs are optional but typed as `OneOrMany`. That is, they should either be omitted or non-empty if present on the struct.

E.g. `Credential`:

```rust
  /// Information used to determine the current status of the `Credential`.
  #[serde(default, rename = "credentialStatus", skip_serializing_if = "OneOrMany::is_empty")]
  pub credential_status: OneOrMany<Status>,
  /// Information used to assist in the enforcement of a specific `Credential` structure.
  #[serde(default, rename = "credentialSchema", skip_serializing_if = "OneOrMany::is_empty")]
  pub credential_schema: OneOrMany<Schema>,
  /// Service(s) used to refresh an expired `Credential`.
  #[serde(default, rename = "refreshService", skip_serializing_if = "OneOrMany::is_empty")]
  pub refresh_service: OneOrMany<Refresh>,
  /// Terms-of-use specified by the `Credential` issuer.
  #[serde(default, rename = "termsOfUse", skip_serializing_if = "OneOrMany::is_empty")]
  pub terms_of_use: OneOrMany<Policy>,
  /// Human-readable evidence used to support the claims within the `Credential`.
  #[serde(default, skip_serializing_if = "OneOrMany::is_empty")]
  pub evidence: OneOrMany<Evidence>,
```

E.g. `Presentation`:

```rust
  /// Service(s) used to refresh an expired `Presentation`.
  #[serde(default, rename = "refreshService", skip_serializing_if = "OneOrMany::is_empty")]
  pub refresh_service: OneOrMany<Refresh>,
  /// Terms-of-use specified by the `Presentation` holder.
  #[serde(default, rename = "termsOfUse", skip_serializing_if = "OneOrMany::is_empty")]
  pub terms_of_use: OneOrMany<Policy>,
```

Ideally `OneOrMany` itself should be changed to prevent creating an empty `Many` variant. This means changing `From` implementations to `TryFrom` etc.

Edit: there are other usages of `OneOrMany` (like in `Policy`) that need to be checked carefully against the specification to ensure we're not violating it with this change.

olivereanderson:
> This also breaks valid `Credential` and `Presentation` instances since many of the fields on those structs are optional but typed as `OneOrMany`

As far as I can tell it does not break valid `Credential` instances as this test still passes: https://github.com/iotaledger/identity.rs/blob/dev/identity-credential/src/credential/credential.rs#L220. As mentioned in the PR description this is due to the usage of these annotations #[serde(default, skip_serializing_if = "OneOrMany::is_empty")] in the definition of `Credential`.  

In other words it is possible to construct the empty variant via deserialization, but it requires an explicit annotation. Perhaps I didn't formulate this well enough in the PR title/description. 

> Ideally OneOrMany itself should be changed to prevent creating an empty Many variant. This means changing From implementations to TryFrom etc.

I am not really sure that this is desirable. It can be nice to initialise an empty `OneOrMany` and then let it grow during runtime. If it was wrapped in an option this would quickly get less ergonomic.

cycraig:
> Perhaps I didn't formulate this well enough in the PR title/description.

Nope, completely my fault for skimming the description, sorry.

> I am not really sure that this is desirable. It can be nice to initialise an empty OneOrMany and then let it grow during runtime.

My concern is that enforcing the non-empty rule only on (de)serialization is that it makes errors harder to catch and debug in Rust by moving where they occurrence. The errors returned by `serde` aren't perfect either since they don't identify the exact cause (the custom error message is not used) or field that caused it. 

E.g. deserializing an empty `OneOrMany` - `type` field on `Credential`.
```json
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "id": "http://example.edu/credentials/58473",
  "type": [],
   ...
}
```
throws:
```rust
DecodeJSON(Error("data did not match any variant of untagged enum OneOrMany", line: 7, column: 12))
```

E.g. serializing an empty `OneOrMany` `type` field on `Credential` 
```rust
credential.types = OneOrMany::Many(vec![]);
credential.to_json().unwrap();
```
throws:
```rust
EncodeJSON(Error("serialization failed: an empty OneOrMany collection cannot be serialized", line: 0, column: 0))
```

So ideally we would catch this before I even get to trying to serialize the struct, by throwing an error when trying to instantiate `OneOrMany` with an empty vector. It's not like we can improve the error messages on deserialization as far as I know so that's a non-issue, it's about when and where errors are thrown when creating a credential improperly in Rust.

> If it was wrapped in an option this would quickly get less ergonomic

The `CredentialBuilder` is meant to be easy-to-use to instantiate such fields, and in Wasm we use json directly so it's not too bad. Options are pretty common when handling real-world data structs in Rust, I don't think dealing with them is too bad but that's just my experience.
```
  pub(crate) status: Vec<Status>,
  pub(crate) schema: Vec<Schema>,
  pub(crate) refresh: Vec<Refresh>,
  pub(crate) policy: Vec<Policy>,
  pub(crate) evidence: Vec<Evidence>,
```

Don't get me wrong, overall this PR is still a good change since it will prevent creating structurally improper credentials up-front, particularly for Wasm where `json` is used in the examples so `serde` is the only check. (Although we should check `Credential.extend` acts as expected and doesn't suddenly break with this).

I'm aware that changing `OneOrMany` to be fallible touches a lot of existing code and types so I'm open to persuasion, I still think restricting the `OneOrMany` type too would be better.

olivereanderson:
I agree that `AtLeastOne`, and `Option<AtLeastOne>` is more type safe than `OneOrMany`, but at the same time (as you mentioned) that refactor would make this PR go from "patch" to a breaking change. This is something we can do at this point, but there are many other pending tasks that arguably have a higher priority. 

This PR makes it harder to produce non-spec compliant `Credentials/Presentations` by deserializing incorrect data at the cost of not getting the best error message (would be great if that could be mitigated, but not sure how at this point).  If this trade-off is not worth it I would be happy to close this PR.

cycraig:
> For more imperative tasks like adding encountered errors to a OneOrMany and either returning immediately (if say a fail_fast flag was set) or continue adding errors it is nicer being able to just type errors.push(error) over if let Some(ref mut errors) = maybe_errors { errors.push(error) }.

I don't quite follow. `OneOrMany` serves to enforce that a collection is non-empty and to have a different JSON serialization based on whether there is one entry or many - both to comply with the DID specification. For a collection of errors a simple `Vec` seems more appropriate as it is not subject to the compliance requirements.

olivereanderson:
> OneOrMany serves to enforce that a collection is non-empty and to have a different JSON serialization based on whether there is one entry or many

Well our implementation of `OneOrMany` has a [`is_empty`](https://github.com/iotaledger/identity.rs/blob/dev/identity-core/src/common/one_or_many.rs#L33) method and even implements `Default` that produces an empty collection so I am not sure it ever was intended to be non-empty.   

> and to have a different JSON serialization based on whether there is one entry or many

This is also not enforced, but rather enabled. 

With regards to errors a simple vec can be nice, but `OneOrMany`better conveys that we are not necessarily enforcing collecting multiple errors.

olivereanderson:
Since it is desirable for (de)serialization to reflect what a type encodes as its invariant I have decided to close this PR.

PR #634, opened 2022-01-28
Rename `MethodQuery` to `DIDUrlQuery`, move `OrderedSet`, `KeyComparable`

### LINKED ISSUE #633: [Task] Move `KeyComparable`, `OrderedSet` to `identity-core`
## Description
Move the `OrderedSet` collection in `identity-did/utils` and its associated `KeyComparable` trait to `identity-core`.

## Motivation
It makes more sense to keep it in the same place as other generic structures like `OneOrMany`.

## To-do list

- [ ] Move `OrderedSet` to `identity-core/common`
- [ ] Move `KeyComparable` to `identity-core/common`

## Change checklist

cycraig:
# Description of change

House-keeping changes: moves the `OrderedSet` collection and associated `KeyComparable` trait from `identity-did` to `identity-core`. This also makes `OrderedSet` more generic by introducing the `Queryable` trait for DID-specific lookups.

Breaking change due to different import paths and the introduction of `Queryable` to handle the `query/query_mut` functions. Also renames `MethodQuery` to a more generic `DIDUrlQuery`, since I intend to use it for a future `resolve_service` function.

### Changed

- Move `OrderedSet` from `identity-did` to `identity-core`.
- Move `KeyComparable` from `identity-did` to `identity-core`.
- Rename `MethodQuery` to `DIDUrlQuery` and move its sub-package from `/method` to `/utils`.
- Rename `KeyComparable::as_key()` to `KeyComparable::key()`.

### Added

- Add `Queryable` trait.

Fixes #633.

## Type of change

- [ ] Documentation Fix

Unit tests and examples pass locally.

## Change checklist


cycraig:
Not entirely happy with all changes, such as:

- the naming of `DIDUrlQuery` (still needs to be something that can apply to querying a service or verification method but it really only uses the fragment).
- the naming of the `identity-did/utils` sub-module.
- the fact that the key-uniqueness of `OrderedSet` can be violated by `query_mut`: this is possible in the original code even without these changes through `IotaDocument/CoreDocument::resolve_method_mut` which use it. This is why collections like `HashSet` do not allow mutable iterators or accessors, to avoid this problem. To fix it would mean removing any function that returns a mutable reference into an `OrderedSet`, which would not be so bad in my opinion (the intended usage is through methods like `OrderedSet::replace` for modification) but would be a significant breaking change.

identity-did/src/utils/queryable.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use identity_core::common::KeyComparable;
+use identity_core::common::OrderedSet;
+
+use crate::did::CoreDIDUrl;
+use crate::utils::DIDUrlQuery;
+
+pub trait Queryable<T, Q> {
+  fn query(&self, query: Q) -> Option<&T>;
+  fn query_mut(&mut self, query: Q) -> Option<&mut T>;
```
PhilippGackstatter, line 12: Add some short doc comments.

identity-did/src/verification/verification_method.rs
```diff
@@ -132,3 +132,11 @@ impl<T> AsRef<CoreDIDUrl> for VerificationMethod<T> {
     self.id()
   }
 }
+
+impl<T> KeyComparable for VerificationMethod<T> {
+  type Key = CoreDIDUrl;
+
+  fn key(&self) -> &Self::Key {
+    self.id()
+  }
```
PhilippGackstatter, line 141: Super nit: You added `#[inline]` in the impl for `MethodRef` but not in other places. Not sure if it's really necessary, but would be nice to be consistent.

PhilippGackstatter (approved):
Looks good to me!

I guess a more general name for `DidUrlQuery` would be `DidDocumentPropertyQuery` but that's unwieldy, so the current name seems perfectly fine to me.

Regarding the mutability: While annoying that it is possible, it's unlikely to happen, as someone would have to change the DID of a e.g. method in the `OrderedSet`, iiuc. The account also doesn't expose that except through `update_document_unchecked`, so it's well contained.

identity-did/src/verification/verification_method.rs
cycraig: I added it to `MethodRef` specifically because almost all implementations for `MethodRef` were `#[inline]` but not for other structs. I'll change it for `KeyComparable` though.

HenriqueNogara (approved):

Issue #306, opened 2021-07-12
[Task] Implement all `controller` features from the DID Spec


JelleMillenaar:
## Description

Add all controller features to the framework as they are mentioned in the DID specification. This includes the DID Document controller field and the controller fields inside of verification methods / relationships. 

## Motivation

Allows the creation of identities that are managed by other identities. 

## Resources
Link to any resources relevant for the task such as Issues, PRs, reference implementations, or specifications.

* [DID Controller field](https://www.w3.org/TR/did-core/#did-controller)
* [DID Controller inside verification methods](https://www.w3.org/TR/did-core/#verification-methods)

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Allow controllers to sign DID Messages (During validation)
- [ ] Research how to deal with circular DID Controllers (a controls b, b control a)
- [ ] Provide an API for signing a DID Message that you control
- [ ] Provide an API for adding / removing controllers to DID Documents and verification methods / relationships
- [ ] Add tests
- [ ] Add example
- [ ] WASM Bindings

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

cycraig:
There are several obstacles to implementing controllers right now due to having to resolve the entire history of a DID. Verifying signatures on both integration chain and diff chain updates from controllers require resolving the controller DID document (for every controller in its history); if the referenced verification method was changed or removed, the signature verification will fail. To solve this we would need to pin signatures to a specific version/hash of a DID document.

This was partially addressed in #458 by allowing embedding verification methods from controllers directly, and allowing the verification method `controller` property to be set, referencing the controlling DID without needing to resolve it.

cycraig:
This issue is still on hold due to technical limitations with the current DID method.

The planned UTXO DID Method atop Stardust will see controllers handled at the protocol level rather than the client, since nodes will validate whether or not an address is allowed to update an Alias Output.

PR #638, opened 2022-01-31
Change document controller type to `OneOrSet`

cycraig:
# Description of change
Allows multiple controllers to be set on a `CoreDocument` using a new generic `OneOrSet` collection, which wraps `OrderedSet`. `OneOrSet` is similar to `OneOrMany` except all entries are unique by key and is guaranteed to be non-empty.

This is to comply with the [W3C DID specification](https://www.w3.org/TR/did-core/#did-controller):

> The controller property is OPTIONAL. If present, the value MUST be a string or a [set](https://infra.spec.whatwg.org/#ordered-set) of strings that conform to the rules in [§ 3.1 DID Syntax](https://www.w3.org/TR/did-core/#did-syntax).

This type change is backwards compatible since a single controller DID will deserialize to `OneOrSetInner::One`.

### Changed

- Change `CoreDocument::controller()` from `Option<&CoreDID>` to `Option<&OneOrSet<CoreDID>`.
- Change `CoreDocument::controller_mut()` from `Option<&mut CoreDID>` to `&mut Option<OneOrSet<CoreDID>>`.

### Added

- Add `OneOrSet` collection.
- Add `IotaDocument::set_controller()`.

See issue #306.

## Type of change
- [ ] Documentation Fix

Added new unit tests to ensure adding and removing document controllers is supported and works with diffs.

## Change checklist


PhilippGackstatter (approved):

identity-core/src/common/one_or_set.rs
```diff
@@ -0,0 +1,484 @@
...
+        let set_t: OrderedSet<T> = set_u.into_vec().into_iter().map(Into::into).collect();
+        // Key equivalence could differ between U and T.
+        if set_t.len() == 1 {
+          OneOrSetInner::One(set_t.into_vec().pop().expect("OneOrSet::from infallible"))
+        } else {
+          OneOrSetInner::Set(set_t)
+        }
+      }
+    })
+  }
```
olivereanderson, line 107: I think the following would be an improvement. 
```
 /// Takes a closure and transforms the [`OneOrSet`] by calling that closure on each element.  
  pub fn map<S, F>(self, mut f: F) -> OneOrSet<S>
  where
    S: KeyComparable,
    F: FnMut(T) -> S,
  {
    OneOrSet(match self.0 {
      OneOrSetInner::One(item) => OneOrSetInner::One(f(item)),
      OneOrSetInner::Set(set_u) => {
        let set_t: OrderedSet<S> = set_u.into_vec().into_iter().map(f).collect();
        // Key equivalence could differ between U and T.
        if set_t.len() == 1 {
          OneOrSetInner::One(set_t.into_vec().pop().expect("OneOrSet::from infallible"))
        } else {
          OneOrSetInner::Set(set_t)
        }
      }
    })
  }
```

identity-core/src/common/one_or_set.rs
```diff
@@ -0,0 +1,484 @@
...
+{
+  /// Constructs a new instance with a single item.
+  pub fn new_one(item: T) -> Self {
+    Self(OneOrSetInner::One(item))
+  }
+
+  /// Constructs a new instance from a set of unique items.
+  ///
+  /// Errors if the given set is empty.
+  pub fn new_set(set: OrderedSet<T>) -> Result<Self> {
```
olivereanderson, line 75: nit: 
These could be perhaps be private as one can use the `From` and `TryFrom` traits as constructors.

olivereanderson (changes requested):
This is a nice improvement to the code base, but I think the associated `from<U>(other: OneOrSet<U>)` method should be refactored before this gets merged.

identity-core/src/common/one_or_set.rs
cycraig: I used the same `from<U: Into<T>>` approach in other places like DIDUrl but I think `map` might be better for a collection. Thanks.

identity-core/src/common/one_or_set.rs
cycraig: I would prefer to keep these because they indicate the canonical ways of constructing a `OneOrSet` in absence of an exposed enum. `From` and `TryFrom` implementations are not obvious or easy to find and would not be the first place to look to instantiate a struct.

This is similar to why we have constructors like `IotaDocument::from_verification_method` and `IotaDocument::try_from_core` (which I also argue should be renamed `new_from_verification_method` and `new_from_core`) rather than `From` and `TryFrom` implementations. This is also a point for the Wasm bindings where we would need to add `From` implementations as functions on the struct itself, but that's moot in this case since `OneOrSet` will not be exposed directly.

olivereanderson: That is fair. I won't complain either way.

olivereanderson (approved):
Thanks for the changes.

Issue #515, opened 2021-11-30
[Bug] Investigate `txm` version 8+ compatibility


cycraig:
## Bug description
The Wasm bindings README.md examples do not work with the latest version of `txm` (version 8.0.0 currently).

Last working version is 7.4.7

Likely related to the ESM migration stuff done in https://github.com/anko/txm/releases/tag/v8.0.0

## IOTA Identity version
Which version of IOTA Identity are you using?

- Version: dev branch, currently commit 9616bab399211093562e2cfb51b223a70efe8333

## Hardware specification

- Operating system: Ubuntu 20.04 / WSL

## Steps To reproduce the bug
Explain how the maintainer can reproduce the bug.

1. Install `txm` version 8+: `npm install -g -f txm@8` or `npm install -g -f txm@latest`
2. Navigate to the Wasm bindings directory: `cd bindings/wasm`
3. Run: `txm README.md`

## Expected behaviour

Runs successfully with the following output:

```
TAP version 13
1..1
ok 1 Nodejs Example

# 1/1 passed
# OK
```

## Actual behaviour

Throws an error.

## Errors
Error when running `txm README.md` from the `bindings/wasm` directory with version 8.0.0 (`txm version` gives the same error):
```
file:///C:/Users/cycra/AppData/Roaming/npm/node_modules/txm/src/cli.js:16
const programName = Object.entries(packageMetadata.bin)
                           ^

TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at file:///C:/Users/cycra/AppData/Roaming/npm/node_modules/txm/src/cli.js:16:28
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12) 
```

cycraig:
Related to https://github.com/iotaledger/identity.rs/pull/513 where `txm` was pinned to version 7+.

cycraig:
Seems to be fixed by `txm` 8.0.1.

PR #644, opened 2022-02-04
Run examples on every test

eike-hass:
# Description of change
This PR combines the `Examples` and `build-and-test` job, so that Rust and Wasm examples now run on every test trigger.
To speed up build and execution the build has been combined to optimally utilize sccache and the Wasm test and examples have been heavily parallilized. 

fixes issue #515 

## Type of change

- [ ] Documentation Fix

## Change checklist


eike-hass:
@cycraig The change detection works as we hoped now. There is a potential issues though.
By design the detection now looks at the difference between the last commit and the current one to determine what changed, usually the `checkout` action would checkout a virtual merge commit between the target branch and the latest commit instead though. 
Let's consider the following scenario: If the most recent change is just Wasm it would not run Rust test and the a PR could be merged. I believe that creates the opportunity for earlier Rust failures slipping through, if accidentally being ignored or cancelled for some reason. Do we think that is problematic?
With the default behavior of inspecting the merge commit, PRs that never touch Rust would still be fast, but once there is a Rust change in it, it would always run all the tests.
Does that make sense? What do you think?

cycraig:
Fantastic! I don't quite follow the new behaviour completely though.

> Let's consider the following scenario: If the most recent change is just Wasm it would not run Rust test and the PR could be merged. I believe that creates the opportunity for earlier Rust failures slipping through, if accidentally being ignored or cancelled for some reason. Do we think that is problematic?

Yes, if a PR changes anything in Rust I would expect both the Rust and Wasm jobs to run. This is important because often there are only minor changes in the commits just before pushing to GitHub, such as fixing README or documentation files.

> With the default behavior of inspecting the merge commit, PRs that never touch Rust would still be fast, but once there is a Rust change in it, it would always run all the tests.
Does that make sense? What do you think?

That would be ideal in my mind. So is that not what happens now according to the previous paragraph?

bindings/wasm/cypress/integration/revokeVC.js
```diff
@@ -0,0 +1,26 @@
...
+            await initIdentity('../../../examples/dist/identity_wasm_bg.wasm');
+
+            // NOTE: `cy.wrap(defaultClientConfig()).as('config')` does not always work to make the config available
+            // from the shared context as `this.config` because it has a race condition with initializing the wasm.
+            // So call `defaultClientConfig()` manually for now.
+        });
+        it("Revoke Verifiable Credential", async () => {
+            await repeatAsyncTest(revokeVC, defaultClientConfig());
+        });
+    }
```
cycraig, line 25: Can we put some more of the boilerplate in a shared file to deduplicate it? Like the `defaultCommandTimeout` value and `initIdentity` stuff. Maybe also move `repeatAsyncTest` somewhere under the `cypress` directory too?

eike-hass:
> Fantastic! I don't quite follow the new behaviour completely though.
> 
> > Let's consider the following scenario: If the most recent change is just Wasm it would not run Rust test and the PR could be merged. I believe that creates the opportunity for earlier Rust failures slipping through, if accidentally being ignored or cancelled for some reason. Do we think that is problematic?
> 
> Yes, if a PR changes anything in Rust I would expect both the Rust and Wasm jobs to run. This is important because often there are only minor changes in the commits just before pushing to GitHub, such as fixing README or documentation files.
> 
> > With the default behavior of inspecting the merge commit, PRs that never touch Rust would still be fast, but once there is a Rust change in it, it would always run all the tests.
> > Does that make sense? What do you think?
> 
> That would be ideal in my mind. So is that not what happens now according to the previous paragraph?

So difference is basically does the determination for which tests to run just look at last commit or all changes on the branch / PR. I take it we want it look at all changes on the branch, even if that means running some tests to often, rather then potentially missing a problem due to skipped or ignored tests.

.github/workflows/build-and-test.yml
```diff
@@ -75,15 +99,23 @@ jobs:
...
           args: --all --release
 
+      - name: Run Rust examples
+        working-directory: examples
+        # run examples only on ubuntu for now
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          cargo read-manifest > ../example-manifest.json
+          cd ../
+          cat ./example-manifest.json | for entry in $(jq -r '.targets[].name'); do cargo run --example "$entry" --release || exit 1; done
```
cycraig, line 118: Two things:

1. You can set which manifest is passed to cargo using `manifest-path`. E.g. `cargo read-manifest --manifest-path ./examples/Cargo.toml`.
2. `cargo run --example <name>` works inside the `examples` directory too (but I see that doing so does _not_ use the pre-built examples from the previous step, weird).

I think this can be simplified either way.

```suggestion
      - name: Run Rust examples
        # Run examples only on Ubuntu for now.
        if: matrix.os == 'ubuntu-latest'
        run: |
          cargo read-manifest --manifest-path ./examples/Cargo.toml | for entry in $(jq -r '.targets[].name'); do cargo run --example "$entry" --release || exit 1; done

```

.github/workflows/build-and-test.yml
```diff
@@ -188,8 +216,8 @@ jobs:
           command: test
           args: --manifest-path ./bindings/wasm/Cargo.toml --release
 
-      - name: Run wasm-bindgen tests # tests annotated with #[wasm_bindgen_test]
-        run: wasm-pack test --node
+      - name: Test Wasm unit and examples
+        run: npm run test
```
cycraig, line 220: Nit: name, optional.

```suggestion
      - name: Run Wasm unit tests and examples
        run: npm run test
```

.github/workflows/scripts/check-if-modified.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+#https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgt--ltpathgt82308203
+if [[ $(git diff HEAD^ HEAD -- $1) != '' ]]; then
+    # modified
+    echo 'true'
+else
+    # unmodified
+    echo 'false'
+fi
```
cycraig, line 10: Nit: this can probably be inlined if it's relatively straightforward and only used in one location?

cycraig (changes requested):
Looking good, minor changes.

.github/workflows/build-and-test.yml
eike-hass: Yeah, the reason for 2. seems to be that `target` is resolved from the current working directory. Maybe there is a way to work around it but I like the idea of running from where it is build, so Option 1 it is :)

cycraig:
The Rust examples are still subject to network problems. E.g. https://github.com/iotaledger/identity.rs/runs/5113744898?check_suite_focus=true
Fails with the following error:
```
Error: ClientError(ResponseError(404, "404 page not found\n"))
```

The Wasm examples repeat once if they error (`repeatAsyncTest`), as do the Rust unit tests that call the Tangle.

We will need something similar before I would be confident about running the Rust examples on every PR.

eike-hass:
> The Rust examples are still subject to network problems. E.g. https://github.com/iotaledger/identity.rs/runs/5113744898?check_suite_focus=true Failed with the following error:
> 
> ```
> Error: ClientError(ResponseError(404, "404 page not found\n"))
> ```
> 
> The Wasm examples repeat once if they error (`repeatAsyncTest`), as do the Rust unit tests that call the Tangle.
> 
> We will need something similar for the the Rust examples before I would be confident about them on every PR.

I can hacks something together from the outside with bash our maybe node. Is there a good Rust native way with something like https://docs.rs/retry/latest/retry/?

bindings/wasm/cypress/integration/revokeVC.js
eike-hass: It is being very finnicky with separating init and config. I suspect a problem with referential equality between Wasm and JS. I would suggest we revisit that with the account examples.

cycraig:
> Is there a good Rust native way with something like https://docs.rs/retry/latest/retry/?

Nothing that can be installed as a binary for use externally as far as I know. We use a custom function for the tests:

```rust
async fn network_resilient_test(
  test_runs: u32,
  f: impl Fn(u32) -> Pin<Box<dyn Future<Output = Result<()>>>>,
) -> Result<()> {
  for test_run in 0..test_runs {
    let test_attempt = f(test_run).await;

    match test_attempt {
      error @ Err(Error::IotaError(identity_iota::Error::ClientError(_))) => {
        eprintln!("test run {} errored with {:?}", test_run, error);

        if test_run == test_runs - 1 {
          return error;
        }
      }
      other => return other,
    }
  }
  Ok(())
}
```

I don't really want to add that to all the examples since they are meant to be straightforward and copy-pasteable.

I'll keep working on an alternative my side but unlikely to be ready any time soon (i.e. not for this PR).

eike-hass:
> > Is there a good Rust native way with something like https://docs.rs/retry/latest/retry/?
> 
> Nothing that can be installed as a binary for use externally as far as I know. We use a custom function for the tests:
> 
> ```rust
> async fn network_resilient_test(
>   test_runs: u32,
>   f: impl Fn(u32) -> Pin<Box<dyn Future<Output = Result<()>>>>,
> ) -> Result<()> {
>   for test_run in 0..test_runs {
>     let test_attempt = f(test_run).await;
> 
>     match test_attempt {
>       error @ Err(Error::IotaError(identity_iota::Error::ClientError(_))) => {
>         eprintln!("test run {} errored with {:?}", test_run, error);
> 
>         if test_run == test_runs - 1 {
>           return error;
>         }
>       }
>       other => return other,
>     }
>   }
>   Ok(())
> }
> ```
> 
> I don't really want to add that to all the examples since they are meant to be straightforward and copy-pasteable.
> 
> I'll keep working on an alternative my side but unlikely to be ready any time soon (i.e. not for this PR).

I will wrap it with a script for now with one retry, same as Wasm.

eike-hass:
> > > Is there a good Rust native way with something like https://docs.rs/retry/latest/retry/?
> > 
> > 
> > Nothing that can be installed as a binary for use externally as far as I know. We use a custom function for the tests:
> > ```rust
> > async fn network_resilient_test(
> >   test_runs: u32,
> >   f: impl Fn(u32) -> Pin<Box<dyn Future<Output = Result<()>>>>,
> > ) -> Result<()> {
> >   for test_run in 0..test_runs {
> >     let test_attempt = f(test_run).await;
> > 
> >     match test_attempt {
> >       error @ Err(Error::IotaError(identity_iota::Error::ClientError(_))) => {
> >         eprintln!("test run {} errored with {:?}", test_run, error);
> > 
> >         if test_run == test_runs - 1 {
> >           return error;
> >         }
> >       }
> >       other => return other,
> >     }
> >   }
> >   Ok(())
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't really want to add that to all the examples since they are meant to be straightforward and copy-pasteable.
> > I'll keep working on an alternative my side but unlikely to be ready any time soon (i.e. not for this PR).
> 
> I will wrap it with a script for now with one retry, same as Wasm.

decided to go with `parallel` which has a build in retry. As a bonus it runs things in parallel, which gets our example execution time to under 5 min :) @cycraig can force a failure in one the examples to make sure we would catch it?

cycraig:
> @cycraig can force a failure in one the examples to make sure we would catch it?

Yup, it caught the error.
https://github.com/iotaledger/identity.rs/runs/5115948763?check_suite_focus=true

Coincidentally it also caught a network error in the Wasm examples.

bindings/wasm/package.json
```diff
@@ -18,12 +18,13 @@
     "build:examples": "webpack --config ./examples/webpack.config.js --mode=production",
     "build": "npm run build:web && npm run build:nodejs && npm run build:docs",
     "build-dev": "npm run build-dev:web && npm run build-dev:nodejs && npm run build:docs",
-    "pretest": "npm run build:nodejs",
     "prepublishOnly": "npm run build",
     "serve": "webpack serve",
     "example:node": "node examples/dist/node.js",
     "example:browser": "http-server ./examples/dist/ -c-1 -o ",
+    "test": "concurrently -g --timings \"npm:test:node\" \"npm:test:browser:parallel\" \"npm:test:readme\" \"npm:test:unit\"",
```
cycraig, line 25: Nit: I think the unit tests and examples should still be separate steps in the `build-and-test-wasm` job. If the unit tests fail there is no point in running the examples, and they're relatively fast even when recompiling for the test profile. Optional since it doesn't matter that much.

```suggestion
    "test:examples": "concurrently -g --timings \"npm:test:node\" \"npm:test:browser:parallel\" \"npm:test:readme\"",
```

.github/workflows/build-and-test.yml
```diff
@@ -12,19 +12,49 @@ on:
...
 env:
   RUST_BACKTRACE: full
+  CARGO_INCREMENTAL: 0 # disabled to reduce target cache size and improve sccache (https://github.com/mozilla/sccache#known-caveats)
   SCCACHE_CACHE_SIZE: 2G
   SCCACHE_IDLE_TIMEOUT: 0
   # SCCACHE_RECACHE: 1 # uncomment to clear sccache cache, then re-comment
 
 jobs:
+
+  determine-modification:
```
cycraig, line 31: Nit: name suggestion.
```suggestion
  check-core-modified:
```

example-manifest.json
```diff
@@ -0,0 +1 @@
...es/low-level-api/manipulate_did.rs","edition":"2021","doc":false,"doctest":false,"test":false},{"kind":["example"],"crate_types":["bin"],"name":"merkle_key","src_path":"/home/eike/Workspace/identity.rs/examples/low-level-api/merkle_key.rs","edition":"2021","doc":false,"doctest":false,"test":false},{"kind":["example"],"crate_types":["bin"],"name":"resolution","src_path":"/home/eike/Workspace/identity.rs/examples/low-level-api/resolution.rs","edition":"2021","doc":false,"doctest":false,"test":false},{"kind":["example"],"crate_types":["bin"],"name":"private_tangle","src_path":"/home/eike/Workspace/identity.rs/examples/low-level-api/private_tangle.rs","edition":"2021","doc":false,"doctest":false,"test":false},{"kind":["example"],"crate_types":["bin"],"name":"revoke_vc","src_path":"/home/eike/Workspace/identity.rs/examples/low-level-api/revoke_vc.rs","edition":"2021","doc":false,"doctest":false,"test":false}],"features":{},"manifest_path":"/home/eike/Workspace/identity.rs/examples/Cargo.toml","metadata":null,"publish":[],"authors":[],"categories":[],"keywords":[],"readme":"README.md","repository":null,"homepage":null,"documentation":null,"edition":"2021","links":null,"default_run":null}
```
cycraig, line 1: Remove this file.

cycraig (changes requested):
Rogue file committed.

.github/workflows/build-and-test.yml
eike-hass: I see this job checking for more modifications, when we add more bindings. Can we live with keeping it generic?

bindings/wasm/package.json
eike-hass: handled it inline in the npm command

.github/workflows/build-and-test.yml
cycraig: I guess so, it's just a bit hard to grok what it's for by the name alone. I'm fine leaving it as-is.

eike-hass: maybe just `check-for-modification` or something?

eike-hass: @cycraig what do you think?

cycraig: Oh! Yeah I'm fine with either `determine-modification` or `check-for-modification`. The latter might be more immediately descriptive?

.github/workflows/build-and-test.yml
```diff
@@ -12,19 +12,49 @@ on:
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run change detection
+        id: change-detection
+        run: |
+          echo comparing $(git rev-parse HEAD^) and $(git rev-parse HEAD)
+          #https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgt--ltpathgt82308203
+          if [[ $(git diff HEAD^ HEAD -- $1) != '' ]]; then
```
cycraig, line 44: What is `$1` meant to be? Previously I think `':!bindings'` was passed to the script?

cycraig (changes requested):
It looks good now but I think the `determine-modification` bash is wrong now with `$1`?

.github/workflows/build-and-test.yml
eike-hass: good catch!

.github/workflows/build-and-test.yml
eike-hass: went with `check-for-modification`

.github/workflows/build-and-test.yml
```diff
@@ -188,8 +220,8 @@ jobs:
           command: test
           args: --manifest-path ./bindings/wasm/Cargo.toml --release
```
cycraig, line 221: Aside: shouldn't this step run with the `bindings/wasm` `working-directory` to avoid the `manifest-path`? I bring this up only because I wonder if the different working directory affects cache hits/artifact re-use between this and other steps.

.github/workflows/build-and-test.yml
```diff
@@ -75,15 +104,22 @@ jobs:
...
           command: test
           args: --all --release
 
+      - name: Run Rust examples
+        # run examples only on ubuntu for now
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          cargo read-manifest --manifest-path ./examples/Cargo.toml > ./example-manifest.json
+          # run examples in parallel with 4 threads and 1 retry
+          cat ./example-manifest.json | jq -r '.targets[].name' | parallel -k -j 4 --retries 1 cargo run --example {} --release
```
cycraig, line 122: Nit: do we still need the intermediate file? If the concern is readability could we pipe these over separate lines?

Super optional, I'm totally fine leaving this as-is and don't know if there are any caveats with the trailing pipes.

E.g.
```bash
          # Run examples in parallel with 4 threads and 1 retry.
          cargo read-manifest --manifest-path ./examples/Cargo.toml |
          cat ./example-manifest.json |
          jq -r '.targets[].name' | 
          parallel -k -j 4 --retries 1 cargo run --example {} --release
```

cycraig (approved):
Looks good to me overall now.

Minor questions.

.github/workflows/build-and-test.yml
cycraig: This is the full step (couldn't select it properly from the UI):

```yml
- name: Run tests
  uses: actions-rs/cargo@v1
  with:
    command: test
    args: --manifest-path ./bindings/wasm/Cargo.toml --release
```

eike-hass: The `cargo` action doesn't offer a working-directory option. So I think we only solution would be to run a script instead. It might explain why the target cache isn't be hit, though. @cycraig what do you think, should we change it to script?

cycraig:
The inconsistent failures of examples are still a problem and may be worth reconsidering running examples every time.

https://github.com/iotaledger/identity.rs/runs/5138275461?check_suite_focus=true

There is also no output for the parallel browser examples so I can't even tell exactly why it failed this time.

olivereanderson (approved):
Thank you for working on this!

.github/workflows/build-and-test.yml
cycraig: I think it's definitely worth a try, we don't seem to depend on any of the advanced options of the `cargo` action so changing it to a `run` step should be equivalent (except with the right `working-directory`).

eike-hass:
> The inconsistent failures of examples are still a problem and may be worth reconsidering running examples every time.
> 
> https://github.com/iotaledger/identity.rs/runs/5138275461?check_suite_focus=true
> 
> There is also no output for the parallel browser examples so I can't even tell exactly why it failed this time.

increased retries to 3 across the board, hope that is good enough 😟

PR #648, opened 2022-02-09
Replace Type `Promise<any>` with `Promise<void>`

abdulmth:
# Description of change
Replace return type `Promise<any>` with `Promise<undefined>` for functions that return `JsValue::undefined()`.

Relevant to #574  

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/bindings/wasm/docs/api-reference.md
+++ b/bindings/wasm/docs/api-reference.md
@@ -117,29 +117,29 @@ publishing to the Tangle.
 **Kind**: global class  
 
 * [Account](#Account)
-    * [.deleteMethod(options)](#Account+deleteMethod) ⇒ <code>Promise.&lt;undefined&gt;</code>
-    * [.deleteService(options)](#Account+deleteService) ⇒ <code>Promise.&lt;undefined&gt;</code>
-    * [.createService(options)](#Account+createService) ⇒ <code>Promise.&lt;undefined&gt;</code>
+    * [.deleteMethod(options)](#Account+deleteMethod) ⇒ <code>Promise.&lt;void&gt;</code>
+    * [.deleteService(options)](#Account+deleteService) ⇒ <code>Promise.&lt;void&gt;</code>
+    * [.createService(options)](#Account+createService) ⇒ <code>Promise.&lt;void&gt;</code>
     * [.did()](#Account+did) ⇒ [<code>DID</code>](#DID)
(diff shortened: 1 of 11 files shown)


cycraig (commented):
Looks good to me overall!

Minor question: is there a reason why we return `Promise<undefined>` rather than `Promise<void>`? I know the returns are mapped to `JsValue::undefined()` but I think that's actually unnecessary (removing `.map(|_| JsValue::undefined())` still compiles).

Furthermore the `wasmbindgen` examples for async functions with no return value map to `Promise<void>`:

[Rust](https://github.com/rustwasm/wasm-bindgen/blob/e4477293c15df8500330b295cb9624dccb2a063b/crates/cli/tests/reference/async-void.rs
):
```rust
#[wasm_bindgen]
pub async fn foo() {}
```
[Typescript](https://github.com/rustwasm/wasm-bindgen/blob/e4477293c15df8500330b295cb9624dccb2a063b/crates/cli/tests/reference/async-void.d.ts):
```ts
/**
* @returns {Promise<void>}
*/
export function foo(): Promise<void>;
```

My concern is mostly ergonomics for Typescript developers since I know `void` is more restricted than `undefined`: https://www.typescriptlang.org/docs/handbook/type-compatibility.html#any-unknown-object-void-undefined-null-and-never-assignability

I think if we remove all the `JsValue::undefined()` calls since they seem unnecessary (which I did not realise before), then returning `Promise<void>` would make sense. As I don't have much experience with Typescript I would still defer to your judgement.

cycraig:
> The reason I used undefined instead of void is for JS developers to expect the type better (since void is more of TS thing). And the actual value of the returned function is undefined.

That makes sense but does `Promise<void>` would actually affect the Javascript developers? I thought it was just a Typescript definition so only those using Typescript would even see it?

What would a Javascript-native `Promise` with no return value return, still undefined?

cycraig (approved):
Thanks for the change!

I think the same `undefined`/`void` thing happens for non-async functions in Javascript too.
E.g.

```js
const config = Config.fromNetwork(clientConfig.network);
let ret = config.setNetwork(clientConfig.network);
console.log(ret);
// prints undefined
```
where `setNetwork` has return type `void` in Typescript:
```ts
setNetwork(network: Network): void;
```

So I think `Promise<void>` is consistent.

PR #651, opened 2022-02-10
Fix Wasm `DID` conversion error names

cycraig:
# Description of change
Corrects the name of errors returned from `Client::resolve` and `Client::resolveHistory` when passed an invalid IOTA DID string. This was a result of the switch to using union type parameters for those functions in PR #616.

E.g. try to resolve an IOTA DID with an invalid tag:
```ts
await client.resolve("did::q2SZ8k2HcxwGWJm5tBT3CwMuGf1Ngr7yNX93MEWsv6D");
```

CURRENT (throws a `serde` error):
```
[serde_json::Error]: Invalid Method Id
```

THIS PR (error name matches the correct variant from Rust):
```
InvalidDID: Invalid Method Name
```

This PR also changes the parameter of `WasmExplorerUrl::resolverUrl` from `&str` to `UWasmDID`, to take either a string or `WasmDID` as input. This change is non-breaking.

Note that the `UWasmDID` parameters were changed from borrowed to owned, this change does not change the current behaviour as it still does not null `WasmDID` inputs, and so is also non-breaking.

Simplified example code by removing unnecessary helper functions to make the examples more copy-pasteable.

## Type of change

- [ ] Documentation Fix

Wasm unit tests and examples pass locally.

## Change checklist


eike-hass:
Tested in a local explorer instance and works like a charm!

cycraig:
> @cycraig I like the simplification of the examples, will the logged output be roughly equivalent?

The output is exactly equivalent:
- `logExplorerUrl` and `logResolverUrl` were just wrappers around calling the `explorerUrl` methods with `console.log`, which I inlined. 
- `prettyPrintJSON` included unnecessary `toString` and `JSON.parse` calls: using `JSON.stringify` directly gives the same output.

abdulmth:
Looks good to me. One thing I noticed is the output of `resolve_history` example in the browser is not formatted correctly, I don't think it's related to this PR though.

I also wonder why taking owned parameter does not null out the original. Isn't this supposed to happen?

abdulmth:
@cycraig  We might need to change the HTML `div` tag of the console to `pre` to preserve new lines. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre.

eike-hass (approved):
Giving it a LGTM! Your call if you want to mess with the console formatting some more :)

cycraig:
> @cycraig We might need to change the HTML div tag of output to pre to preserve white space and new lines. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre.

Yeah that works, thanks! As a side-effect it switched to a monospaced font though but I consider that an improvement too.

Issue #647, opened 2022-02-09
[Task] Set `controller`, `alsoKnownAs` fields from `Account`


cycraig:
## Description
Add update methods to the `Account` to allow setting the `controller` and `alsoKnownAs` fields.

## Motivation
The `IotaDocument` allows altering the `controller` and `alsoKnownAs` fields as part of the DID specification. We should expose this functionality in the `Account`, otherwise the only option currently is to perform an unchecked update, which is dangerous.

The exact design of the update methods is unclear. It may not be necessary to create four methods (create and remove operations for each field ) and instead just have two: `set_controller` and `set_also_known_as`? 

## To-do list

- [x] Allow setting `controller` on the `Account` #658
- [x] Allow setting `also_known_as` on the `Account` #658
- [x] Implement WASM bindings.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

cycraig:
Implemented in #658 for Rust, still TODO for Wasm bindings.

Issue #654, opened 2022-02-13
[Task] Improve `Account::attach/detach_method_relationship`


cycraig:
## Description
Modify `Account::attach_method_relationship` and `Account::detach_method_relationship`:

- pluralise the names to match the Wasm bindings `Account`, i.e. `attach_method_relationships` and `detach_method_relationships`.
- allow specifying multiple relationships at once rather than individually, this would allow passing collections of relationships.

Current behaviour:
```rust
account
  .update_identity()
  .attach_method_relationship()
  .fragment("key1")
  .relationship(MethodRelationship::Authentication)
  .relationship(MethodRelationship::CapabilityInvocation)
  .apply()
  .await?;
```
Desired behaviour:

```rust
account
  .update_identity()
  .attach_method_relationship()
  .fragment("key1")
  .relationships(&[MethodRelationship::Authentication, MethodRelationship::CapabilityInvocation])
  .apply()
  .await?;
```

The `.relationships` implementation can either take a slice, vec, or arbitrary `IntoIter` generic depending on which we find is most convenient for developers to use.

We should also think about making attaching/detaching relationships easier for developers, since it is currently restricted to non-embedded methods. That is, trying to attach/detach an embedded method always causes an error, we could instead move the method to be a `verificationMethod`.

## Motivation
Porting the `Account` for the Wasm bindings in #574 demonstrated that using the builder pattern to add/remove multiple method relationships is cumbersome due to ownership/borrowing rules on intermediate structs. This issue is an attempt to make these specific methods more user-friendly and to reduce differences with the Wasm bindings.

## To-do list

- [ ] rename `Account::attach_method_relationship` to `attach_method_relationships`
- [ ] rename `Account::detach_method_relationship` to `detach_method_relationships`
- [ ] allow `Account::attach_method_relationships` to specify multiple relationships at once.
- [ ] allow `Account::detach_method_relationships` to specify multiple relationships at once.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #658, opened 2022-02-14
Set `controller`, `alsoKnownAs` fields from Account

abdulmth:
# Description of change
The change allows setting the `controller` and `alsoKnownAs` fields in the Account. 

Relevant issue #647

## Type of change
- [ ] Documentation Fix

locally tested in examples.


cycraig (changes requested):
Looks good but needs tests similar to the existing `Account` update methods.

cycraig (approved):

identity-account/src/updates/update.rs
```diff
@@ -130,6 +133,12 @@ pub(crate) enum Update {
   DeleteService {
     fragment: String,
   },
+  SetController {
+    controllers: Option<OneOrSet<IotaDID>>,
+  },
+  SetAlsoKnownAs {
+    urls: OrderedSet<Url>,
+  },
```
PhilippGackstatter, line 141: This doesn't really follow the convention the account updates have so far. The existing updates add or remove a service/method from the set of services/methods in the document. These `Set*` updates overwrite whatever was there previously. Either we change this to be `AddController` & `DeleteController` and `AddAlsoKnownAs` & `DeleteAlsoKnownAs` or we change the existing updates to also be `Set*` updates. 

I think the `Set*` variants are rather unwieldy for non-trivial updates, where e.g. one wants to add a controller to a document. You'd have to copy over the old controllers. So, I'm in favor of having the `Add*` and `Delete*`-style updates. One way or another, we should be consistent.

PhilippGackstatter (changes requested):
We should discuss which style to follow for the update types.

identity-account/src/updates/update.rs
cycraig: This is mostly due to my description in issue #647. The `controller` and `also_known_as` fields are simple collections rather than complex structs like `VerificationMethod` and `Service`, so it is my opinion that they don't need fine-grained operations which contribute to API pollution (maybe too strong a term).

PhilippGackstatter: Considering how simple the other account updates are, I don't think set-style updates fit nicely with the ease of the API that we have so far. Just adding a controller requires:

```rust
let set = match account.document().controller() {
  Some(set) => {
    let mut set = OrderedSet::from_iter(set.to_vec());
    set.append(iota_did3);
    OneOrSet::new_set(set).unwrap()
  }
  None => OneOrSet::new_one(iota_did3),
};

let update: Update = Update::SetController {
  controllers: Some(set),
};
account.process_update(update).await.unwrap();
```

I assume these updates "in code" are infrequent, to either controller or alsoKnownAs, compared to services or methods, but when they occur it's quite verbose.
I generally don't see set operations for collections exposed on APIs because they are unwieldy to work with. And since in our API we don't have the possibility to return a `&mut OrderedSet` to let users append through that, I think the nicest solution would be to expose add and delete operations. Of course, I would like to avoid the pollution of the API, but that takes lower priority in my mind.

cycraig: That example can be simplified somewhat:

```rust
let set = match account.document().controller().cloned() {
  Some(mut one_or_set) => {
    one_or_set.append(iota_did3);
    one_or_set
  }
  None => OneOrSet::new_one(iota_did3),
};

let update: Update = Update::SetController {
  controllers: Some(set),
};
account.process_update(update).await.unwrap();
```

Would they be better if they took `Vec` instead, to work around the awkward `OneOrSet` restrictions at the risk of runtime errors?

> I generally don't see set operations for collections exposed on APIs because they are unwieldy to work with.
My problem with individual operations on collections is that it's even more difficult to use when you actually have a collection you want to pass to the builder. As was seen when exposing the `attach/detach_method_relationships` API to the Wasm bindings due to intermediate, temporary structs in the builder pattern being dropped.

```rust
  let mut account: RefMut<Account> = account.borrow_mut();
  let mut updater: IdentityUpdater<'_> = account.update_identity();
  let mut attach_relationship: AttachMethodRelationshipBuilder<'_> =
    updater.attach_method_relationship().fragment(fragment);

  for relationship in relationships {
    attach_relationship = attach_relationship.relationship(relationship);
  }
```

cycraig: Regarding `relationships` I opened #654 previously and recommended either a vec, slice, or iterator as input since `MethodRelationship` implements `Copy`. 

I wouldn't mind using `TryInto`. The only difference is that it moves the error from `apply` up to the field call at `relationships`, so really the example would be:

```rust
pub fn relationships<T, E>(self, t: T) -> Result<Self>
where
  E: KeyComparable,
  T: TryInto<OneOrSet<E>> {}
```

```rust
  account
    .update_identity()
    .attach_method_relationship()
    .fragment("my-next-key")
    .relationships(MethodRelationship::CapabilityDelegation)?
    .relationships(vec![MethodRelationship::CapabilityInvocation, MethodRelationship::AssertionMethod])?
    .apply()
    .await?;
```

Not much different but it sets a precedent about where update builder errors can be thrown. One good thing about requiring the actual type (`OneOrSet`) is that it forces the developer to handle those errors up-front rather than having to interpret errors from the builder, but I do admit it is less easy-to-use.

> We could do that similarly for controllers/alsoKnownAs. I think that would at least solve the "extend an existing collection with a collection" problem.

Back to this: what's the conclusion? Are we staying with `set_controller`/`set_also_known_as` (with improved ergonomics for the input parameters such as `Vec` or `TryInto`) or are we set on splitting them into four methods for create/remove each? I'm still of the opinion that `set_*` methods are preferable for these fields.

PhilippGackstatter: For builders, I think I'd prefer them not to throw errors. Not sure how common errors in builders are. If `From<Vec<T>> for OrderedSet` was implemented, that would make things a lot easier, and we could by ignoring duplicates. But I suppose that was a deliberate choice, so I don't want to reopen that discussion, too.

> what's the conclusion? Are we staying with set_controller/set_also_known_as (with improved ergonomics for the input parameters such as Vec or TryInto)

I'm still more in favor of the create/delete methods, since with the `Vec` or `TryInto`, it should make working with single and collection values fairly easy, and because adding any value with set-style methods is cumbersome.
Since we're not really coming to a conclusion, though, I'm fine with going with set-updates for now and see how they turn out in practice. In that case, I would go with `Vec` to avoid errors on the builder method.

Unless there's more opinions, @abdulmth this might be the conclusion.

abdulmth: @PhilippGackstatter and I agreed in a call that we might want to keep the implementation as-is since it's not trivial to implement the delete methods because it must deal with `IotaDID` and `Url` objects rather than enum variants (for example in `detach_relationships`). We also don't wanna use `Vec` to keep parity with the return values of the getter methods.

PR #659, opened 2022-02-15
Skip build and test on draft PRs

eike-hass:
# Description of change
This PR adds conditions to only run the jobs in `build-and-test` if the PR is not in "Draft" mode. The branch triggers should be unaffected by this.

## Type of change

- [ ] Documentation Fix

In "Draft" mode the test where skipped in this PR. On setting "Ready for Review" the tests where executed.

## Change checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -38,7 +38,7 @@ jobs:
 
   check-for-modification:
     needs: check-for-run-condition
-    if: ${{ needs.check-for-run-condition.outputs.should-run }}
+    if: ${{ needs.check-for-run-condition.outputs.should-run == 'true' }}
     runs-on: ubuntu-latest
     outputs:
       core-modified: ${{ steps.change-detection.outputs.core-modified }} # map step output to job output
@@ -64,7 +64,7 @@ jobs:
   build-and-test:
     runs-on: ${{ matrix.os }}
     needs: [check-for-run-condition, check-for-modification]
-    if: ${{ needs.check-for-run-condition.outputs.should-run && needs.check-for-modification.outputs.core-modified == 'true' }}
+    if: ${{ needs.check-for-run-condition.outputs.should-run == 'true' && needs.check-for-modification.outputs.core-modified == 'true' }}
     strategy:
       fail-fast: false
       matrix:
@@ -141,7 +141,7 @@ jobs:
 
   build-and-test-libjose:
     needs: check-for-run-condition
-    if: ${{ needs.check-for-run-condition.outputs.should-run }}
+    if: ${{ needs.check-for-run-condition.outputs.should-run == 'true' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -53,7 +55,7 @@ jobs:
   build-and-test:
     runs-on: ${{ matrix.os }}
     needs: check-for-modification
-    if: ${{ needs.check-for-modification.outputs.core-modified == 'true' }}
+    if: ${{ (!github.event.pull_request || github.event.pull_request.draft == false) && needs.check-for-modification.outputs.core-modified == 'true' }}
```
cycraig, line 58: Can we make the check its own job and have all other jobs in this workflow specify it in `needs`? That would avoid the duplicate if-conditions if nothing much else.

.github/workflows/build-and-test.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
       - dev
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
cycraig, line 9: I assume `synchronize` includes new pushes to the PR? Just wondering if this needs `edited` too (I assume that actually means the description is edited not the content but I couldn't find the right documentation).

What scenario do these conditions prevent by the way?

eike-hass: `open`, `synchronize` and `reopened` are equivalent to just having `pull_request` (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). So the change to before is the addition of `ready_for_review`. It sucks that can't go "default + `ready_for_review`", but I don't think the defaults would change much.

.github/workflows/build-and-test.yml
```diff
@@ -26,8 +27,17 @@ env:
...
+    runs-on: ubuntu-latest
+    outputs:
+      should-run: ${{ !github.event.pull_request || github.event.pull_request.draft == false }}
+    steps:
+      - run: |
+          # this run step does nothing, but is needed to get the job output
 
   check-for-modification:
+    needs: check-for-run-condition
+    if: ${{ needs.check-for-run-condition.outputs.should-run == 'true' }}
```
cycraig, line 40: I was thinking more along the lines of placing the condition in the `if` section of `check-for-run-condition`. Then add it to `needs` for the rest of the jobs. That way if it's skipped the rest of the jobs are skipped too (if I understand correctly).

E.g.
```yml
  check-for-run-condition:
    runs-on: ubuntu-latest
    if: ${{ !github.event.pull_request || github.event.pull_request.draft == false }}
    steps:
      - run: |
          # this run step does nothing, but is needed to get the job output

  check-for-modification:
    needs: check-for-run-condition
```

eike-hass: Oh, for me that feels unidiomatic. In my understanding `needs` orders the jobs. It is meant to handle success and failure (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds), I think absence / skipped is undefined. Another possibility I looked into was putting the result of the condition in a env variable, but they can't be evaluated in the context of `if` .

cycraig (approved):
Cool, looks good to me.

PR #599, opened 2022-01-17
Feat/credential validation mvp

olivereanderson:
# Description of change 
This PR introduces a new VC & VP (verifiable credentials and verifiable presentations respectively) validation API. The main differences from the current API are summarised here: 

- Introduces more fine grained control over what gets validated and how. 
- The DID Documents of credential subjects are no longer resolved during credential validation. 
- Adds support for utilising previously resolved DID Documents (belonging to an issuer or presenter) when validating credentials. 
- Introduces granular errors describing more precisely how validation failed (closes #312).  

Here are the changes in more detail: 
## Changes to the high level API  
The old `CredentialValidator` provided the following high level functions for verifying credentials: 

```
pub async fn validate_credential<T>(
    &self,
    credential: Credential<T>,
    options: VerifierOptions,
  ) -> Result<CredentialValidation<T>>;

pub async fn validate_presentation<T, U>(
    &self,
    presentation: Presentation<T, U>,
    options: VerifierOptions,
  ) -> Result<PresentationValidation<T, U>>; 

```
The latter of these has been replaced with the following method on the new `Resolver`: 

``` 
 pub async fn verify_presentation<U: Serialize, V: Serialize>(
    &self,
    presentation: &Presentation<U, V>,
    options: &PresentationValidationOptions,
    holder: Option<&ResolvedIotaDocument>,
    issuers: Option<&[ResolvedIotaDocument]>,
    fail_fast: FailFast,
  ) -> Result<()> ;
```

The new `verify_presentation` method will resolve the DID Document(s) of the holder and/or issuers if `None` is passed in their place. Moreover the `options` parameters enables the caller to declare some validation rules such as "when is the earliest a credential is allowed to expire" etc. 

There is no replacement for `validate_credential` in this PR as we want to emphasize that one is mostly expected to verify presentations and to a much lesser degree credentials. It is however still possible to validate credentials with the changes from this PR, but it involves the new low level validation API. 

## New low level validation API 

There are new structs `PresentationValidator` and `CredentialValidator` (not to be confused with the old validator of the same name) that provide various methods for single concern validations of the `Credential/Presentation`. More precisely (the new) `CredentialValidator` has the following (associated) methods: 

- check_structure
- check_expires_on_or_after
- check_is_issued_on_or_before
- verify_signature 
and a somewhat higher level method `validate` which is similar to the old `validate_credential` but is different in the sense that it takes the issuer's DID Document as an additional argument rather than resolving it during validation. 

[...]

PR #660, opened 2022-02-15
Feat/stronghold napi

HenriqueNogara:
# Description of change
This PR adds NodeJs bindings for the Stronghold wrapper. 

## Type of change
- [ ] Documentation Fix

All Rust examples have been written in Typescript. 

## Change checklist


bindings/nodejs/.github/workflows/CI.yml
```diff
@@ -0,0 +1,495 @@
+name: CI
```
eike-hass, line 1: This file needs to moved in the root workflows folder and cleaned up. I can totally help with hat.

bindings/nodejs/examples/create_did.mjs
```diff
@@ -0,0 +1,41 @@
...
+ */
+async function createIdentity() {
+
+    // Sets the location and password for the Stronghold
+    //
+    // Stronghold is an encrypted file that manages private keys.
+    // It implements best practices for security and is the recommended way of handling private keys.
+    let strongholdPath = "./example-strong.hodl";
+    let password = "my-password";
+    let stronghold = new Stronghold(strongholdPath, password, true);
```
eike-hass, line 19: can we use an option object similar to the AccountBuilder options here?

bindings/nodejs/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "nodejs",
+  "version": "0.0.0",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "napi": {
+    "name": "nodejs",
+    "triples": {
+      "additional": [
+        "aarch64-apple-darwin",
```
eike-hass, line 10: Is this the place where we control the target arches? Should probably be ["aarch64-apple-darwin", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-pc-windows-msvc"]

bindings/nodejs/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+edition = "2021"
+name = "nodejs"
+version = "0.0.0"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+bincode = { version = "1.3.3" }
+identity = { version = "=0.5.0-dev.3", path = "../../identity", default-features = false, features = ["account", "stronghold", "async"] }
```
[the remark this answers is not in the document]
cycraig: I concur, we might need an `identity-account-storage` sub-crate to reduce overlap and Wasm size. We would really need to work on feature-gating core crates to minimise overlap but cutting out `identity-iota` crate would be great already.

bindings/nodejs/code/stronghold_storage.ts
```diff
@@ -0,0 +1,110 @@
+import { Buffer } from 'buffer';
+import { NapiStronghold, NapiDID, NapiKeyLocation, NapiChainState, NapiIdentityState, NapiGeneration } from '../index.js'
+import { DID, KeyLocation, Generation, DIDLease, Signature, ChainState, IdentityState, Storage } from "../../../wasm/node/identity_wasm.js";
```
eike-hass, line 3: Skeptical about this dependency. If the Wasm bindings change, without these bindings being updated things might break. Ideally the TS types and JS functions get created from the Rust implementation in these bindings.

bindings/nodejs/npm/android-arm64/README.md
```diff
@@ -0,0 +1,3 @@
+# `nodejs-android-arm64`
```
eike-hass, line 1: are the arch specific folders automatically (re-)created, or do they need to be checked in? Otherwise we should gitignore them.

bindings/nodejs/package.json
```diff
@@ -0,0 +1,38 @@
...
+        "x86_64-unknown-freebsd",
+        "i686-pc-windows-msvc",
+        "armv7-linux-androideabi"
+      ]
+    }
+  },
+  "license": "MIT",
+  "devDependencies": {
+    "@napi-rs/cli": "^2.4.2"
+  },
```
eike-hass, line 26: We can express the link to `@iota/identity-wasm` and the version contraint through `peerDependencies`

bindings/nodejs/package.json
HenriqueNogara: no idea if this is the best place to control this. It was automatically added when I created the folder with `napi new`

bindings/nodejs/code/stronghold_storage.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export class Stronghold implements Storage {
+    private napiStronghold: NapiStronghold;
+
+    constructor(snapshot: string, password: string, dropsave: boolean) {
+        this.napiStronghold = NapiStronghold.create(snapshot, password, dropsave);
+    }
+
+    public async setPassword(encryptionKey: Uint8Array): Promise<void> {
```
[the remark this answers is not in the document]
HenriqueNogara: I have already removed this. Turns out we have problems when serializing structs annotated with `#[serde(flatten)]` to bytes.

bindings/nodejs/npm/android-arm64/README.md
HenriqueNogara: I guess we need them, since they are create with `napi new`

bindings/nodejs/package.json
cycraig: Either way I don't think we should support all these platforms arbitrarily since we can't provide guarantees that they work, and we would prefer developers to use the Rust library directly.

bindings/nodejs/rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+tab_spaces = 2
+edition = "2021"
```
cycraig, line 2: This should match the `rustfmt.toml` in the Rust directory and enforce licence headers.

bindings/nodejs/src/account/identity/did_lease.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use identity::account::DIDLease;
+
+#[napi(js_name = NapiDIDLease)]
+pub struct NapiDIDLease(pub(crate) DIDLease);
+
```
cycraig, line 8: This can be removed after #664.

bindings/nodejs/src/account/identity/identity_state.rs
```diff
@@ -0,0 +1,30 @@
...
+#[napi]
+#[derive(Deserialize, Serialize)]
+pub struct NapiIdentityState(pub(crate) IdentityState);
+
+#[napi]
+impl NapiIdentityState {
+  #[napi(js_name = fromJSON)]
+  pub fn from_json(json_value: serde_json::Value) -> Result<NapiIdentityState> {
+    serde_json::from_value(json_value).napi_result()
+  }
```
cycraig, line 18: Unnecessary derive.

```suggestion
#[napi]
pub struct NapiIdentityState(pub(crate) IdentityState);

#[napi]
impl NapiIdentityState {
  #[napi(js_name = fromJSON)]
  pub fn from_json(json_value: serde_json::Value) -> Result<NapiIdentityState> {
    serde_json::from_value(json_value).map(Self).napi_result()
  }
```

bindings/nodejs/src/account/identity/chain_state.rs
```diff
@@ -0,0 +1,30 @@
...
+  #[napi(js_name = fromJSON)]
+  pub fn from_json(json_value: serde_json::Value) -> Result<NapiChainState> {
+    serde_json::from_value(json_value).napi_result()
+  }
+
+  #[napi(js_name = toJSON)]
+  pub fn to_json(&self) -> Result<serde_json::Value> {
+    serde_json::to_value(&self).napi_result()
+  }
+}
```
cycraig, line 24: Unnecessary derive.

```suggestion
#[napi]
pub struct NapiChainState(pub(crate) ChainState);

#[napi]
impl NapiChainState {
  #[napi(js_name = fromJSON)]
  pub fn from_json(json_value: serde_json::Value) -> Result<NapiChainState> {
    serde_json::from_value(json_value).map(Self).napi_result()
  }

  #[napi(js_name = toJSON)]
  pub fn to_json(&self) -> Result<serde_json::Value> {
    serde_json::to_value(&self.0).napi_result()
  }
}
```

bindings/nodejs/src/account/types/key_location.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Deserialize)]
+pub struct NapiKeyLocation(pub(crate) KeyLocation);
+
+#[napi]
+impl NapiKeyLocation {
+  #[napi(js_name = fromJSON)]
+  pub fn from_json(json_value: serde_json::Value) -> Result<NapiKeyLocation> {
+    serde_json::from_value(json_value).napi_result()
+  }
+}
```
cycraig, line 19: Unnecessary derive.

```suggestion
#[napi]
pub struct NapiKeyLocation(pub(crate) KeyLocation);

#[napi]
impl NapiKeyLocation {
  #[napi(js_name = fromJSON)]
  pub fn from_json(json_value: serde_json::Value) -> Result<NapiKeyLocation> {
    serde_json::from_value(json_value).map(Self).napi_result()
  }
}
```

bindings/nodejs/src/account/types/signature.rs
```diff
@@ -0,0 +1,25 @@
...
+  }
+}
+
+#[napi]
+impl NapiSignature {
+  #[napi(js_name = toJSON)]
+  pub fn to_json(&self) -> Result<serde_json::Value> {
+    serde_json::to_value(&self.0).napi_result()
+  }
+}
```
cycraig, line 25: Unnecessary derive.

```suggestion
#[napi]
pub struct NapiSignature(pub(crate) Signature);

#[napi]
impl NapiSignature {
  #[napi(js_name = toJSON)]
  pub fn to_json(&self) -> Result<serde_json::Value> {
    serde_json::to_value(&self.0).napi_result()
  }
}

impl From<Signature> for NapiSignature {
  fn from(signature: Signature) -> Self {
    NapiSignature(signature)
  }
}
```

bindings/nodejs/src/did/napi_did.rs
```diff
@@ -0,0 +1,19 @@
...
+#[derive(Deserialize)]
+pub struct NapiDID(pub(crate) IotaDID);
+
+#[napi(js_name = NapiDID)]
+impl NapiDID {
+  #[napi(js_name = fromJSON)]
+  pub fn from_json(json_value: serde_json::Value) -> Result<NapiDID> {
+    serde_json::from_value(json_value).napi_result()
+  }
+}
```
cycraig, line 19: Unnecessary derive.

```suggestion
#[napi(js_name = NapiDID)]
pub struct NapiDID(pub(crate) IotaDID);

#[napi(js_name = NapiDID)]
impl NapiDID {
  #[napi(js_name = fromJSON)]
  pub fn from_json(json_value: serde_json::Value) -> Result<NapiDID> {
    serde_json::from_value(json_value).map(Self).napi_result()
  }
}
```

bindings/nodejs/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::all)]
+
+#[macro_use]
+extern crate serde;
```
cycraig, line 7: Nit: we no longer need `extern crate` since the 2018 edition. Should also forbid unsafe code.

```suggestion
#![forbid(unsafe_code)]
#![deny(clippy::all)]
```

bindings/wasm/src/account/identity/chain_state.rs
```diff
@@ -4,9 +4,12 @@
 use identity::account::ChainState;
 use wasm_bindgen::prelude::*;
 
+use crate::error::Result;
+use crate::error::WasmResult;
 use crate::tangle::WasmMessageId;
 
 #[wasm_bindgen(js_name = ChainState, inspectable)]
+#[derive(Serialize, Deserialize)]
```
cycraig, line 12: Unnecessary derive.

```suggestion
```

bindings/wasm/src/account/identity/chain_state.rs
```diff
@@ -50,6 +53,18 @@ impl WasmChainState {
...
+  #[wasm_bindgen(js_name = toJSON)]
+  pub fn to_json(&self) -> Result<JsValue> {
+    JsValue::from_serde(&self.0).wasm_result()
+  }
+
+  /// Deserializes a JSON object as `ChainState`.
+  #[wasm_bindgen(js_name = fromJSON)]
+  pub fn from_json(json_value: JsValue) -> Result<WasmChainState> {
+    json_value.into_serde().wasm_result()
+  }
```
cycraig, line 67:
```suggestion
  /// Deserializes a JSON object as `ChainState`.
  #[wasm_bindgen(js_name = fromJSON)]
  pub fn from_json(json_value: JsValue) -> Result<WasmChainState> {
    json_value.into_serde().map(Self).wasm_result()
  }
```

bindings/wasm/src/account/identity/identity_state.rs
```diff
@@ -4,11 +4,36 @@
 use identity::account::IdentityState;
 use wasm_bindgen::prelude::*;
 
+use crate::error::Result;
+use crate::error::WasmResult;
+
 #[wasm_bindgen(js_name = IdentityState, inspectable)]
+#[derive(Serialize, Deserialize)]
```
cycraig, line 11: Unnecessary derive.

```suggestion
```

bindings/wasm/src/account/identity/identity_state.rs
```diff
@@ -4,11 +4,36 @@
...
+  #[wasm_bindgen(js_name = toJSON)]
+  pub fn to_json(&self) -> Result<JsValue> {
+    JsValue::from_serde(&self.0).wasm_result()
+  }
+
+  /// Deserializes a JSON object as `IdentityState`.
+  #[wasm_bindgen(js_name = fromJSON)]
+  pub fn from_json(json_value: JsValue) -> Result<WasmIdentityState> {
+    json_value.into_serde().wasm_result()
+  }
```
cycraig, line 26:
```suggestion
  /// Deserializes a JSON object as `IdentityState`.
  #[wasm_bindgen(js_name = fromJSON)]
  pub fn from_json(json_value: JsValue) -> Result<WasmIdentityState> {
    json_value.into_serde().map(Self).wasm_result()
  }
```

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -45,52 +46,46 @@ extern "C" {
   pub type PromiseSignature;
   #[wasm_bindgen(typescript_type = "Promise<boolean>")]
   pub type PromiseBool;
-  #[wasm_bindgen(typescript_type = "Promise<Generation>")]
-  pub type PromiseOptionGeneration;
-  #[wasm_bindgen(typescript_type = "Promise<ChainState>")]
+  #[wasm_bindgen(typescript_type = "Promise<ChainState | undefined | null>")]
   pub type PromiseOptionChainState;
-  #[wasm_bindgen(typescript_type = "Promise<IdentityState>")]
+  #[wasm_bindgen(typescript_type = "Promise<IdentityState | undefined | null>")]
```
cycraig, line 51: Do we need both `undefined` and `null`?

bindings/wasm/src/account/types/generation.rs
```diff
@@ -8,6 +8,7 @@ use crate::error::Result;
 use crate::error::WasmResult;
 
 #[wasm_bindgen(js_name = Generation)]
+#[derive(Deserialize, Serialize)]
```
cycraig, line 11: Unnecessary derive.

```suggestion
```

bindings/wasm/src/account/types/key_location.rs
```diff
@@ -6,8 +6,11 @@ use wasm_bindgen::prelude::*;
 
 use crate::account::types::WasmGeneration;
 use crate::did::WasmMethodType;
+use crate::error::Result;
+use crate::error::WasmResult;
 
 #[wasm_bindgen(js_name = KeyLocation, inspectable)]
+#[derive(Serialize)]
```
cycraig, line 13: Unnecessary derive.

```suggestion
```

bindings/wasm/src/account/types/key_location.rs
```diff
@@ -40,6 +43,12 @@ impl WasmKeyLocation {
   pub fn generation(&self) -> WasmGeneration {
     self.0.generation().into()
   }
+
+  /// Serializes `Signature` as a JSON object.
+  #[wasm_bindgen(js_name = toJSON)]
+  pub fn to_json(&self) -> Result<JsValue> {
+    JsValue::from_serde(&self.0).wasm_result()
+  }
```
cycraig, line 51: Ideally we should have both complementary `fromJSON` and `toJSON` implementations.

bindings/wasm/src/account/types/signature.rs
```diff
@@ -9,8 +9,10 @@ use wasm_bindgen::prelude::*;
 
 use crate::error::wasm_error;
 use crate::error::Result;
+use crate::error::WasmResult;
 
 #[wasm_bindgen(js_name = Signature, inspectable)]
+#[derive(Deserialize)]
```
cycraig, line 15: Unnecessary derive.

```suggestion
```

bindings/wasm/src/account/types/signature.rs
```diff
@@ -33,4 +35,16 @@ impl WasmSignature {
   pub fn data(&self) -> Vec<u8> {
     self.0.data().to_vec()
   }
+
+  /// Deserializes a JSON object as `Signature`.
+  #[wasm_bindgen(js_name = fromJSON)]
+  pub fn from_json(json_value: JsValue) -> Result<WasmSignature> {
+    json_value.into_serde().wasm_result()
+  }
```
cycraig, line 43: Ideally we should have both complementary `fromJSON` and `toJSON` implementations.

```suggestion
  /// Deserializes a JSON object as `Signature`.
  #[wasm_bindgen(js_name = fromJSON)]
  pub fn from_json(json_value: JsValue) -> Result<WasmSignature> {
    json_value.into_serde().map(Self).wasm_result()
  }
```

bindings/nodejs/Cargo.toml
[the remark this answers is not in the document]
PhilippGackstatter: If the required feature-gating is significant, maybe it also makes sense to add a `identity-iota-core` crate. Then `identity-iota` can depend on that crate. And `identity-account-storage` can only depend on `identity-iota-core` and contain only the `Storage` trait by default. Note that the `Storage` trait also required `IotaDID` for its declaration. Stronghold could be enabled through a feature flag. Then napi stronghold (this binding) can depend only on `iota-account-storage` with the `stronghold` feature flag. All that has to start with `identity-iota` though. I think I'm leaning more towards the separate crate, but not a strong opinion.

bindings/wasm/src/account/types/signature.rs
```diff
@@ -9,8 +9,10 @@ use wasm_bindgen::prelude::*;
 
 use crate::error::wasm_error;
 use crate::error::Result;
+use crate::error::WasmResult;
 
 #[wasm_bindgen(js_name = Signature, inspectable)]
+#[derive(Deserialize)]
 pub struct WasmSignature(pub(crate) Signature);
```
cycraig, line 16: Can we investigate whether we actually need a custom `Account` signature? From what I see we don't use the returned public key, only the signature data vector.

bindings/wasm/src/did/wasm_did.rs
```diff
@@ -14,7 +14,7 @@ use crate::tangle::WasmNetwork;
 
 /// @typicalname did
 #[wasm_bindgen(js_name = DID, inspectable)]
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize)]
```
cycraig, line 17: Unnecessary derive.

```suggestion
```

cycraig (commented):
Preliminary comments.

bindings/nodejs/code/stronghold_storage.ts
cycraig: It uses serialization to convert the values so as long as the JSON representations remain compatible (which should be true for non-major semver changes, even when adding new fields) it should always work.

bindings/nodejs/code/stronghold_storage.ts
```diff
@@ -0,0 +1,93 @@
+import { NapiStronghold, NapiDID, NapiKeyLocation, NapiChainState, NapiIdentityState } from '../index.js'
```
PhilippGackstatter, line 1: Nit: Can we rename the directory of this file (`code`) to something more specific, like `node`, `node-adapter` or `adapter` or something like that?

PhilippGackstatter (commented):
Some first comments below. I hope we can get rid of a lot of the napi-generated files relating to architectures, as was already pointed out by others.

bindings/nodejs/code/stronghold_storage.ts
```diff
@@ -0,0 +1,93 @@
...
+import { DID, KeyLocation, DIDLease, Signature, ChainState, IdentityState, Storage } from "../../wasm/node/identity_wasm.js";
+
+export class Stronghold implements Storage {
+    private napiStronghold: NapiStronghold;
+
+    constructor(snapshot: string, password: string, dropsave: boolean) {
+        this.napiStronghold = NapiStronghold.create(snapshot, password, dropsave);
+    }
```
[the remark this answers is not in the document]
abdulmth: A constructor can return a promise in JS, although it's considered a bad practice, see [here](https://stackoverflow.com/a/24686979/7090185). I'm not sure why using a `block_on` is bad in this case. But the other option is as you said by using an async static method and throw an error in the constructor, since the class will have the default constructor otherwise.

bindings/nodejs/src/account/storage/stronghold.rs
```diff
@@ -0,0 +1,202 @@
...
+  /// Set whether to save the storage changes on drop.
+  /// Default: true
+  #[napi(setter)]
+  pub fn set_dropsave(&mut self, dropsave: bool) {
+    self.0.set_dropsave(dropsave);
+  }
+
+  /// Sets the account password.
+  #[napi]
+  pub async fn set_password(&self, password: Vec<u32>) -> Result<()> {
```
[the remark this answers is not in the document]
HenriqueNogara: Unfortunately  we can't. 
```
#[napi] has type `napi::bindgen_prelude::Uint8Array` which is not `Send`
```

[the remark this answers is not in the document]
PhilippGackstatter: Looks like a shortcoming of napi. It requires the `Send` bound on `async fn`s because of `tokio`. Guess `Uint8Array` can only be used in sync functions for now. As a slight improvement, you could put the `Vec<u32>` -> `Vec<u8>` conversion in a separate function and call it from `key_sign` and `set_password`.

bindings/nodejs/Cargo.toml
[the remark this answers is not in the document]
HenriqueNogara: As we discussed, we will be doing this on a separate PR.

bindings/stronghold-nodejs/.cargo/config.toml
```diff
@@ -0,0 +1,7 @@
+[target.aarch64-unknown-linux-gnu]
```
eike-hass, line 1: Our current target are 
´´´
"aarch64-apple-darwin",
"x86_64-apple-darwin",
"x86_64-pc-windows-msvc",
"x86_64-unknown-linux-musl",
"x86_64-unknown-linux-gnu"
´´´
so I think this can be removed

bindings/stronghold-nodejs/.gitignore
```diff
@@ -0,0 +1,15 @@
+# Rust
+target/
+Cargo.lock
```
eike-hass, line 3: Since we don't ship this as a Rust library a committed Cargo.lock might make sense

bindings/stronghold-nodejs/rustfmt.toml
```diff
@@ -0,0 +1,9 @@
+comment_width = 120
```
eike-hass, line 1: I think in the Wasm binding this is linked to the root config, we should probably do the same here.

bindings/stronghold-nodejs/.gitignore
cycraig: We should be consistent though, so if we include the `Cargo.lock` here it should be committed with the Wasm bindings too.

eike-hass:
The final triggers for the NAPI CI or a merge with existing workflows will be tackled on the epic.

eike-hass (approved):

identity-account/Cargo.toml
```diff
@@ -48,6 +48,7 @@ optional = true
 rand = { version = "0.8" }
 rusty-fork = { version = "0.3" }
 tokio = { version = "1.14", features = ["macros", "rt", "rt-multi-thread", "sync"] }
+identity-account= { path = ".", features = ["send-sync-storage"] }
```
cycraig, line 51: As discussed, I think `send-sync-storage` should be a default feature on `identity-account` instead.

bindings/stronghold-nodejs/src/error.rs
```diff
@@ -0,0 +1,31 @@
...
+/// Convenience trait to simplify `result.map_err(account_error)` to `result.napi_result()`
+pub trait NapiResult<T> {
+  fn napi_result(self) -> Result<T>;
+}
+
+impl<T> NapiResult<T> for AccountResult<T> {
+  fn napi_result(self) -> Result<T> {
+    self.map_err(|account_error| Error::from_reason(account_error.to_string()))
+  }
+}
```
cycraig, line 19: Could we implement `NapiResult` for any `std::error::Error` or does that change the error message?

Does `error.to_string()` include the error variant name in the message? It might not which would make the messages less useful.

bindings/stronghold-nodejs/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![deny(clippy::all)]
+#![forbid(unsafe_code)]
+
+#[macro_use]
+extern crate napi_derive;
```
cycraig, line 8: Nit: I don't think `extern crate` is necessary? We should be able to use its macros via imports.

bindings/wasm/src/account/storage/traits.rs
```diff
@@ -43,52 +43,44 @@ extern "C" {
...
```
cycraig, line 48: This comment still stands: https://github.com/iotaledger/identity.rs/pull/660/files#r809865709

I think we should just use `null`, not `undefined`, for optional return values.

cycraig (changes requested):
Looking good but there are still outstanding questions/TODOs. Some of these have already been raised and others are noted in issue #669. They are still important to address, however.

1. Split part of `identity-iota` into a separate `identity-iota-core` crate to prevent unnecessary artifact bloat for the Stronghold bindings - will be handled by @HenriqueNogara in a separate PR.
2. Address publishing packages per target in GitHub Actions. Is it part of the same Wasm workflow or separate? - https://github.com/iotaledger/identity.rs/pull/660#issuecomment-1055561386
3. Consider replacing the `Signature` struct in `identity-account` because the `public_key` it contains is currently unused - noted in #669.
4. Storage interface errors should use their own enum to avoid needing feature-gated variants in the `identity-account` error - noted in #669.
5. Switching to `Vec<u8>`/`UInt8Array` key representations in the bindings, rather than Base58 strings.
6. Remove the `bindings/stronghold-nodejs/stronghold-nodejs.linux-x64-gnu.node` artifact from git.
7. Add README.md file to Stronghold bindings explaining what they are, how to build them (build Wasm first etc.), how to test them, and a README.md example (unless we put the example in the Wasm bindings).

bindings/stronghold-nodejs/build/Dockerfile.alpine
```diff
@@ -0,0 +1,5 @@
+FROM node:16-alpine
+
+# libc6-compat for gcc compatibility and binutils for debugging
+RUN apk --no-cache add libc6-compat binutils
+RUN ln -s /lib/libc.musl-x86_64.so.1 /lib/ld-linux-x86-64.so.2
```
cycraig, line 5: Is this used within this project or only in the GitHub Actions workflow? Should we maybe move it to the `.github` directory?

bindings/stronghold-nodejs/.npmignore
```diff
@@ -0,0 +1,9 @@
+target
+Cargo.lock
+.cargo
+.github
+npm
+.eslintrc
+.prettierignore
+rustfmt.toml
+package-lock.json
```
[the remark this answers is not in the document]
eike-hass: huh, wasm bindings have both `.npmignore` and `files`. I prefer the allow list approach through `files`. I will make a separate PR to fix both bindings to only use `files`.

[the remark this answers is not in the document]
cycraig: Use ` npm pack --dry-run` to verify the contents if you can. I have no idea if `napi publish` follows the exact same procedure as `npm publish`, however.

bindings/wasm/src/account/storage/traits.rs
HenriqueNogara: The reason for this is that `napi` automatically generates this types when using `Option`

cycraig: Also, `stronghold.d.ts` incorrectly links to the wrong path from the Wasm bindings (probably why the source mapping fails). Is this a problem for publishing?

```
import { DID, KeyLocation, ChainState, IdentityState, Storage } from "../../wasm/node/identity_wasm";
```

Aside: do we need the `index.ts` file? It also has relative links that are incorrect.

eike-hass: turns out we needed a whole build step to resolve this. Went with rollup and a couple plugins to properly bundle the sources. Also added the dependency in the `@iota/identity-wasm` package, which requires linking because storage is not released yet.

bindings/stronghold-nodejs/src/error.rs
HenriqueNogara: Here's an example of error message for the `keySign` method: 
```
Invalid Proof Value: remote sign
```

bindings/stronghold-nodejs/src/account/identity/chain_state.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[napi]
+pub struct NapiChainState(pub(crate) ChainState);
+
+#[napi]
+impl NapiChainState {
+  #[napi(js_name = fromJSON)]
+  pub fn from_json(json_value: serde_json::Value) -> Result<NapiChainState> {
+    serde_json::from_value(json_value).map(Self).napi_result()
+  }
```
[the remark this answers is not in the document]
HenriqueNogara: I'll add it to #669

bindings/stronghold-nodejs/src/error.rs
cycraig: It looks like it omits the variant, it just happens to be in the error message:

```rust
  #[error("Invalid Proof Value: {0}")]
  InvalidProofValue(&'static str),
```

This may or may not be a problem based on how we choose to write error messages. The alternative is to append the error variant (which we can get via [`strum` deriving `IntoStaticStr`](https://docs.rs/strum/latest/strum/derive.IntoStaticStr.html)), similar to how it's used as the error name in the wasm bindings: https://github.com/iotaledger/identity.rs/blob/dev/bindings/wasm/src/error.rs

We might also need recursive printing of the error `source`s, like what #599 adds: https://github.com/iotaledger/identity.rs/blob/895a6fb768440cc364b2f75f3e3961dd679a597d/bindings/wasm/src/error.rs#L123

I consider improving the error messages for the Stronghold bindings optional since we can do it in a non-breaking change, this just needs to be recorded somewhere as a TODO I think.

PR #664, opened 2022-02-18
Remove `DIDLease` account feature

### LINKED ISSUE #662: [Task] Remove `DIDLease`
## Description

Remove `DIDLease` from the `Account` and the `Storage` interface.

## Motivation

The `DIDLease` was originally added to prevent two accounts from concurrently modifying the same identity. It is now causing problems in the bindings due to the Wasm memory leak problem (which is a separate issue, to be fair), but also due to leases being a storage concern. Preventing this relatively rare bug comes at the cost of every storage implementation having to provide it. This issue could be solved by not exposing the `lease_did` function to the bindings storage interface (e.g. `WasmStorage`) and implementing it on the Rust side within the bindings. This is, at least as of now, not a solution due to Wasm memory leaks.

Another option would be to solve it on an `AccountBuilder` level, which would be better than not having it at all. However, creating a second `AccountBuilder` is even easier than using a second storage - which would also work around the check - and that potentially forgoes the check entirely.

Overall, it doesn't seem like preventing this rather niche bug seems worth the headaches it causes for bindings and storage implementors. Ultimately, this is an application concern.

## To-do list

- [ ] Remove `DIDLease` from the storage implementations
- [ ] Remove `Storage:::lease_did`

## Change checklist

PhilippGackstatter:
# Description of change

Removes the `DIDLease` feature from the account. See #662 for details.

fixes #662 

## Type of change

- [ ] Documentation Fix

n/a

## Change checklist


identity-account/src/account/builder.rs
```diff
@@ -194,6 +194,9 @@ impl AccountBuilder {
 
   /// Loads an existing identity with the specified `did` using the current builder configuration.
   /// The identity must exist in the configured [`Storage`].
+  ///
+  /// Callers are expected **not** to load the same [`IotaDID`] into more than one account,
+  /// as that would cause race conditions when updating the identity.
```
olivereanderson, line 199: nit: Consider placing this under a "# Warning" header.

olivereanderson (approved):
Looks good to me. Only one minor comment.

HenriqueNogara (approved):

Issue #608, opened 2022-01-21
[Task] Upgrade to wasm-pack 0.10.X


eike-hass:
## Description
Update wasm-pack to the latest version and make sure build and tests succeed.

## Motivation
In order to stay up to date with the latest features we should upgrade to the current wasm-pack version.
An overview of the wasm-pack releases can be found here: https://github.com/rustwasm/wasm-pack/releases

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

eike-hass:
blocked by #629

cycraig:
Upgrade when https://github.com/rustwasm/wasm-pack/issues/1097 is resolved.

PR #665, opened 2022-02-18
Pin axios version

eike-hass:
# Description of change

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/bindings/wasm/package-lock.json
+++ b/bindings/wasm/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.5.0-dev.4",
       "license": "Apache-2.0",
       "dependencies": {
+        "axios": "^0.26.0",
         "node-fetch": "^2.6.7"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


cycraig (changes requested):
Not sure if this is worth it honestly: `axios` is a transient dependency of `wasm-pack`, which is just a `dev-dependency` of ours, so the security warning should not affect users of the package (as far as I know).

Is there a way to avoid falsely marking `axios` as one of our dependencies? Because with this PR it would force all users of our package to download it too, if I understand correctly.

Something like what `wasm-pack` does with `resolutions`, or perhaps a `dev-dependency` rather?

https://github.com/rustwasm/wasm-pack/blob/4ae6306570a0011246c39c8028a4f11a4236f54b/npm/package.json#L36
```
"resolutions": {
  "axios": "0.21.2"
}
```

eike-hass:
I agree that axios should not be shipped to users and this probably is the wrong approach.
We can flat out tell dependabot to ignore axios, but I think it does so no matter the dependency tree, so if we would add it later as an actual dependency it would still be ignored.
I fiddled around with overrides (https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides), but they don't seem to work for `devDependencies`. 
Should we maybe just keep the later as an reminder to update wasm-pack?

cycraig (changes requested):
We still should not add unnecessary dependencies for developers installing the `identity-wasm` package, especially not to fix a dev-dependency security warning that does not affect them.

We should rather just address security warning by bumping `follow-redirects` to 1.14.8/1.14.9 which #656 did (which is essentially what this PR is meant to address).

We already have an open issue to upgrade wasm-pack when they next release a working version for Windows #608.

eike-hass:
going with the original approach in https://github.com/iotaledger/identity.rs/pull/677

Issue #622, opened 2022-01-25
[Task] Complete WASM bindings for `IotaDocument`


abdulmth:
## Description
Write Wasm bindings in `WasmDocument` for all unported public functions in `IotaDocument`

## Motivation
Some functionalities in `IotaDocument` were introduced but not yet ported. for example `attach_method_relationship(..)`

## To-do list

The following methods are yet to be ported from `IotaDocument`:

- [ ] `is_signing_method_type` 
- [ ] `controller`
- [ ] `also_known_as`
- [ ] `properties/properties_mut` - we might need a convenient way to add/edit/remove arbitrary properties in Wasm besides using `toJSON`, editing it there, then converting back with `fromJSON`?
- [ ] `service`
- [ ] `methods`
- [ ] `attach_method_relationship`
- [ ] `detach_method_relationship`
- [ ] `resolve_method_with_scope` - maybe we can just add an optional `WasmMethodScope` parameter on `WasmDocument::resolveMethod`?
- [ ] `resolve_signing_method`

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

cycraig:
I added the differences in the To-do list. Unclear whether some were intentionally not ported if they can be accessed via JSON, e.g. `services`, `methods`, and `properties`.

Left out certain methods like `signer` (since we don't want to port `DocumentSigner` to Wasm), `core_document` (Wasm doesn't have the distinction of an `IotaDocument` versus a `CoreDocument`).

PR #679, opened 2022-02-28
Complete `Document` Wasm bindings

abdulmth:
# Description of change
Adds WASM bindings for some of the unported methods in `IotaDocument`. 

fixes issue #622

## Type of change

- [ ] Documentation Fix

Tested the methods locally.


cycraig (commented):
As discussed, this should be done on the dev branch rather than the Account bindings epic branch.

olivereanderson:
Would it be possible to include some examples that illustrate how to use this functionality in javascript?

abdulmth:
@olivereanderson do you mean examples for the library users or just for the review?

olivereanderson:
> @olivereanderson do you mean examples for the library users or just for the review?

I was thinking for the library users, but it would also be useful for the review.

abdulmth:
> I was thinking for the library users, but it would also be useful for the review.

@olivereanderson  I think we try to keep parity in the examples in Rust and WASM now and since there's no examples for that in Rust, I didn't add them to WASM.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +98,84 @@ impl WasmDocument {
...
+    self
+      .0
+      .also_known_as()
+      .iter()
+      .cloned()
+      .map(|url| url.to_string())
+      .map(JsValue::from)
+      .collect::<js_sys::Array>()
+      .unchecked_into::<DIDArray>()
+  }
```
cycraig, line 128: The `alsoKnownAs` entries as URLs, not necessarily DIDs. Change this to return an array of strings rather than `DIDArray`.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +98,84 @@ impl WasmDocument {
...
+        .map(JsValue::from)
+        .collect::<js_sys::Array>()
+        .unchecked_into::<DIDArray>(),
+      None => js_sys::Array::new().unchecked_into::<DIDArray>(),
+    }
+  }
+
+  /// Returns a set of the document's `alsoKnownAs`.
+  #[wasm_bindgen(js_name = alsoKnownAs)]
+  pub fn also_known_as(&self) -> DIDArray {
```
cycraig, line 118: The `controller` and `alsoKnownAs` fields need setters as well to be equivalent to Rust. We should also consider returning None/null in the accessors if they are not present, since that's more accurate?

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +98,84 @@ impl WasmDocument {
...
+  }
+
+  /// Adds a custom property to the DID Document.
+  /// If the Document already has this key, the value will be overwritten.
+  /// If the value is set to `null`, the property will be removed.
+  ///
+  /// ### WARNING
+  /// This method can overwrite existing properties like `id`.
+  #[wasm_bindgen(js_name = setPropertyUnchecked)]
+  pub fn set_property(&mut self, key: String, value: &JsValue) -> Result<()> {
```
cycraig, line 137: Still undecided if we should expose setting arbitrary properties, it's not really a use-case we want to encourage---the `Unchecked` suffix helps convey that at least.

Minor suggestion on doc-comment and function name.

```suggestion
  /// Adds a custom property to the DID Document.
  /// If the value is set to `null`, the custom property will be removed.
  ///
  /// ### WARNING
  /// This method can overwrite existing properties like `id` and result in an invalid document.
  #[wasm_bindgen(js_name = setPropertyUnchecked)]
  pub fn set_property_unchecked(&mut self, key: String, value: &JsValue) -> Result<()> {
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +98,84 @@ impl WasmDocument {
...
+    }
+    Ok(())
+  }
+
+  /// Returns the custom DID Document properties.
+  #[wasm_bindgen(js_name = properties)]
+  pub fn properties(&mut self) -> Result<JSMap> {
+    let properties = self.0.properties();
+    let properties_map = js_sys::Map::new();
+    for (key, value) in properties.iter() {
```
cycraig, line 155: Nit: remove unnecessary variable.

```suggestion
    let properties_map = js_sys::Map::new();
    for (key, value) in self.0.properties().iter() {
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +98,84 @@ impl WasmDocument {
...
+    Ok(properties_map.unchecked_into::<JSMap>())
+  }
+
   // ===========================================================================
   // Services
   // ===========================================================================
 
+  /// Return a set of all `Service`s in the document.
+  #[wasm_bindgen(js_name = service)]
+  pub fn service(&self) -> ArrayService {
```
cycraig, line 167: Nit: js_name is unnecessary. I'm not opposed, it just seems to be the convention that we omit it when the name matches exactly.

```suggestion
  /// Return a set of all `Service`s in the document.
  #[wasm_bindgen]
  pub fn service(&self) -> ArrayService {
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +98,84 @@ impl WasmDocument {
...
+
+  /// Returns a set of the document's `alsoKnownAs`.
+  #[wasm_bindgen(js_name = alsoKnownAs)]
+  pub fn also_known_as(&self) -> DIDArray {
+    self
+      .0
+      .also_known_as()
+      .iter()
+      .cloned()
+      .map(|url| url.to_string())
```
cycraig, line 124: Clone is unnecessary.

```suggestion
      .also_known_as()
      .iter()
      .map(|url| url.to_string())
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -107,6 +192,20 @@ impl WasmDocument {
...
 
+  /// Returns a list of all {@link VerificationMethod} in the DID Document.
+  #[wasm_bindgen(js_name = methods)]
+  pub fn methods(&self) -> ArrayMethods {
+    self
+      .0
+      .methods()
+      .collect::<Vec<&IotaVerificationMethod>>()
+      .into_iter()
+      .cloned()
```
cycraig, line 203: Nit: unnecessary `collect`.

```suggestion
      .methods()
      .cloned()
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -139,20 +238,84 @@ impl WasmDocument {
   ///
   /// Throws an error if the method is not found.
   #[wasm_bindgen(js_name = resolveMethod)]
-  pub fn resolve_method(&self, query: &UDIDUrlQuery) -> Result<WasmVerificationMethod> {
+  pub fn resolve_method(&self, query: &UDIDUrlQuery, scope: Option<WasmMethodScope>) -> Result<WasmVerificationMethod> {
```
cycraig, line 241: Aside: we should probably replace `resolve_method` and `resolve_method_with_scope` with a single `resolve_method` with an optional parameter...

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -139,20 +238,84 @@ impl WasmDocument {
...
+          .0
+          .resolve_method_with_scope(&method_query, scope.0)
+          .ok_or(identity::did::Error::MethodNotFound)
+          .wasm_result()?
+          .clone(),
+      ));
+    }
     Ok(WasmVerificationMethod(
       self.0.try_resolve_method(&method_query).wasm_result()?.clone(),
     ))
```
cycraig, line 255: Nit: this can be simplified. Optional.

We should also consider returning an option rather than an error here, or are errors preferred to handle than null in Javascript?

```suggestion
    let method: &IotaVerificationMethod = if let Some(scope) = scope {
      self
        .0
        .resolve_method_with_scope(&method_query, scope.0)
        .ok_or(identity::did::Error::MethodNotFound)
        .wasm_result()?
    } else {
      self.0.try_resolve_method(&method_query).wasm_result()?
    };
    Ok(WasmVerificationMethod(method.clone()))
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -139,20 +238,84 @@ impl WasmDocument {
...
   }
 
+  /// Attempts to resolve the given method query into a method capable of signing a document update.
+  #[wasm_bindgen(js_name = resolveSigningMethod)]
+  pub fn resolve_signing_method(&mut self, query: &UDIDUrlQuery) -> Result<WasmVerificationMethod> {
```
cycraig, line 260: Discussion: another place where we should consider returning `Option` rather than erroring? This would affect a lot of methods, so I'm okay leaving everything as fallible in this PR, to be clear.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -139,20 +238,84 @@ impl WasmDocument {
...
     let method_query: String = query.into_serde().wasm_result()?;
     let method: &mut IotaVerificationMethod = self.0.try_resolve_method_mut(&method_query).wasm_result()?;
     method.revoke_merkle_key(index).wasm_result()
   }
 
+  /// Attaches the relationship to the given method, if the method exists.
+  ///
+  /// Note: The method needs to be in the set of verification methods,
+  /// so it cannot be an embedded one.
+  #[wasm_bindgen(js_name = attachMethodRelationShips)]
```
cycraig, line 278: Nit: casing.

```suggestion
  #[wasm_bindgen(js_name = attachMethodRelationships)]
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -139,20 +238,84 @@ impl WasmDocument {
...
+  pub fn attach_method_relationships(&mut self, options: &MethodRelationshipOptions) -> Result<()> {
+    let relationships: Vec<MethodRelationship> = options
```
cycraig, line 280: Why does this use a duck-typed interface? I think it should take a `DIDUrl` parameter and a custom Typescript type (assuming we want to allow multiple relationships).

This should match the Rust API, so if we want to allow multiple relationships here we should allow it in Rust too.

E.g.
```rust
  pub fn attach_method_relationships(&mut self, did_url: &WasmDIDUrl, relationships: &UMethodRelationships) -> Result<()> {
...

#[wasm_bindgen]
extern "C" {
  #[wasm_bindgen(typescript_type = "MethodRelationship | MethodRelationship[]")]
  pub type UMethodRelationships;
}
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -139,20 +238,84 @@ impl WasmDocument {
...
+        .0
+        .attach_method_relationship(&options.did_url().0, relationship)
+        .wasm_result()?;
+    }
+    Ok(())
+  }
+
+  /// Detaches the given relationship from the given method, if the method exists.
+  #[wasm_bindgen(js_name = detachMethodRelationShips)]
+  pub fn detach_method_relationships(&mut self, options: &MethodRelationshipOptions) -> Result<()> {
```
cycraig, line 300: Nit: casing.

Also same comment as `attachMethodRelationships` regarding the parameters.

```suggestion
  /// Detaches the given relationship from the given method, if the method exists.
  #[wasm_bindgen(js_name = detachMethodRelationships)]
  pub fn detach_method_relationships(&mut self, did_url: &WasmDIDUrl, relationships: &UMethodRelationships) -> Result<()> {
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -497,3 +660,57 @@ extern "C" {
   #[wasm_bindgen(typescript_type = "DIDUrl | string")]
   pub type UDIDUrlQuery;
 }
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "DID[]")]
+  pub type DIDArray;
+}
```
cycraig, line 668: Nit: this should match current naming conventions.

```suggestion
#[wasm_bindgen]
extern "C" {
  #[wasm_bindgen(typescript_type = "DID[]")]
  pub type ArrayDID;
}
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -497,3 +660,57 @@ extern "C" {
...
+extern "C" {
+  #[wasm_bindgen(typescript_type = "VerificationMethod[]")]
+  pub type ArrayMethods;
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "Map<string, any>")]
+  pub type JSMap;
+}
```
cycraig, line 686: Nit: name suggestion, either `MapStringAny` or `WasmProperties`. I prefer the former for its verboseness when looking at the return type.

```suggestion
#[wasm_bindgen]
extern "C" {
  #[wasm_bindgen(typescript_type = "Map<string, any>")]
  pub type MapStringAny;
}
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -497,3 +660,57 @@ extern "C" {
...
+    /**
+     * The identifier of the method in the document.
+     */
+    did_url: DIDUrl,
+
+    /**
+     * The relationships to add;
+     */
+    relationships: MethodRelationship | MethodRelationship[]
+};
```
cycraig, line 715: I don't think this is necessary.

cycraig (changes requested):
Some changes requested, notably regarding `attach/detachMethodRelationships`.

bindings/wasm/src/did/wasm_document.rs
abdulmth: what do you mean? there's no `resolve_method_with_scope` here.

cycraig: Yes, I meant we should do the same thing in the Rust library proper and only have a single method.

bindings/wasm/src/did/wasm_document.rs
abdulmth: In Rust it doesn't return an `Option` so I kept the same behavior. But an empty array is not wrong here in my opinion.

bindings/wasm/src/did/wasm_document.rs
abdulmth: If having an `Option` as parameter common in Rust then that could be fine, otherwise keeping 2 methods isn't that bad IMO.

bindings/wasm/src/did/wasm_document.rs
cycraig: The `controller` returns an `Option`, just not `alsoKnownAs`. Returning an empty array should be fine, but we should allow taking `null` for the setters to unset them?

bindings/wasm/src/did/wasm_document.rs
cycraig: Indeed it is a separate discussion (i.e. not for this PR). The counter-argument is that the `Account` allows adding multiple relationships easily (at least in the Wasm bindings pending #654 for Rust) so it's fine for the Document to only allow adding relationships one-at-a-time given that it's part of the low-level API.

bindings/wasm/src/did/wasm_document.rs
cycraig: It's actually way more than 2 methods since we have fallible and mutable variants (mostly in `CoreDocument`):

- `resolve_method`
- `try_resolve_method`
- `resolve_method_mut`
- `try_resolve_method_mut`
- `resolve_method_with_scope`
- `try_resolve_method_with_scope`

I say we replace all of them with just two methods that take in an optional `MethodScope` and return `Option` instead of `Result`:

- `resolve_method`
- `resolve_method_mut`

bindings/wasm/src/did/wasm_document.rs
abdulmth: I changed it to allow only one relationship

bindings/wasm/src/did/wasm_document.rs
cycraig: Opened issue #686 for this.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -153,6 +313,39 @@ impl WasmDocument {
...
+  ///
+  /// Note: The method needs to be in the set of verification methods,
+  /// so it cannot be an embedded one.
+  #[wasm_bindgen(js_name = attachMethodRelationships)]
+  pub fn attach_method_relationships(
+    &mut self,
+    did_url: &WasmDIDUrl,
+    relationship: &MethodRelationshipType,
+  ) -> Result<()> {
+    let relationship: WasmMethodRelationship = relationship.into_serde().wasm_result()?;
```
cycraig, line 326: Take `WasmMethodRelationship` as a parameter directly.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -153,6 +313,39 @@ impl WasmDocument {
...
+  }
+
+  /// Detaches the given relationship from the given method, if the method exists.
+  #[wasm_bindgen(js_name = detachMethodRelationships)]
+  pub fn detach_method_relationships(
+    &mut self,
+    did_url: &WasmDIDUrl,
+    relationship: &MethodRelationshipType,
+  ) -> Result<()> {
+    let relationship: WasmMethodRelationship = relationship.into_serde().wasm_result()?;
```
cycraig, line 341: Take `WasmMethodRelationship` as a parameter directly.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +101,128 @@ impl WasmDocument {
...
+            for controller in controllers {
+              set.append(controller);
+            }
+            Some(OneOrSet::new_set(set).wasm_result()?)
+          }
+        }
+      }
+    } else {
+      None
+    };
```
cycraig, line 129: This can be simplified.

```suggestion
    let controller_set: Option<OneOrSet<IotaDID>> = if let Some(controllers) = controllers.map(OneOrMany::into_vec) {
      if controllers.is_empty() {
        None
      } else {
        Some(OneOrSet::try_from(OrderedSet::from_iter(controllers)).wasm_result()?)
      }
    } else {
      None
    };
```

It can probably be simplified further with `map` but this is probably fine.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +101,128 @@ impl WasmDocument {
...
+      }
+      None => {
+        self.0.properties_mut().remove(&key);
+      }
+    }
+    Ok(())
+  }
+
+  /// Returns the custom DID Document properties.
+  #[wasm_bindgen(js_name = properties)]
```
cycraig, line 199: Nit: unnecessary `js_name` (I don't mind it, just current conventions).

```suggestion
  #[wasm_bindgen]
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -107,6 +239,18 @@ impl WasmDocument {
   // Verification Methods
   // ===========================================================================
 
+  /// Returns a list of all {@link VerificationMethod} in the DID Document.
+  #[wasm_bindgen(js_name = methods)]
```
cycraig, line 243: Nit: unnecessary `js_name`.

```suggestion
  #[wasm_bindgen]
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -153,6 +313,39 @@ impl WasmDocument {
     method.revoke_merkle_key(index).wasm_result()
   }
 
+  /// Attaches the relationship to the given method, if the method exists.
+  ///
+  /// Note: The method needs to be in the set of verification methods,
+  /// so it cannot be an embedded one.
+  #[wasm_bindgen(js_name = attachMethodRelationships)]
+  pub fn attach_method_relationships(
```
cycraig, line 321: This should be singular after the latest changes.

```suggestion
  #[wasm_bindgen(js_name = attachMethodRelationship)]
  pub fn attach_method_relationship(
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -153,6 +313,39 @@ impl WasmDocument {
...
+    self
+      .0
+      .attach_method_relationship(&did_url.0, relationship.into())
+      .wasm_result()?;
+    Ok(())
+  }
+
+  /// Detaches the given relationship from the given method, if the method exists.
+  #[wasm_bindgen(js_name = detachMethodRelationships)]
+  pub fn detach_method_relationships(
```
cycraig, line 336: This should be singular after the latest changes.

```suggestion
  #[wasm_bindgen(js_name = detachMethodRelationship)]
  pub fn detach_method_relationship(
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +101,128 @@ impl WasmDocument {
...
+    };
+    *self.0.controller_mut() = controller_set;
+    Ok(())
+  }
+
+  /// Returns a list of document controllers.
+  ///
+  /// Note: Duplicates will be ignored.
+  /// Use `null` to remove all controllers.
+  #[wasm_bindgen(js_name = controller)]
```
cycraig, line 138: Nit: unnecessary `js_name`.

```suggestion
  #[wasm_bindgen]
```

cycraig (changes requested):
Minor changes.

bindings/wasm/src/did/wasm_document.rs
abdulmth: This will not work, not sure why, but maybe because it's an enum.

cycraig: I tested it in an example and it works just fine.

```rust
#[wasm_bindgen(js_name = attachMethodRelationship)]
pub fn attach_method_relationship(
  &mut self,
  did_url: &WasmDIDUrl,
  relationship: WasmMethodRelationship,
) -> Result<bool> {
  self
    .0
    .attach_method_relationship(&did_url.0, MethodRelationship::from(relationship))
    .wasm_result()
}
```

bindings/wasm/src/did/wasm_document.rs
cycraig: Also return `Result<bool>` to match the Rust function.

bindings/wasm/src/did/wasm_document.rs
abdulmth: My bad, I was confused by a different error. I fixed it.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +101,118 @@ impl WasmDocument {
...
+      None
+    };
+    *self.0.controller_mut() = controller_set;
+    Ok(())
+  }
+
+  /// Returns a list of document controllers.
+  ///
+  /// Note: Duplicates will be ignored.
+  /// Use `null` to remove all controllers.
```
cycraig, line 127: Nit: fix comment.

```suggestion
  /// Returns a list of document controllers.
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +101,118 @@ impl WasmDocument {
...
+      None => {
+        self.0.properties_mut().remove(&key);
+      }
+    }
+    Ok(())
+  }
+
+  /// Returns the custom DID Document properties.
+  #[wasm_bindgen]
+  pub fn properties(&mut self) -> Result<MapStringAny> {
```
cycraig, line 190:
```suggestion
  /// Returns a copy of the custom DID Document properties.
  #[wasm_bindgen]
  pub fn properties(&mut self) -> Result<MapStringAny> {
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -87,10 +101,118 @@ impl WasmDocument {
...
+    Ok(properties_map.unchecked_into::<MapStringAny>())
+  }
+
   // ===========================================================================
   // Services
   // ===========================================================================
 
+  /// Return a set of all `Service`s in the document.
+  #[wasm_bindgen]
+  pub fn service(&self) -> ArrayService {
```
cycraig, line 204: Is there a way for the doc-comment to link to `Service` but replace the content with `Services`? Like we do in Rust: `[Services](Service)`.

```suggestion
  /// Return a set of all {@link Service}s in the document.
  #[wasm_bindgen]
  pub fn service(&self) -> ArrayService {
```

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -496,4 +675,25 @@ impl From<IotaDocument> for WasmDocument {
...
+  pub type ArrayDID;
+
+  #[wasm_bindgen(typescript_type = "Service[]")]
+  pub type ArrayService;
+
+  #[wasm_bindgen(typescript_type = "Array<string>")]
+  pub type ArrayString;
+
+  #[wasm_bindgen(typescript_type = "VerificationMethod[]")]
+  pub type ArrayMethods;
```
cycraig, line 695: Nit: follow naming conventions.

```suggestion
  #[wasm_bindgen(typescript_type = "VerificationMethod[]")]
  pub type ArrayVerificationMethod;
```

bindings/wasm/src/did/mod.rs
```diff
@@ -21,6 +21,7 @@ mod wasm_did_url;
 mod wasm_diff_message;
 mod wasm_document;
 mod wasm_document_metadata;
+mod wasm_method_relationship;
```
cycraig, line 24: Add `pub use wasm_method_relationship::WasmMethodRelationship` in `mod.rs`.

cycraig (approved):
Looks good to me overall, thanks for the changes.

Minor comments, mostly cosmetic. Previous unresolved comments will be addressed by #686,

olivereanderson (approved):
Looks good to me!

bindings/wasm/src/did/wasm_document.rs
abdulmth: Yes, `{@link Service Services}` achieves exactly that.

Issue #683, opened 2022-03-02
[Task] Simple Duration API 


olivereanderson:
## Description
Port some of the `Duration` functionality from the `time` crate in order to conveniently increment and decrement timestamps. 
 
## Motivation
This will be useful when we introduce the new VC & VP validation API (See issue #631 and PR #599) as that allows the verifier to conveniently express earliest and latest allowed expiration and issuance dates respectively.

More precisely without this functionality a caller would need to do something like this 
```
// Do not allow credentials that expire within the next 10 hours
  let credential_validation_options: CredentialValidationOptions = CredentialValidationOptions::default()
    .earliest_expiry_date(Timestamp::from_unix(Timestamp::now_utc().to_unix() + 10 * 60 * 60)?);
```
while it ideally should be as simple as this:

```
  let credential_validation_options: CredentialValidationOptions = CredentialValidationOptions::default()
    .earliest_expiry_date(Timestamp::now_utc().try_add(Duration::hours(10))?);
```

## To-do list
Create a task-specific to-do list . 

- [ ] Implement a struct `Duration` that wraps `time::Duration` without exposing the `time` dependency. 
- [ ] Implement methods on `Timestamp` to increment it and decrement it with the given duration (make sure that RFC3339 compatibility is preserved). 
- [ ] Expose the functionality from items 1 & 2 to the wasm bindings. 

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #684, opened 2022-03-02
Feat/duration

olivereanderson:
# Description of change
Introduces a minimal API to increment and decrement a `Timestamp`. 

## Added 
New struct `Duration` representing a span of time. 

`Timestamp` now has two new methods: 

```
pub fn try_add(self, duration: Duration) -> Result<Self> ; 

pub fn try_sub(self, duration: Duration) -> Result<Self>;
```

to increment and decrement the `Timestamp` respectively. These methods return `Result` as these operations may lead to timestamps not compatible with RFC3339 which is an invariant of our `Timestamp` type.  

Closes issue #683. 

## Type of change

- [ ] Documentation Fix

Tests run locally. 

## Change checklist

- [x]  New and existing unit tests pass locally with my changes

### DIFF
--- a/identity-core/src/common/timestamp.rs
+++ b/identity-core/src/common/timestamp.rs
@@ -1,4 +1,4 @@
-// Copyright 2020-2021 IOTA Stiftung
+// Copyright 2020-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use core::convert::TryFrom;
@@ -88,24 +88,27 @@ impl Timestamp {
   }
 
   /// Computes `self + duration`
-  /// 
-  /// # Errors 
+  ///
+  /// # Errors
   /// Errors if the operation leads to a timestamp not in the valid range for [RFC 3339](https://tools.ietf.org/html/rfc3339).
-  #[must_use]
   pub fn try_add(self, duration: Duration) -> Result<Self> {
-    self.0.checked_add(duration.0).ok_or(
-      time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into()
-    ).and_then(|offset_date_time| Self::from_unix(offset_date_time.unix_timestamp()))
+    self
+      .0
+      .checked_add(duration.0)
+      .ok_or_else(|| time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into())
+      .and_then(|offset_date_time| Self::from_unix(offset_date_time.unix_timestamp()))
   }
 
   /// Computes `self - duration`
-  /// 
-  /// # Errors 
+  ///
+  /// # Errors
   /// Errors if the operation leads to a timestamp not in the valid range for [RFC 3339](https://tools.ietf.org/html/rfc3339).
   pub fn try_sub(self, duration: Duration) -> Result<Self> {
-    self.0.checked_sub(duration.0).ok_or(
-      time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into()
-    ).and_then(|offset_date_time| Self::from_unix(offset_date_time.unix_timestamp()))
+    self
+      .0
+      .checked_sub(duration.0)
+      .ok_or_else(|| time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into())
+      .and_then(|offset_date_time| Self::from_unix(offset_date_time.unix_timestamp()))
   }
 }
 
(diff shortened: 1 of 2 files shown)


identity-core/src/common/timestamp.rs
```diff
@@ -165,6 +188,49 @@ impl Diff for Timestamp {
...
+  /// A second
+  pub const SECOND: Self = Self(time::Duration::SECOND);
+  /// A minute
+  pub const MINUTE: Self = Self(time::Duration::MINUTE);
+  /// An hour
+  pub const HOUR: Self = Self(time::Duration::HOUR);
+  /// A day
+  pub const DAY: Self = Self(time::Duration::DAY);
+  /// A week
+  pub const WEEK: Self = Self(time::Duration::WEEK);
```
cycraig, line 212: Do we need to expose these constants? There's no way to compose them so they're not very useful right now. 

I.e. we do not implement arithmetic operators, so it's not possible to do something like `Duration::WEEK + 3*Duration::DAY + 5*Duration::MINUTE`. This means developers will need to convert any desired duration to the smallest unit, e.g. `Duration::minutes(7*24*60 + 3*24*60 + 5)`. Not great but at least there's a way I guess. 

We could also expose something like `Duration::new(weeks, days, hours, minutes, seconds)`. Not sure which direction to go really, but the point is that without arithmetic operations these constants are not useful in my opinion.

identity-core/src/common/timestamp.rs
```diff
@@ -165,6 +188,49 @@ impl Diff for Timestamp {
...
+
+  /// Create a new [`Duration`] with the given amount of hours.
+  pub const fn hours(hours: u32) -> Self {
+    Self(time::Duration::hours(hours as i64))
+  }
+
+  /// Create a new [`Duration`] with the given amount of weeks.
+  pub const fn weeks(weeks: u32) -> Self {
+    Self(time::Duration::weeks(weeks as i64))
+  }
```
cycraig, line 231: Missing `days()`.

identity-core/src/common/timestamp.rs
```diff
@@ -87,6 +86,30 @@ impl Timestamp {
...
+
+  /// Computes `self + duration`
+  ///
+  /// # Errors
+  /// Errors if the operation leads to a timestamp not in the valid range for [RFC 3339](https://tools.ietf.org/html/rfc3339).
+  pub fn try_add(self, duration: Duration) -> Result<Self> {
+    self
+      .0
+      .checked_add(duration.0)
+      .ok_or_else(|| time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into())
```
cycraig, line 98: Not sure if we should return an error rather than an option. On a related point, we could change the timestamp error variant to have an optional source so we can return just a message here.

E.g.
```rust
  /// Caused by attempting to parse or perform arithmetic with a `Duration` that would lead to an invalid `Timestamp`.
  #[error("Invalid Timestamp: {0}")]
  InvalidTimestamp(&'static str, #[source] Option<time::error::Error>),
```

Either way, I think the error message should be more descriptive.

```suggestion
  pub fn try_add(self, duration: Duration) -> Result<Self> {
    self
      .0
      .checked_add(duration.0)
      .ok_or_else(|| time::error::Error::Format(time::error::Format::InvalidComponent("addition exceeded maximum RFC3999 timestamp")).into())
```

identity-core/src/common/timestamp.rs
```diff
@@ -87,6 +86,30 @@ impl Timestamp {
...
+
+  /// Computes `self - duration`
+  ///
+  /// # Errors
+  /// Errors if the operation leads to a timestamp not in the valid range for [RFC 3339](https://tools.ietf.org/html/rfc3339).
+  pub fn try_sub(self, duration: Duration) -> Result<Self> {
+    self
+      .0
+      .checked_sub(duration.0)
+      .ok_or_else(|| time::error::Error::Format(time::error::Format::InvalidComponent("invalid year")).into())
```
cycraig, line 110: Same comment regarding the error message as `try_add`.

identity-core/src/common/timestamp.rs
cycraig, line 231: Nit: use `number` instead of `amount`. Optional.

```suggestion
  /// Create a new [`Duration`] with the given number of seconds.
  pub const fn seconds(seconds: u32) -> Self {
    Self(time::Duration::seconds(seconds as i64))
  }
  /// Create a new [`Duration`] with the given number of minutes.
  pub const fn minutes(minutes: u32) -> Self {
    Self(time::Duration::minutes(minutes as i64))
  }

  /// Create a new [`Duration`] with the given number of hours.
  pub const fn hours(hours: u32) -> Self {
    Self(time::Duration::hours(hours as i64))
  }

  /// Create a new [`Duration`] with the given number of weeks.
  pub const fn weeks(weeks: u32) -> Self {
    Self(time::Duration::weeks(weeks as i64))
  }
```

cycraig (changes requested):
Minor comments. I can't think of a better approach that would avoid creating our own newtype wrapper (e.g. adding a `time-0.3` feature would still lead to a breaking change, and just using the `Duration` from the std/core library would still need a wrapper in the Wasm bindings).

Overall this is good but I think we can do slightly more to make constructing arbitrary durations easier for developers, and we should consider our new error handling strategy or return an `Option` instead.

identity-core/src/common/timestamp.rs
olivereanderson: Minimal API is preferable at this point so I will simply remove them. We can add `Duration::new(weeks, days, hours, minutes, seconds)` at a later point if someone requests it.

identity-core/src/common/timestamp.rs
```diff
@@ -165,6 +188,49 @@ impl Diff for Timestamp {
...
+///
+/// This type is typically used to increment or decrement a [`Timestamp`].
+#[derive(Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct Duration(time::Duration);
+
+// copy a small subset of the functionality offered by time::Duration. By not copying everything it will be easier
+// to migrate to another implementation if that will be necessary in the future.
+// Do not allow negative Durations hence all constructors take u32 instead of i64, this makes it possible for us to
+// switch to relying on (for instance) std::time::Duration if we want to.
```
cycraig, line 201: Nit: condense comment slightly.

```suggestion
// Re-expose a small subset of time::Duration and use u32 instead of i64 
// to disallow negative durations. This gives us the flexibility to migrate
// the internal representation to e.g. std::time::Duration in the future
// if required.
```

identity-core/src/common/timestamp.rs
olivereanderson: I think returning an `Option` and changing the name to `checked_add` is more in line with how the rest of the ecosystem deals with arithmetic overflow. This also enables us to implement the `CheckedAdd` trait from [num_traits](https://docs.rs/num-traits/0.2.14/num_traits/ops/checked/trait.CheckedAdd.html) in the future.  

Moreover I expect the real world cases where one actually ends up with an RFC3339 incompatible timestamp from these operations to be extremely rare so I would prefer not to pollute the API with more (or more complicated) error conditions.

bindings/wasm/src/common/timestamp.rs
```diff
@@ -30,10 +31,69 @@ impl WasmTimestamp {
...
+
+  /// Create a new `Duration` with the given number of days.
+  pub fn days(days: u32) -> WasmDuration {
+    Self(Duration::days(days))
+  }
+
+  /// Create a new `Duration` with the given number of weeks.
+  pub fn weeks(weeks: u32) -> WasmDuration {
+    Self(Duration::weeks(weeks))
+  }
```
cycraig, line 86: I think it works regardless but we should place `#[wasm_bindgen]` annotations on exported functions.

cycraig (approved):
Looks good to me!

PR #693, opened 2022-03-07
Chore/split identity iota

HenriqueNogara:
# Description of change

## Type of change

- [ ] Documentation Fix

## Change checklist


bindings/wasm/src/did/wasm_document_metadata.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020-2021 IOTA Stiftung
```
cycraig, line 1: Nit: update licence headers to 2022 (we should really just find and replace all of these at some point now, not in this PR though).

identity-account-core/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+strum = { version = "0.21", features = ["derive"] }
+thiserror = { version = "1.0" }
+tokio = { version = "1.5", features = ["sync"] }
+zeroize = { version = "1.4" }
+
+[dependencies.iota-client]
+git = "https://github.com/iotaledger/iota.rs"
+rev = "7f5ae8f5b27f5948d8e1b4717419cc1321d862da"
+features = ["tls"]
+default-features = false
```
cycraig, line 32: We should not need `iota-client` in `identity-account-core`.

identity/Cargo.toml
```diff
@@ -31,10 +33,10 @@ harness = false
 default = ["async"]
 
 # Enables async runtime support (Tokio)
-async = ["identity-iota/async"]
+async = ["identity-iota/async", "identity-iota-core/async"]
```
cycraig, line 36: `identity-iota-core` should not need an `async` feature.

identity/Cargo.toml
```diff
@@ -31,10 +33,10 @@ harness = false
...
 
 # Enables support for secure storage of DID Documents
-account = ["identity-account"]
+account = ["identity-account", "identity-account-core"]
```
cycraig, line 39: Is there no top-level `stronghold` feature for `identity`? If not we definitely need to add one.

identity/src/lib.rs
```diff
@@ -87,14 +95,24 @@ pub mod account {
...
+  //! Storage Trait and Types definitions
+
+  pub use identity_account_core::crypto::*;
+  pub use identity_account_core::error::*;
+  pub use identity_account_core::identity::*;
+  pub use identity_account_core::storage::*;
+  pub use identity_account_core::stronghold::*;
+  pub use identity_account_core::types::*;
+  pub use identity_account_core::utils::*;
 }
```
cycraig, line 116: Discussion: I think we should re-export types from the `_core` crates under the non-core namespace (unless there are conflicts). That is, move all these `pub use identity_account_core` lines under `pub mod account`. Similarly for `identity_iota_core` and `identity_iota`.

All this does is simplify imports, which I think is desirable. The `Error` and `Result` instances are a problem, I'm not sure if we want to adopt a convention of prefixing the crate name onto those like in this PR where e.g. `IotaCoreError` is used instead of just `Error`? Not sure, need to decide and implement the convention everywhere.

identity-iota/src/diff/mod.rs
```diff
@@ -1,11 +1,4 @@
...
 
-pub use self::diff_iota_document::DiffIotaDocument;
-pub use self::diff_iota_document_metadata::DiffIotaDocumentMetadata;
-pub use self::diff_message::DiffMessage;
-
-mod diff_iota_did;
-mod diff_iota_document;
-mod diff_iota_document_metadata;
-mod diff_message;
+pub mod diff_message;
```
cycraig, line 4: This folder can be removed if we just move the trait implementation to where `TangleRef` is defined.

identity-iota/Cargo.toml
```diff
@@ -19,6 +19,7 @@ futures = { version = "0.3" }
 identity-core = { version = "=0.5.0-dev.4", path = "../identity-core" }
 identity-credential = { version = "=0.5.0-dev.4", path = "../identity-credential" }
 identity-did = { version = "=0.5.0-dev.4", path = "../identity-did" }
+identity-iota-core = { version = "=0.5.0-dev.4", path = "../identity-iota-core" }
```
cycraig, line 22: We should really be passing `default-features = false` to our own crates too and re-exposing feature flags or enabling them as-needed.

identity-iota-core/src/types/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+mod network;
+
+use iota_client::bee_message::MessageId;
+use iota_client::bee_message::MESSAGE_ID_LENGTH;
+
+static NULL: &[u8; MESSAGE_ID_LENGTH] = &[0; MESSAGE_ID_LENGTH];
+
+pub fn message_id_is_null(message_id: &MessageId) -> bool {
+  message_id.as_ref() == NULL
+}
```
cycraig, line 16: Nit: the convention in the codebase is to create a trait instead of bare functions. So here we could have, for example, a `MessageIdNull` trait to add the `is_null` function back to it.

identity-iota-core/src/error.rs
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Debug, thiserror::Error, strum::IntoStaticStr)]
+pub enum Error {
+  #[error("{0}")]
+  CoreError(#[from] identity_core::Error),
+  #[error("{0}")]
+  DiffError(#[from] identity_core::diff::Error),
+  #[error("{0}")]
+  InvalidDID(#[from] identity_did::did::DIDError),
+  #[error("{0}")]
+  InvalidDoc(#[from] identity_did::Error),
```
cycraig, line 15: This might be opportunity to implement the new error handling guidelines and add new enum variants that do not simply bubble up errors from sub-crates without context. Completely optional for this PR since it is rather bulky though and we would like to prevent merge conflicts.

identity-iota-core/src/types/mod.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+pub use network::Network;
+pub use network::NetworkName;
+
+mod network;
```
cycraig, line 7: Nit: consider moving `network.rs` from `types` to `did`. Still undecided on this since the `MessageIdNull` implementation would not have a natural home apart from `types`...

identity-iota-core/src/document/iota_document.rs
```diff
@@ -584,14 +583,14 @@ impl IotaDocument {
   ///
   /// This is the Base58-btc encoded SHA-256 digest of the hex-encoded message id.
   pub fn diff_index(message_id: &MessageId) -> Result<String> {
-    if message_id.is_null() {
+    if message_id_is_null(message_id) {
       return Err(Error::InvalidDocumentMessageId);
     }
 
-    Ok(IotaDID::encode_key(message_id.encode_hex().as_bytes()))
+    Ok(IotaDID::encode_key(message_id.to_string().as_bytes()))
```
cycraig, line 590: Looks like I was wrong and we probably need that trait in `identity_iota_core`? It seems a bit dangerous to assume the display implementation for `MessageId` will always to hex either way. Otherwise we might need a comment here?

identity-iota-core/src/diff/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+pub use self::diff_iota_document::DiffIotaDocument;
+pub use self::diff_iota_document_metadata::DiffIotaDocumentMetadata;
+pub use self::diff_message::DiffMessage;
+
+mod diff_iota_did;
+mod diff_iota_document;
+mod diff_iota_document_metadata;
+mod diff_message;
```
cycraig, line 11: Aside: consider feature-gating all the diff structs and methods in `identity-iota-core`. Optionally in this PR or needs an issue.

identity-iota-core/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+lazy_static = { version = "1.4", default-features = false }
+serde = { version = "1.0", default-features = false, features = ["std", "derive"] }
+strum = { version = "0.21", features = ["derive"] }
+thiserror = { version = "1.0", default-features = false }
+
+[dependencies.iota-client]
+git = "https://github.com/iotaledger/iota.rs"
+rev = "7f5ae8f5b27f5948d8e1b4717419cc1321d862da"
+features = ["tls"]
+default-features = false
```
cycraig, line 25: The `identity-iota-core` crate should not need `iota-client` at all. If the only reason for it is `MessageId` we can import `bee-message` directly instead, which `iota.rs` just re-exports.

identity-did/src/document/core_document.rs
```diff
@@ -842,6 +842,18 @@ where
...
+where
+  D: DID + KeyComparable + Serialize,
+  T: Serialize + std::default::Default,
+  U: Serialize,
+  V: Serialize,
+{
+  fn default() -> Self {
+    CoreDocument::builder(Default::default()).build().unwrap()
+  }
+}
```
cycraig, line 855: This panics 100% of the time because the builder lacks an `id`. Always avoid `unwrap` if possible. We do not need a default implementation here so this can be safely removed/

identity-account/src/updates/update.rs
```diff
@@ -316,7 +317,7 @@ async fn insert_method_secret(
         )
       );
 
-      store.key_insert(did, location, private_key).await
+      store.key_insert(did, location, private_key).await.map_err(|e| e.into())
```
cycraig, line 320: Nit: prefer `Into::into`.

```suggestion
      store.key_insert(did, location, private_key).await.map_err(Into::into)
```

identity-account/src/updates/update.rs
```diff
@@ -171,7 +172,7 @@ impl Update {
         let public: PublicKey = if let Some(method_private_key) = method_secret {
           insert_method_secret(storage, did, &location, type_, method_private_key).await
         } else {
-          storage.key_new(did, &location).await
+          storage.key_new(did, &location).await.map_err(|e| e.into())
```
cycraig, line 175: Nit: prefer `Into::into`.

```suggestion
          storage.key_new(did, &location).await.map_err(Into::into)
```

identity-account-core/src/error.rs
```diff
@@ -0,0 +1,75 @@
...
+  #[error(transparent)]
+  DIDError(#[from] identity_did::Error),
+  /// Caused by attempting to perform an invalid IO operation.
+  #[error(transparent)]
+  IoError(#[from] std::io::Error),
+  /// Caused by errors from the [iota_stronghold] crate.
+  #[cfg(feature = "stronghold")]
+  #[error(transparent)]
+  StrongholdError(#[from] iota_stronghold::Error),
+
```
cycraig, line 25: Many of these errors have been reworked in #691, wait until that PR is merged.

cycraig (changes requested):
I approve of the overall approach but there are still dependencies that can be removed like `iota-client` in `identity-iota-core` and we should discuss the finer-details.

1. I think we have to delay this PR until both #691 and #599 have been merged unfortunately, since they both alter many of the files moved in this PR. I do not think the merge conflicts would be easily resolvable without losing changes otherwise. 

2. We should decide whether to keep the `iota_core` and `account_core` modules in the top-level `identity` crate or re-export their types under the `iota` and `account` modules respectively. I am leaning towards re-exporting the types at the top-level for convenience to developers unless it doesn't make sense (e.g. the documentation and doc-comments still refer to those types under `iota_core`). We should probably not re-export those types in the sub-crates, however, i.e. do not re-export `identity_iota_core::IotaDocument` from` identity_iota`.
3. There are some dependencies like `iota-client` which can be removed/replaced with `bee-message`.
4. Error handling is a bit tricky since things are just bubbled up from the sub-crates now. It's probably fine to leave it as-is for this PR? Not sure.

identity/src/lib.rs
cycraig: The convenience of condensed imports may not be worth any confusion, just to be clear.

olivereanderson: Overall I agree with re-exporting types from `_core` crates under the non-core namespace in `identity`, as consumers of the main `identity` are likely not interested in how it is organized internally. This same principle also applies to errors hence I would prefer to avoid variants/names like  `IotaCoreError` if possible. 

I am less sure about how much re-exports from sub(-sub) crates to unified modules in `identity` protect us against introducing breaking changes in the external API upon further sub-division however.  The types in our public API need to be interchangeable between minor versions, but cargo may not be able to "understand" that the types are actually still the same when their origin has changed from one sub crate to another. It might be possible to help express the compatibility of types between minor versions  by using something like [the semver trick](https://github.com/dtolnay/semver-trick).

cycraig: > I would prefer to avoid variants/names like IotaCoreError if possible.

The problem is that we cannot export two `Error` enums of the same name under `identity::iota` from `identity_iota` and `identity_iota_core` without changing the name of one or the other.

If we cannot avoid breaking changes by re-exporting types then I am somewhat less in favour of re-exporting them. The semver trick is fascinating but we should hopefully not plan to use it as a crutch.

Cargo.toml
```diff
@@ -9,6 +9,8 @@ members = [
   "identity-did",
   "identity-diff",
   "identity-iota",
+  "identity-iota-core",
+  "identity-account-core",
```
PhilippGackstatter, line 13: This crate seems 95% storage-related, so can we rename this to `identity-account-storage`, or preferably, even just `identity-storage`?

cycraig: It originally was `identity-account-storage` but changed after several types there too. If all those types are related to the `Storage` I would agree with the name change then. I wouldn't choose just `identity-storage` since it really is tied to the `Account`, it's not usable/useful without it.

PhilippGackstatter: To me, it seems like almost all of it is storage-related, so the name seems fine. If we can be more specific than `core`, I think we should do it. Happy with `identity-account-storage`, too!

identity-did/src/document/core_document.rs
HenriqueNogara: I left this a placeholder just so I can find the best way to avoid a document clone in resolver.rs.

identity/Cargo.toml
cycraig: A `stronghold` feature flag was added in the epic branch, so it doesn't matter here.

identity-iota-core/src/diff/mod.rs
HenriqueNogara: issue #710

identity-iota-core/src/error.rs
cycraig: Made some improvements, defer to another PR to prevent too many changes.

identity-iota-core/src/types/mod.rs
cycraig: Moved `MessageIdExt` and `Network` to the `tangle` sub-module.

identity/Cargo.toml
cycraig: Removed dependency of `identity-iota-core` on `iota-client`, thus removing it's async feature.

cycraig (approved):
Thanks for the changes! I pushed some commits to fix the last few comments, hopefully without causing too many merge conflicts for the epic branch.

PhilippGackstatter (commented):
I know this has been merged now, but: looks good to me. Just one comment unrelated to the split.

identity-iota/Cargo.toml
```diff
@@ -48,7 +49,7 @@ features = ["blake2b"]
 
 [dev-dependencies]
 proptest = { version = "1.0.0", default-features = false, features = ["std"] }
-tokio = { version = "1.15", default-features = false, features = ["macros"] }
+tokio = { version = "1.17.0", default-features = false, features = ["macros"] }
```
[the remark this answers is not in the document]
cycraig: I'm unsure of the circumstances in which we should not update to the latest version of Rust dependencies, particularly when they are semver compatible.

In this case I found it prudent specifically because [`tokio 1.17.0` contains fixes to avoid panics](https://github.com/tokio-rs/tokio/blob/master/tokio/CHANGELOG.md#1170-february-16-2022).

I'll keep it in mind that we may not want to update all dependencies for compatibility reasons, specifically if those updates contain API-breaking changes.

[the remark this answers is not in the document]
cycraig: I have no such justification for `strum`, however. I just wanted the latest version (and forgot to update it everywhere).

PR #694, opened 2022-03-07
Enable Wasm weak references

cycraig:
# Description of change
Enable [Weak References](https://rustwasm.github.io/wasm-bindgen/reference/weak-references.html) in the Wasm bindings. This ensures that Wasm memory is freed automatically by the garbage collector, otherwise it will continue to grow unless the developer explicitly calls `.free()` on every exported instance or the instance is passed back to a function which takes ownership of it, allowing Rust to free it.

Unfortunately `wasm-pack` does not support passing the `--weak-refs` flag to `wasm-bindgen` ([issue](https://github.com/rustwasm/wasm-pack/issues/930)), so we have to invoke `cargo`, `wasm-bindgen`, and `wasm-opt` manually.

Another downside is that `wasm-bindgen` currently has a bug where parameters passed by value fail to deregister `Finalizer` handlers. This causes runtime errors from automatic garbage collection trying to free the instance when it is no longer referenced from the Javascript side, since `WasmRefCell` sees the instance as still borrowed.
See: https://github.com/rustwasm/wasm-bindgen/pull/2677

Fortunately this is easy-enough to circumvent by borrowing and cloning owned parameters instead. I added a build-time lint to prevent this problem going forward until the above PR is merged.

## Type of change

- [ ] Documentation Fix

Wasm tests and examples pass locally. I checked the memory usage locally with Chrome dev tools to confirm that the memory is freed by automatic garbage collection now.

## Change checklist


PhilippGackstatter (approved):
Looks good to me! Good idea with the lint, too.

.github/actions/publish/publish-wasm/action.yml
```diff
@@ -18,6 +18,10 @@ runs:
         override: true
         target: wasm32-unknown-unknown
 
+    # Download a pre-compiled wasm-bindgen binary.
+    - name: Install wasm-bindgen-cli
+      uses: jetli/wasm-bindgen-action@v0.1.0
```
eike-hass, line 23: for other more niche actions we pin to commit instead of version, I think we should do the same here

bindings/wasm/package.json
```diff
@@ -57,6 +57,7 @@
     "jsdoc-to-markdown": "^7.1.1",
     "mocha": "^9.2.0",
     "txm": "^8.0.1",
+    "wasm-opt": "^1.3.0",
     "wasm-pack": "0.10.1",
```
eike-hass, line 61: Since we don't use wasm-pack anymore, I assume this can be removedm

eike-hass:
other then the two minor points, LGTM! Will accept now already.
Let me know when it is ready, then we can admin merge (because of the windows runner issue).

eike-hass (approved):

.github/actions/publish/publish-wasm/action.yml
cycraig: Pinned to a commit hash.

yisibl:
As noted in the comments starting with v0.2.91 "wasm-bindgen does use the TC39 weak references proposal if support is detected. At the time of this writing all major browsers do support it."

PR #696, opened 2022-03-09
Replace Wasm `Config` with `ClientConfig` interface

### LINKED ISSUE #685: [Task] Replace Wasm Client Config with a Typescript interface
## Description
Investigate whether the `Config` class in the Wasm bindings for instantiating `Clients` can be replaced for a Typescript interface for consistency with the rest of the bindings where constructors use duck-typed interfaces instead of builders.

## Motivation
The current `Config` class in the Wasm bindings is analogous to the `ClientBuilder` except it does not expose options that take lists as parameters and it does not follow a builder pattern.

E.g. current usage:
```ts
const config = new Config();
config.setNetwork(Network.mainnet());
config.setEncoding(DIDMessageEncoding.JsonBrotli);
config.setPrimaryNode("http://127.0.0.1:14265/");
const client = await Client.fromConfig(config);
```
compared to the `ClientBuilder`

```rust
let client: Client = ClientBuilder::new()
  .network(Network::Mainnet)
  .encoding(DIDMessageEncoding::JsonBrotli)
  .primary_node("http://127.0.0.1:14265/", None, None)?
  .build()
  .await?;
```

Ideally, an `IClientConfig` interface would be more ergonomic for Typescript, if possible:

```ts
const client = await Client.fromConfig({
  network: Network.mainnet(),
  encoding: DIDMessageEncoding.JsonBrotli,
  primaryNode: "http://127.0.0.1:14265/"
});
```

## To-do list

- [x] Investigate whether a Typescript interface is feasible for all `Config`/`ClientBuilder` options, implement if so.
- [ ] Rename `Config` to `ClientConfig` for clarity if not possible.

## Change checklist

cycraig:
# Description of change
This replaces the `Config` class used in the Wasm bindings as an analogue to the `ClientBuilder` to create a `Client`. The new `IClientConfig` uses a duck-typed Typescript interface to pass configuration options, in line with other how other builders are translated to the Wasm bindings.

BEFORE:

```ts
const config = new Config();
config.setNetwork(Network.mainnet());
config.setEncoding(DIDMessageEncoding.JsonBrotli);
config.setPrimaryNode("http://127.0.0.1:14265/");
const client = Client.fromConfig(config);
```

AFTER (this PR):

```ts
const client = await Client.fromConfig({
  network: Network.mainnet(),
  encoding: DIDMessageEncoding.JsonBrotli,
  primaryNode: { url: "http://127.0.0.1:14265/" }
});
```

### Added

- Add `IClientConfig` Typescript interface.

### Changed

- Change `Client.fromConfig` to return a `Promise` rather than blocking.
- Fix `WasmNetwork` JSON serialization.

### Removed

- Remove `Config` class from Wasm bindings.
- Remove `Client::from_network`/`Client.fromNetwork`.

Fixes #685.

## Type of change

- [ ] Documentation Fix

Added unit test, updated examples. All Wasm tests and examples pass locally.

## Change checklist


bindings/wasm/src/tangle/config.rs
```diff
@@ -1,204 +0,0 @@
-// Copyright 2020-2022 IOTA Stiftung
```
HenriqueNogara, line 1: Do we want to keep this empty file?

bindings/wasm/src/tangle/client.rs
```diff
@@ -41,34 +42,22 @@ pub struct WasmClient {
...
-    Self {
-      client: Rc::new(client),
-    }
-  }
-
   /// Creates a new `Client` with default settings.
   #[wasm_bindgen(constructor)]
   pub fn new() -> Result<WasmClient> {
-    Self::from_config(&mut Config::new())
+    // TODO: is there a way to avoid blocking? Async constructors are not supported.
```
HenriqueNogara, line 48: In Napi we wrote a static method to avoid the `block_on` call. Not sure if we want this here though

HenriqueNogara (approved):

bindings/wasm/src/tangle/config.rs
cycraig: Thanks for the catch, I think a merge re-added it.

bindings/wasm/src/tangle/client.rs
cycraig: The places where this pattern occurs are legacy code from a while back; I'm not sure if we want to replace them with a static method everywhere given the convenience/expectation of constructors in our public API though.

abdulmth: Well, async constructors are possible in JS, but it's not recommended to use them since it will confuse users. Also in JS the constructor, is not supposed to do anything other than make an instance of a class. Anything else should be done in a separate method. we can do something like
```JavaScript
let client = new Client();
await client.init();
```
But this doesn't look great either. 

I personally don't think that a static method is a bad idea. But if the blocking only needs a constant time, it's not a problem IMO.

abdulmth (changes requested):
Looks good, minor questions.

bindings/wasm/src/tangle/client_config.rs
```diff
@@ -0,0 +1,345 @@
...
+struct ConfigOptions {
+  network: Option<Network>,
+  encoding: Option<WasmDIDMessageEncoding>,
+  nodes: Option<Vec<String>>,
+  primary_node: Option<NodeAuth>,
+  primary_pow_node: Option<NodeAuth>,
+  node_auth: Option<Vec<NodeAuth>>,
+  node_sync_interval: Option<u32>,
+  node_sync_disabled: Option<bool>,
+  node_pool_urls: Option<Vec<String>>,
```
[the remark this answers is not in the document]
cycraig: Thanks for the catch. I couldn't figure out a way to handle it gracefully since it's the only option which requires an `async` call, so I decided to remove it for now.

bindings/wasm/src/tangle/client.rs
cycraig: I'm of the opinion that Javascript is wrong here, building the class should make it immediately usable.

Looking at the code it doesn't seem like the `ClientBuilder.build/finish` method actually awaits anything (except in some dead, commented-out code) so blocking should be fine for now.

abdulmth (approved):
Thanks for the changes.

Issue #673, opened 2022-02-24
[Task] Add deep clone function in Wasm bindings


cycraig:
## Description
Expose a `clone` function on most Wasm bindings structs.

## Motivation
The current method of cloning objects we expose in the Wasm bindings is by using `toJSON` to serialize them, then `fromJSON` to deserialize a copy of the object.

We should instead expose a `clone` method on exported structs (that support cloning in Rust) to perform this operation. The benefits are clarity and avoiding overhead of (de)serialization to and from JSON. This will also enable cloning objects that are cloneable in Rust but not (de)serializable.

Note that `fromJSON` should still be retained as developers may use it to deserialize objects serialized with `toJSON` and stored or sent over the wire.

## To-do list

- [ ] Add `clone`.
- [ ] Update Wasm examples.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #607, opened 2022-01-20
Refactor `Document` and `ResolvedDocument` for interior mutability

cycraig:
# Description of change
This refactors `WasmDocument` to use `Rc<RefCell>` internally to reduce clones and for interior mutability when accessed from `WasmResolvedDocument`.

The `WasmDocument` API should be unchanged in terms of usage; `WasmResolvedDocument` on the other hand has several breaking changes listed below.

TODO

## Links to any relevant issues
See discussion #603.

## Type of change

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] Enhancement (a non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

## How the change has been tested
Describe the tests that you ran to verify your changes.
Make sure to provide instructions for the maintainer as well as any relevant configurations.

TODO TODO

## Change checklist

TODO TODO

- [ ] I have followed the contribution guidelines for this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #705, opened 2022-03-10
Add deep clone function in Wasm

abdulmth:
# Description of change
Expose a clone function on Wasm bindings structs.

fixes issue #673

## Type of change

- [ ] Documentation Fix


PhilippGackstatter (approved):
Looks good to me!

The only question is, whether `WasmCredentialValidationOptions` and `WasmPresentationValidationOptions` should also have the clone impl. Seems to me like they should. I think we should also add clone for `WasmNetwork`.

bindings/wasm/src/macros.rs
```diff
@@ -7,3 +7,16 @@ macro_rules! log {
...
+#[macro_export]
+macro_rules! impl_wasm_clone {
+  ($wasm_class:ident, $js_class:ident) => {
+    #[wasm_bindgen(js_class = $js_class)]
+    impl $wasm_class {
+      #[wasm_bindgen(js_name = clone)]
+      pub fn deep_clone(&self) -> $wasm_class {
+        return $wasm_class(self.0.clone());
+      }
+    }
```
cycraig, line 20: Nit: we should document that this is a deep clone.

```suggestion
    impl $wasm_class {
      /// Deep clones the object.
      #[wasm_bindgen(js_name = clone)]
      pub fn deep_clone(&self) -> $wasm_class {
        return $wasm_class(self.0.clone());
      }
    }
```

cycraig (approved):
Looks good to me. We should just take care not to implement this on structs that wrap shared `Rc` instances (not a problem unless #607 is merged really). No idea if we can programmatically prevent `impl_wasm_clone` on `Rc` wrappers at compile-time, so it may just be something we have to remember.

Issue #686, opened 2022-03-02
[Task] Combine `resolve_method` functions


cycraig:
## Description
Replace the six `*resolve_method*` functions in `IotaDocument`, `CoreDocument` and `WasmDocument` with two functions which return `Option` instead of `Result`.

## Motivation
There are currently six very similar functions on `CoreDocument` (with a subset re-exported from `IotaDocument`) for retrieving verification methods (pseudocode):

- `resolve_method(query) -> Option<&method>`
- `resolve_method_mut(query) -> Option<&mut method>`
- `try_resolve_method(query) -> Result<&method>`
- `try_resolve_method_mut(query) -> Result<&mut method>`
- `resolve_method_with_scope(query, scope) -> Option<&method>`
- `try_resolve_method_with_scope(query, scope) -> Result<&method>`

This seems unnecessary and contributes to API bloat/pollution. The pattern of having separate `try_*` functions is more for when the non-try functions panic on errors, which does not apply here. 

We should rather have only two functions which return `Option`. This moves the responsibility of handling missing methods to the caller and avoids the problem of bubbling up errors without context, which is in accordance with the new error handling guidelines.

Proposed functions (pseudocode):

- `resolve_method(query, Option<scope>) -> Option<&method>`
- `resolve_method_mut(query, Option<scope>) -> Option<&mut method>`

These are sufficient for resolving methods with/without scope and mutably/immutably. This may require far-reaching changes to error handling across the codebase, particularly in places like the `Account` where the returned option will need to be mapped to a new/different error variant.

## To-do list

- [ ] Replace the six `resolve_*` functions on `CoreDocument` and `IotaDocument` with the proposed functions.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #709, opened 2022-03-13
Combine resolve_method functions

abdulmth:
# Description of change
Replaces the six`resolve_method` functions in `IotaDocument`, `CoreDocument` and `WasmDocument` with two functions which return Option instead of Result.

- `resolve_method(query) -> Option<&method>`
- `resolve_method_mut(query) -> Option<&mut method>`
- `try_resolve_method(query) -> Result<&method>`
- `try_resolve_method_mut(query) -> Result<&mut method>`
- `resolve_method_with_scope(query, scope) -> Option<&method>`
- `try_resolve_method_with_scope(query, scope) -> Result<&method>`

with

- `resolve_method(query, Option<scope>) -> Option<&method>`
- `resolve_method_mut(query) -> Option<&mut method>`

`fixes issue #686`.

## Type of change

- [ ] Documentation Fix


identity-iota-core/src/document/iota_document.rs
```diff
@@ -293,43 +293,24 @@ impl IotaDocument {
...
-  /// and the verification relationship specified by `scope`.
-  pub fn resolve_method_with_scope<'query, Q>(&self, query: Q, scope: MethodScope) -> Option<&IotaVerificationMethod>
-  where
-    Q: Into<DIDUrlQuery<'query>>,
-  {
-    self.document.resolve_method_with_scope(query, scope)
+    self
+      .document
+      .resolve_method_mut(query)
+      .ok_or(Error::InvalidDoc(identity_did::Error::MethodNotFound))
```
cycraig, line 313: `IotaDocument::try_resolve_method_mut` should be renamed `resolve_method_mut` and return an `Option`, same as `CoreDocument`.

identity-did/src/document/core_document.rs
```diff
@@ -521,79 +521,45 @@ where
...
-    Q: Into<DIDUrlQuery<'query>>,
-  {
-    self
-      .resolve_method_with_scope(query, scope)
-      .ok_or(Error::MethodNotFound)
-  }
-
   /// Returns a mutable reference to the first [`VerificationMethod`] with an `id` property
   /// matching the provided `query`.
   pub fn resolve_method_mut<'query, Q>(&mut self, query: Q) -> Option<&mut VerificationMethod<D, U>>
```
cycraig, line 565: `CoreDocument::resolve_method_mut` should also be refactored to take `Option<MethodScope>`.

cycraig (changes requested):

identity-did/src/document/core_document.rs
```diff
@@ -56,23 +56,32 @@ pub struct CoreDocument<D = CoreDID, T = Object, U = Object, V = Object>
...
 }
 
+macro_rules! method_mut_ref {
+  ($doc:ident, $method: ident, $query: ident) => {
+    match $doc.$method.query_mut($query.into())? {
+      MethodRef::Embed(method) => Some(method),
+      MethodRef::Refer(ref did) => $doc.verification_method.query_mut(did),
+    }
+  };
+}
```
cycraig, line 83: Nit: suggest `method_ref_mut_helper` as the name, also a brief comment why this is needed would be appreciated.

```suggestion
// Workaround for lifetime issues with a mutable reference to self preventing closures from being used.
macro_rules! method_ref_mut_helper{
  ($doc:ident, $method: ident, $query: ident) => {
    match $doc.$method.query_mut($query.into())? {
      MethodRef::Embed(method) => Some(method),
      MethodRef::Refer(ref did) => $doc.verification_method.query_mut(did),
    }
  };
}
```

identity-iota-core/src/document/iota_document.rs
```diff
@@ -296,43 +296,28 @@ impl IotaDocument {
...
-  where
-    Q: Into<DIDUrlQuery<'query>>,
-  {
-    Ok(self.document.try_resolve_method(query)?)
+    self.document.resolve_method(query, scope)
   }
 
   #[doc(hidden)]
-  pub fn try_resolve_method_mut<'query, Q>(&mut self, query: Q) -> Result<&mut IotaVerificationMethod>
+  pub fn resolve_method_mut<'query, Q>(
```
cycraig, line 309: This no longer needs `#[doc(hidden)]` since developers can no longer violate unsafe restrictions using it as far as I'm aware. It needs a proper doc-comment and warning though, since changing the id fragments may violate the uniqueness of verification methods.

identity-iota-core/src/document/iota_document.rs
```diff
@@ -296,43 +296,28 @@ impl IotaDocument {
...
+    &mut self,
+    query: Q,
+    scope: Option<MethodScope>,
+  ) -> Result<&mut IotaVerificationMethod>
```
cycraig, line 313: Can we also change this to return `Option<&mut IotaVerificationMethod>` like `CoreDocument` and `resolve_method` do?

cycraig (changes requested):
Looking good, thanks for the changes.

Please update the Wasm bindings `WasmDocument.resolveMethod` function to match, returning `Result<Option<WasmVerificationMethod>>`.

I think we should rename `try_resolve_signing_method` to just `resolve_signing_method` for consistency.

We may also want to update `remove_method` to return a `bool` instead of erroring to be consistent too. Up to you if we want it in this PR or to open another issue for it.

abdulmth:
@cycraig I've implemented the suggestions. Regarding `remove_method` we might wanna keep it as `Result` since `insert_method` has the same signature. Also returning a boolean could encourage not dealing with the case where removing a method fails which could indicate a bug in the code. And that is different from `resolve_method` since the return value is the reason for calling the method. 
If we decide to do it anyways, it would be good to create an PR for it since it's used in many places and that would make the review easier.

cycraig:
> Regarding `remove_method` we might wanna keep it as `Result` since `insert_method` has the same signature.
Yeah it's more consistent then but we could change both. I believe the thinking was that failing to insert a method should be an error as should trying to remove a non-existent method, because they deal with public keys and affect whether signatures are revoked etc. I think it would be fine to return a bool since developers should check on their end. Either way, not for this PR.

bindings/wasm/src/did/wasm_document.rs
```diff
@@ -265,28 +265,31 @@ impl WasmDocument {
...
+      self.0.resolve_method(&method_query, Some(scope))
     } else {
-      self.0.try_resolve_method(&method_query).wasm_result()?
+      self.0.resolve_method(&method_query, None)
     };
-    Ok(WasmVerificationMethod(method.clone()))
+    match method {
+      None => Ok(None),
+      Some(method) => Ok(Some(WasmVerificationMethod(method.clone()))),
+    }
```
cycraig, line 284: Nit: can be condensed somewhat. Optional opinion.

```suggestion
    Ok(method.map(|method| WasmVerificationMethod(method.clone())))
```

cycraig (approved):

Issue #714, opened 2022-03-15
[Task] Move `proof` in `IotaDocument` outside of metadata


cycraig:
## Description
Move the `proof` section in `IotaDocument` from inside its `metadata` to the root level. 

Current JSON:
```json
"doc": {},
"meta": { 
  "proof": {},
  ... 
}
```

Proposed JSON:
```json
"doc": {},
"meta": {},
"proof": {}
```

## Motivation
This change would remove the special case for signing DID Documents, which stands in contrast to all other structs which have their proof at the root level, such as `Credential` and `Presentation`.

While in Rust this is handled by the `SetSignature` trait, this discrepancy has caused problems in the Wasm bindings and will likely continue to cause problems. This also causes problems when verifying data, because we need to know what type of data we're dealing with in order to know where its signature is located.

The simplest solution is to move the DID Document `proof` to align with other objects.

## To-do list

- [ ] Move `proof` in `IotaDocument`.
- [ ] Update the IOTA DID Method Specification.

## Change checklist

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

eike-hass:
We need to verify this change against the explorer application. It expects only the the `doc` and `meta` root keys. I assume ATM you would not be able to inspect the proof in the UI with this change.

cycraig:
I believe the `proof` field was manually removed from the JSON in the initial release of the explorer, if I'm not mistaken. Was it added back in the refactor for 0.5? Needs investigating either way, thanks for the heads-up.

PR #728, opened 2022-03-17
Move DID Document proof outside metadata

abdulmth:
# Description of change
Move the `proof` section in `IotaDocument` from inside its `metadata` to the root level. 

Previous JSON:
```json
"doc": {},
"meta": { 
  "proof": {},
  ... 
}
```

This PR:
```json
"doc": {},
"meta": {},
"proof": {}
```
fixes issue #714.

## Type of change

- [ ] Documentation Fix


cycraig (changes requested):
The IOTA DID Method specification in the Wiki needs to be updated with the change, including its examples (as per the issue).

abdulmth:
@cycraig yes, that's on my list, I opened this PR for review just to see if CI passes.

identity-iota-core/src/diff/diff_iota_document.rs
```diff
@@ -55,7 +55,7 @@ impl Diff for IotaDocument {
       .transpose()?
       .unwrap_or_else(|| self.metadata.clone());
 
-    Ok(IotaDocument::from((document, metadata)))
+    Ok(IotaDocument::from((document, metadata, None)))
```
cycraig, line 58: It might be worth adding the same comment that was in `DiffIotaDocumentMetadata` that the proof is intentionally excluded from the diff.

`// NOTE: proof intentionally excluded.`

identity-iota-core/src/diff/diff_iota_document.rs
```diff
@@ -71,7 +71,7 @@ impl Diff for IotaDocument {
       .transpose()?
       .ok_or_else(|| Error::convert("Missing field `metadata`"))?;
 
-    Ok(IotaDocument::from((document, metadata)))
+    Ok(IotaDocument::from((document, metadata, None)))
```
cycraig, line 74: Same comment.

`// NOTE: proof intentionally excluded.`

cycraig (changes requested):
Looks good. Only change required would be creating a DID Document with a real signature for the specification example.

cycraig (commented):
Oh, should we remove the `WasmDocument.signDocument` function? Its only purpose was to enable signing other documents in Wasm since the proof location was different, it doesn't even perform any extra checks like `signSelf` does.

Also please check that signing and verifying documents in Wasm works as expected (i.e. something like `doc2Signed = doc1.sign(doc2); doc1.verifyDocument(doc2Signed);`).

abdulmth:
It's still good to keep the `signDocument` method since it takes a `Document` as parameter, which prevents the need of using `fromJSON` and `toJSON` calls if we would use `signData` instead.

PhilippGackstatter (approved):
Looks good to me!

cycraig (approved):

Issue #567, opened 2021-12-20
[Request] List all DIDs in an Account Storage


cycraig:
## Description
Add a method to list all DIDs stored in an `Account` `Storage` instance.

## Motivation
Prior to refactoring the `Account` to manage a single IOTA DID document, there was an `Account::list_identities` method to list all DIDs stored locally. A single Stronghold, for example, can contain multiple identities managed by separate `Accounts`, so it might still be useful to list which DIDs are stored in a local Stronghold file. The list should still be encrypted and require a password to access the snapshot so there should be no security risk in adding this.

## Requirements
Write a list of what you want this feature to do.

1. List the `IotaDIDs` of all DID documents in a `Storage` or `Stronghold` instance.

## Open questions (optional)

Still need to investigate if placing this method on the `Storage` interface or just the `Stronghold` struct is best, as requiring this for e.g. remote `Storage` implementations might not always make sense or even be possible.

We may also decide that this is unnecessary and that users should continue to manage lists of identities as they see fit rather than forcing a particular method.

## Are you planning to do it yourself in a pull request?

Probably not.

PhilippGackstatter:
After thinking about it some more, I believe listing identities was removed because by itself, it is not all that useful. A DID is just a random identifier, so getting a list of them is not very helpful, unless I'm missing something? Instead we expect applications to ascribe meaning to each DID, such as having profiles that map to some DID.

So maybe we should enable users to store that mapping by exposing a generic key-value store via the `Storage` trait. That would make the trait more useful, in general, for all sorts of things users could store. We have to be careful with that, though, because that might give the impression that the storage can be treated like a database. But since our default storage is stronghold, and that is always completely in memory, it is not very suitable for storing large amounts of data. That should be made clear. For the mapping, it's more than fine.

cycraig:
Well currently I believe there is no way of retrieving which DIDs are stored in a Stronghold file unless you have prior knowledge of them. This might be useful for backup restoration where one does not know which DIDs were stored in one's old Stronghold for instance, otherwise we expect users to devise their own backup strategies and store the DID index themselves too.

We expect applications to ascribe extra information to each DID but the question is whether that information needs to be secured/encrypted to the same degree as the private keys for instance. I think allowing the interface to treat a storage as a generic key-value database would lead to bloating the stronghold file significantly and thus degrade performance, which is not desirable.

PR #729, opened 2022-03-17
Refactor `KeyLocation`

### LINKED ISSUE #508: [Task] Derive `KeyLocation` from inherent fields of methods

PhilippGackstatter:
# Description of change

Derive the `KeyLocation`s only from a DID document, no longer requiring any additional state to be kept.

The motivation for this PR is to remove the `IdentityState`, which is only required because it stores where a key for a given verification method is stored.

See #508 for more details.

### Circular Dependency

If we want to use the DID as the storage's partition key (i.e. the top-level identifier to divide storage), then, in order to generate a key from which the DID is derived, we need the DID itself...

We already work around this same problem on a deeper level, that of `KeyLocation`s. Here, we also need to generate a key for a verification method, but the `KeyLocation::new` function requires the `MethodData` of that verification method, so another circle. Here, we can just use a random `KeyLocation` as a temporary location, and then move the key to the actual location after having created the verification method. Why can't we do this for the initial DID generation? Well, we can only (easily) move a key _within_ a vault of a Stronghold client (identified by the partition key). All that happens in the same client. And, again, in order to generate the key for a DID, we need to use some client identified by that same DID. So the issue is really only the initial DID generation, every other operation would work perfectly fine with the DID as the partition key. How do we resolve the circular dependency, then?

1. Introduce some identifier `AccountId` (like the `IdentityId` we once had), and use that to identify everything in Storage rather than DID. Basically, the "every problem in computer science can be solved with an additional level of indirection"-solution. This means keeping some small additional state (one `Uuid` per account - a counter design like `IdentityId` would require synchronization), but it's not as clean as only having the DID for everything, so not great, not terrible.
2. Use one client for all accounts, i.e. hardcode the client identifier. Each client has multiple vaults, so we could still partition by DID within a client. Since we can easily move keys within a client's vaults now, we can generate a key at some random location, derive the DID and then move the key to its actual location. With this, we don't take advantage of stronghold as much as with client-partitioning. It's not unlikely that the single client would be a bottleneck as all stronghold operations have to go through this single client, even though two different accounts could theoretically operate concurrently without affecting each other. The stronghold team suggested that this would most likely perform less well than multiple clients. However, no conclusive benchmark was done.
3. Use the stronghold sync feature, which allows synchronizing clients, and therefore also moving keys across clients. This is similar to the move functionality described in 2, except also across clients. This would solve the problem described in the introduction. However, this would still require adding a method `Storage::did_create` to the interface, which requires every single storage implementations to generate a key, create a did, and return the did and its `KeyLocation`. This method is required, because other interface functions take `DID` and thus require a DID to already be generated before calling them. Not using this method would complicate the Storage interface further, e.g. by also requiring a `key_move` function.
4. Generate the initial KeyPair in memory, derive the DID and insert it. Since the initial key pair has a high-value due to the DID being derived from it and it representing the initial verification method, I don't think this is a proper solution.

4 is the current solution on `dev`, and was always intended to be temporary.

While Option 1 was the previously implemented option, the additional `AccountId` seemed like an unclean solution. Option 3 is implemented now, because requiring storage implementations to call `IotaDID::new` is fairly straightforward, even though it still leaves a possibility of screwing this process up. Given that implementing a `Storage` is a very advanced feature already, since it requires correct implementations of cryptographic operations, this still seems okay.
This option still adds an index to the storage interface, in order to implement #567 easily in the future.

### Temporary Stronghold solution

The now implemented stronghold solution is temporarily using a `KeyPair` to generate the private key, if none was provided through `Storage::did_create`. This is because the required sync feature is not in a stable version of stronghold yet, and we would like to go ahead with a release. We should be able to change the implementation in a future update without breaking the interface nor the stronghold file layout.

### Other noteworthy decisions

- Using a non-cryptographically secure 64-bit hash for storage seems perfectly fine. Deriving the key location by itself is not an information leak, and maliciously crafted collisions would overwrite keys, but they still require access to the stronghold in the first place. To improve on collision resistance (and, I guess, human readability), the `KeyLocation` additionally includes a method's fragment.
- We're using `seahash` because the only (as far as I'm aware) hasher in `std` - which is `std::collections::hash_map::DefaultHasher` - does not have a stable output. 
  - "The internal algorithm is not specified, and so it and its hashes should not be relied upon over releases."
  - Alternatives to `seahash` would probably be `siphasher` or `highway`, thanks @olivereanderson.
- For stronghold storage, use a client per account (identified by `did.to_string()`), a fixed vault path `$vault`, and `KeyLocation.to_string()` as the record path. Thus, in each client, there is a single vault, containing all keys (=records).

## Added

- `Storage::index_has`, which returns whether a given `IotaDID` is present in the storage.
- `Storage::index`, which returns all contained `IotaDID`s. This can be used easily to implement #567 in a future PR.
- `Storage::did_create` to create the initial did from a provided private key, or by letting the storage generate one.

## Changed

- `IdentityState::sign_data` is now `remote_sign_data`.
  - An extension trait for `IotaDocument` would've been nicer, but due to `async_trait` introducing an explicit `Send` or `?Send` bound - as opposed to letting the compiler figure it out - this didn't work.
- Renamed `key_new` to `key_generate` to be more specific.
- Renamed `key_get` to `key_public` to be clearer for implementers.
- Renamed `key_del` to `key_delete` since we're generally verbose now.
- Rename all methods of the interface to follow the `$noun_$verb` template (except for `flush_changes`).
  - `Storage::state` -> `Storage::document`.
  - `Storage::set_state` -> `Storage::set_document`.
  - `Storage::set_chain_state` -> `Storage::chain_state_set`.
  - `Storage::chain_state` -> `Storage::chain_state_get`.
- Documentation to `Storage` methods and the type itself to be more clear about the expected behavior of methods.
- Rename `Storage::purge` -> `Storage::did_purge` to indicate that it is, in a way, the destructor for `did_create`.
- Rename `Account::fetch_state` to `Account::fetch_document`.

## Removed

- `IdentityState`.
- `Generation`.

This PR does not fully implement `Storage::purge` for `Stronghold`. It leaves key deletion for a later PR, since it is a more complicated issue than anticipated. #757 was opened to track that issue.

fixes #508

## Type of change

- [ ] Documentation Fix

- Existing account tests were adjusted and some were expanded, to include a test with `Stronghold`, where it made sense. This, has not been added permanently, though, for reasons outlined below.
- Additional tests were added for the index and the key move feature.

### Testing Stronghold

There is _some_ problem with creating new `Stronghold` instances and using them concurrently, as it happens in tests. Tests would actually run in CI if we were to pass the `--test-threads=1` flag when testing the `identity-account` package, but that is probably not easily possible and not worth spending time on, imo. Most likely, there is some problem in our wrapper, which uses a static singleton to talk with the actual stronghold and only fake `Stronghold` instances. I didn't look into this any deeper, because we're going to replace this very soon, anyway. (And for an increasing number of good reasons, it looks like.) As mentioned above, the new functionality was tested locally on stronghold, and it would be appreciated if reviewers would run this locally as well, to confirm.

To test locally, make some changes to `identity-account/src/tests/util.rs`. Add/modify:

```rust
pub(super) static TEST_STRONGHOLD: once_cell::sync::Lazy<Arc<dyn Storage>> = once_cell::sync::Lazy::new(|| {
  let temp_file = _temporary_random_path();
  let stronghold: Stronghold = futures::executor::block_on(Stronghold::new(&temp_file, "password", None)).unwrap();
  Arc::new(stronghold)
});

// ...

pub(super) async fn storages() -> [Arc<dyn Storage>; 2] {
  [Arc::new(MemStore::new()), Arc::clone(&TEST_STRONGHOLD)]
}
```

and run with `cargo test -p identity-account -- --test-threads=1`.

## Change checklist


eike-hass:
Could you also take a look at how this changes affects the typescript MemStore implementation in `/bindings/wasm/examples-account/src/memory_storage.ts`?

PhilippGackstatter:
> Could you also take a look at how this changes affects the typescript MemStore implementation in `/bindings/wasm/examples-account/src/memory_storage.ts`?

Any specific concern? I will apply the changes from the Rust refactor to Wasm once the Rust parts are polished, so I will definitely touch the TS `MemStore` in that process.

PhilippGackstatter:
I'm marking this as ready for review in order for tests to run. I consider the Rust side of the PR done, while I haven't touched Wasm or Napi yet, because making changes to both Rust and bindings would be more work than nailing down the Rust version first.

identity-account/Cargo.toml
```diff
@@ -30,6 +30,7 @@ features = ["blake2b", "ed25519", "hmac", "pbkdf", "sha", "slip10", "std", "x255
 
 [dev-dependencies]
 futures = { version = "0.3" }
+once_cell = "1.10"
```
cycraig, line 33:
```suggestion
once_cell = { version = "1.10", default-features = false, features = ["std"] }
```

identity-account-storage/src/types/key_location.rs
```diff
@@ -63,3 +67,30 @@ impl Debug for KeyLocation {
...
+  }
+}
+
+pub fn method_to_key_type(method_type: MethodType) -> KeyType {
+  match method_type {
+    MethodType::Ed25519VerificationKey2018 => KeyType::Ed25519,
+    MethodType::X25519KeyAgreementKey2019 => KeyType::X25519,
+    MethodType::MerkleKeyCollection2021 => todo!(),
+  }
+}
```
cycraig, line 96: Try avoid exporting public functions directly, this is only used in two places (outside of tests) so it's probably fine to inline. Also this wouldn't work in general for e.g. `JsonWebKey2020` which can hold any sort of key, so I don't think we can support a stable function like this.

identity-account-storage/src/types/key_location.rs
```diff
@@ -5,56 +5,60 @@ use core::fmt::Debug;
...
+      key_hash,
+    }
   }
 }
 
 impl Display for KeyLocation {
   fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-    f.write_fmt(format_args!("({}:{}:{})", self.generation, self.fragment, self.method))
+    f.write_fmt(format_args!("({}:{})", self.fragment, self.key_hash))
   }
```
cycraig, line 62: Fun fact: one can have multiple methods in a DID Document with the same fragment but different fragments (could even be the same hash come to think of it). Not sure if this is sufficient but I don't have a better suggestion.

identity-account-storage/src/types/key_location.rs
```diff
@@ -5,56 +5,60 @@ use core::fmt::Debug;
...
+#[derive(Clone, Hash, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+#[non_exhaustive]
 pub struct KeyLocation {
-  method: MethodType,
-  fragment: Fragment,
-  generation: Generation,
+  pub key_type: KeyType,
+  pub fragment: String,
+  pub key_hash: u64,
 }
```
cycraig, line 26: We may want to keep/refactor `Fragment` since a `String` alone does not satisfy the constraints on DIDUrl fragments. Probably fine to use `String` for now though.

identity-account-storage/src/stronghold/mod.rs
```diff
@@ -4,7 +4,8 @@
 mod context;
 mod error;
 mod hint;
-mod records;
+// TODO: Remove or refactor for key deletion.
+//mod records;
 mod snapshot;
```
cycraig, line 9: TODO

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +28,82 @@ mod storage_sub_trait {
...
+  ///
+  /// Note that his operation needs to be synchronized globally for a given storage instance,
+  /// in order to prevent race conditions when inserting index entries concurrently.
+  async fn index_set(&self, did: IotaDID, account_id: AccountId) -> Result<()>;
+
+  /// Looks up the [`AccountId`] of the given `did`, returning `None` if no mapping exists.
+  async fn index_get(&self, did: &IotaDID) -> Result<Option<AccountId>>;
+
+  /// Retrieves the list of [`IotaDID`]s stored in the index.
+  async fn index(&self) -> Result<Vec<IotaDID>>;
```
cycraig, line 105: Shouldn't this return a map of `IotaDID`s to `AccountId`s? Maybe `index_keys` would make more sense?

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +28,82 @@ mod storage_sub_trait {
...
 
-  /// Returns the state of the identity specified by `did`.
-  async fn state(&self, did: &IotaDID) -> Result<Option<IdentityState>>;
+  /// Returns the [`IotaDocument`] of the identity specified by `account_id`.
+  async fn document(&self, account_id: &AccountId) -> Result<Option<IotaDocument>>;
 
-  /// Sets a new state for the identity specified by `did`.
-  async fn set_state(&self, did: &IotaDID, state: &IdentityState) -> Result<()>;
+  /// Sets a new state for the identity specified by `account_id`.
+  async fn set_document(&self, account_id: &AccountId, state: &IotaDocument) -> Result<()>;
```
cycraig, line 88: What naming scheme are we following in the `Storage`? We have `noun_verb` elsewhere like `key_insert`, `key_sign`, `index_set`, `index_get` but here we have `verb_noun` for `set_chain_state` and `set_document`. Should we be consistent?

identity-account-storage/src/storage/memstore.rs
```diff
@@ -177,51 +174,78 @@ impl Storage for MemStore {
...
+    if let Some(account_id) = self.index_get(did).await? {
+      let _ = self.documents.write()?.remove(&account_id);
+      let _ = self.vaults.write()?.remove(&account_id);
+      let _ = self.chain_states.write()?.remove(&account_id);
+
+      self.index.write()?.remove(did);
+    }
+
+    Ok(())
+  }
```
cycraig, line 216: It's often useful to know if a command like this did anything, even if it's idempotent. Maybe return a `bool`?

identity-account-storage/src/storage/memstore.rs
```diff
@@ -124,51 +108,64 @@ impl Storage for MemStore {
...
-  async fn key_del(&self, did: &IotaDID, location: &KeyLocation) -> Result<()> {
+  async fn key_del(&self, account_id: &AccountId, location: &KeyLocation) -> Result<()> {
     let mut vaults: RwLockWriteGuard<'_, _> = self.vaults.write()?;
-    let vault: &mut MemVault = vaults.get_mut(did).ok_or(Error::KeyVaultNotFound)?;
+    let vault: &mut MemVault = vaults.get_mut(account_id).ok_or(Error::KeyVaultNotFound)?;
 
     vault.remove(location);
 
     Ok(())
   }
```
cycraig, line 160: It's often useful to know if a command like this did anything, even if it's idempotent. Maybe return a `bool` if possible?

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -79,138 +87,192 @@ impl Stronghold {
...
+    };
+
+    // Remove index entry.
+    let store: Store<'_> = self.store(INDEX_PATH);
+    let index_lock: RwLockWriteGuard<'_, _> = self.index_lock.write().await;
+    let mut index: HashMap<IotaDID, AccountId> = get_index(&store).await?;
+    index.remove(did);
+    set_index(&store, index).await?;
+    // Explicitly release the lock early.
+    std::mem::drop(index_lock);
```
cycraig, line 214: This might need double-checked locking? (Not sure if that's an anti-pattern). Since a second instance calling `purge` and waiting on the lock would proceed with the deletion since it doesn't know it's been/is being deleted already? Checking the return of `index.remove` would address that edge-case.

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +28,82 @@ mod storage_sub_trait {
...
+/// as well as key-value like storage of data structures, such as DID documents.
+///
+/// This interface works with [`AccountId`] and [`IotaDID`] as the top-level identifiers.
+/// `AccountId` is intended as a partition-key, i.e. everything related to an account
+/// can be stored in its own partition. Keys belonging to an account are identified by [`KeyLocation`]s
+/// in that partition.
+/// An `IotaDID` can be resolved to an `AccountId` using the index, which is a global data structure.
+/// Therefore, index operations need to be carefully synchronized per [`Storage`] instance, while other operations
+/// only need to be synchronized within the same `AccountId`.
+/// For example, two `key_generate` operations with different account ids can be executed concurrently.
```
PhilippGackstatter, line 43: This is actually incorrect, we don't expect `Storage` to synchronize per account. I will reword this.

cycraig (commented):
I'm tempted to say that the circular dependency is a `Stronghold`-specific problem rather than one affecting all `Storage` implementers, but I'm not sure that's true. Obviously the current "solution" of deriving `KeyPair` in memory, bypassing `Stronghold's` security is not acceptable, so this at least addresses that.

> `Storage` methods take `AccountId` by reference rather than a copy.

That's fine and probably preferred. We might refactor it to something like a `String` that isn't `Copy` one day so a reference is preferred actually.

> the `KeyLocation` additionally includes a method's fragment.

I think I mentioned in a comment that there can be duplicate fragments within a DID Document if the DID part of the DID-Url is different.

identity-account-storage/src/types/account_id.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use uuid::Uuid;
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash, serde::Serialize, serde::Deserialize)]
+pub struct AccountId(Uuid);
+
+impl AccountId {
+  /// Creates a new random identifier.
+  pub fn random() -> Self {
+    Self(Uuid::new_v4())
```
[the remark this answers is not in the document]
PhilippGackstatter: [`ksuid`](https://docs.rs/ksuid/latest/ksuid/) might be an alternative. It's 20 bytes instead of 16 like `Uuid` or `Ulid`, but also provides that timestamp. I considered it, but couldn't come up with a reason for needing the timestamp. Do we need to offer an API to get an account's creation time? Or do we otherwise need to sort these ids?

identity-account/src/updates/update.rs
```diff
@@ -162,41 +150,49 @@ impl Update {
...
 
-        state.document_mut().insert_method(method, scope)?;
+        // The key location must be available.
+        ensure!(
+          !storage.key_exists(&account_id, &location).await?,
+          UpdateError::DuplicateKeyLocation(location)
+        );
+
+        // Move the key from the tmp to the expected location.
+        storage.key_move(&account_id, &tmp_location, &location).await?;
```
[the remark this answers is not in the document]
PhilippGackstatter: That is a general problem. If any of the operations between `key_generate` and `key_move` fail, they bail and leave the key in storage. But most of these are unlikely or impossible to fail unless we introduce a bug. I handled it explicitly for this case, deleting the temporary key if `location` is occupied before returning the error.

identity-account/src/updates/update.rs
```diff
@@ -61,56 +58,41 @@ pub(crate) async fn create_identity(
...
-    }
-    _ => {
-      return Err(UpdateError::InvalidMethodSecret("expected None or Ed25519 private key".to_owned()).into());
-    }
-  };
+  if let Some(inner_method_secret @ MethodSecret::Ed25519(_)) = setup.method_secret {
+    insert_method_secret(store, account_id, &tmp_location, setup.key_type, inner_method_secret).await?;
+  } else {
+    store.key_generate(&account_id, &tmp_location).await?;
+  }
```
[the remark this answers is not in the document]
PhilippGackstatter: This case is technically handled a few lines above with:

```rust
// The method type must be able to sign document updates.
  ensure!(
    IotaDocument::is_signing_method_type(method_type),
    UpdateError::InvalidMethodType(method_type)
  );
```

Would you prefer handling the case of other `MethodSecret`s explicitly anyway?

identity-account-storage/src/storage/traits.rs
PhilippGackstatter: Should be handled with #757.

identity-account-storage/src/types/account_id.rs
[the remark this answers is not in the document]
cycraig: Not so concerned about the timestamp, more that the ids are sortable/monotonically increasing, since that property may be quite useful for `Storage` implementers (very useful for database performance etc. and could simplify their designs). I actually wondered if the creation timestamp would be a security issue, so it was not the reason for suggesting `ulid`.

I would rather err on the side of `uuid` which is more widely supported and understood rather than use something like `ksuid` just to have the `AccountId` have that property though.

identity-account/src/updates/update.rs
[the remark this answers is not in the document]
cycraig: That case is not handled. You can reproduce the problem by passing `MethodType::Ed25519` with `MethodSecret::X25519`, it just generates a new key and ignore the `X25519` private key (I ran into that case myself).

Probably made irrelevant with #764 since I removed `MethodSecret` here altogether.

identity-account-storage/src/storage/traits.rs
cycraig: Not really sure how key deletion affects the `AccountId` index and `index_keys`?

PhilippGackstatter: Ah, sorry! I meant to post this on the comment above. I will address the `index_keys` comment.

identity-account-storage/src/stronghold/mod.rs
PhilippGackstatter: Should be handled with https://github.com/iotaledger/identity.rs/issues/757, where we might want to use the `Records` or go with another solution to keep track of keys. There we can make a final decision on whether to remove or use it.

identity-account-storage/src/storage/traits.rs
PhilippGackstatter: To address the actual comment: I felt like a vector is easier to construct for storage and it should also be faster than copying dids and account ids from storage into a hashmap. I don't think we ever need to return a `HashMap<IotaDID, AccountId>` to users, because `AccountId` is a private type. Thus, a list of DIDs should be good enough? For internal lookups, `index_get` should be used instead of (potentially wastefully) creating the hash map and using that to do the lookup.
For both our storage implementations returning the entire hash map would be trivial, but that's due to a rather wasteful implementation (specifically in stronghold), but that doesn't need to hold true for other implementations.

identity-account-storage/src/storage/traits.rs
PhilippGackstatter: I would make the state data structure consistent with the rest, so e.g. `document_set`. Should we even use `chain_state_get` and `document_get`, that feels maximally consistent?

cycraig: Yeah I would change all functions in `Storage` to be `noun_verb` in then, as long as it's internally consistent, even though it's inconsistent with the rest of the project. Rust and Wasm function names should always match as much as possible.

identity-account-storage/src/types/key_location.rs
PhilippGackstatter: If it's the same key hash, which is based on the public key, then it must be the same key material in storage, right? Then that only complicates key deletion. If we remove a verification method and mark the key in storage as deletable, then we also need to handle the case where the same key might be referenced by another verification method. It only makes that more complicated, though, it should still possible to figure out whether a key is still needed.

identity-account-storage/src/types/key_location.rs
PhilippGackstatter: Perhaps we can use `String` in the public interface for `KeyLocation`, but use `Fragment` internally, only to be able to consistently return the fragment without the leading `#`, independent of what the caller passed in? The `fragment` field would be made private, and a `KeyLocation::fragment` getter would be added, returning `Fragment::name`. `KeyLocation::fmt` would use that function then.

identity-account-storage/src/storage/stronghold.rs
PhilippGackstatter: Thanks for catching that! Added a check on the return value of `index.remove`.

identity-account/src/account/account.rs
```diff
@@ -48,12 +50,13 @@ pub struct Account<C = Arc<Client>>
 where
   C: SharedPtr<Client>,
 {
+  pub(crate) account_id: AccountId,
```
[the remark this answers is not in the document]
PhilippGackstatter: Since `AccountId` is just 16 bytes, I figured it would be much more efficient to look it up once during `Account::load_identity`, and then just read the field for any future `Storage::*` calls.

  Making Storage look up the account id of a DID for every call results in a less consistent / more confusing interface than just passing `AccountId` everywhere, I think. `Storage::{key_generate, key_insert}` still need to take the `AccountId` while all others would take `IotaDID`. That is, unless we make `Storage` generate the `DID`. Also the lookup also costs a bit of performance every time, though it probably doesn't matter too much.

identity-account-storage/Cargo.toml
```diff
@@ -20,10 +20,14 @@ identity-did = { version = "=0.5.0-dev.5", path = "../identity-did", default-fea
 identity-iota-core = { version = "=0.5.0-dev.5", path = "../identity-iota-core", default-features = false }
 once_cell = { version = "1.7", default-features = false, features = ["std"] }
 parking_lot = { version = "0.12" }
+rand = { version = "0.8" }
```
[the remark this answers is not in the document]
PhilippGackstatter: Just to note: I will look into whether this needs to be different when working on Wasm support shortly.

identity-account-storage/src/types/key_location.rs
cycraig: Indeed, with the `key_hash` it would at worst just point to the same private key. I do wonder if it won't cause problems for other `Storage` implementations though, since there's no guarantee they use `KeyLocation::to_string` or compare them in the same way. It's a bit odd not to depend on the `KeyType` field but that's probably irrelevant.

identity-account-storage/src/types/key_location.rs
cycraig: That would be fine but I would add getters for all fields for consistency then.

As far as I see the `fragment` is only used in `remote_sign_data` internally, and that could be refactored to take a `VerificationMethod` directly (for the `DIDUrl`), bypassing this issue entirely so we can use `String` instead?

PhilippGackstatter:
I have updated the Rust implementation to no longer use `AccountId` and instead use only `IotaDID` as the identifier (as it was until now). I've also updated the PR description to be accurate for the current version.

I'm going to request another review for just the Rust part, so we can hopefully get this PR done sooner. I'm also particularly interested in feedback on the documentation of the Storage trait. I will start working on Wasm/Napi in the meantime

identity-account-storage/src/types/key_location.rs
PhilippGackstatter: I think the `KeyType` is only needed as the indication for what type of key is at that location, but is not otherwise part of the location itself. It could also be removed from the `KeyLocation` and passed explicitly, but that's one more parameter for certain functions (`key_sign`, `key_public`).
We could additionally prepend the key location string with the string representation of the key type, though I think that shouldn't absolutely be required?

> there's no guarantee they use `KeyLocation::to_string`

I've added documentation to the `KeyLocation` type to explicitly state the expectations, so technically implementers can choose their own formatting, as long as it depends on the fragment and the hash.

identity-account-storage/src/types/key_location.rs
PhilippGackstatter: I think you're right, but I'm not completely sure that we can always guarantee the higher-level validation, especially when we now expect storage implementations to construct key locations. Technically, of course, we will just pass them the fragment for them to just pass it into the constructor.
I've switched to `Fragment` now, but I'm also fine with changing it back. Not sure what to do.

identity-account-storage/src/stronghold/snapshot.rs
```diff
@@ -44,25 +43,12 @@ impl Snapshot {
...
+    Vault::new(&self.path, client_path, &[])
   }
 
-  pub fn records<T>(&self, name: &T, flags: &[StrongholdFlags]) -> Records<'_>
-  where
-    T: AsRef<[u8]> + ?Sized,
-  {
-    Records::new(&self.path, name, flags)
+  pub fn store(&self, client_path: ClientPath) -> Store<'_> {
+    Store::new(&self.path, client_path, &[])
```
cycraig, line 51: Would making the parameter generic with `T: Into<ClientPath>` be better or do we want to avoid generics since these are private structs anyway?

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +29,88 @@ mod storage_sub_trait {
...
-  async fn state(&self, did: &IotaDID) -> Result<Option<IdentityState>>;
+  /// Returns the [`IotaDocument`] of the identity specified by `did`.
+  async fn document_get(&self, did: &IotaDID) -> Result<Option<IotaDocument>>;
 
   /// Sets a new state for the identity specified by `did`.
-  async fn set_state(&self, did: &IotaDID, state: &IdentityState) -> Result<()>;
+  async fn document_set(&self, did: &IotaDID, state: &IotaDocument) -> Result<()>;
+
+  /// Returns `true` if the index contains the given `did`.
+  async fn index_has(&self, did: &IotaDID) -> Result<bool>;
```
cycraig, line 109: I think `did_exists` would be more consistent.

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +29,88 @@ mod storage_sub_trait {
...
-  /// Deletes the keypair specified by `location`.
-  async fn key_del(&self, did: &IotaDID, location: &KeyLocation) -> Result<()>;
+  /// Creates a new identity for the given `network`.
+  ///
+  /// - Uses the given Ed25519 `private_key` or generates a new key if it's `None`.
+  /// - Returns an [`IdentityAlreadyExists`](crate::Error::IdentityAlreadyExists) error if the DID already exists.
+  /// - Adds the newly created DID to an index which can be accessed via [`Storage::index`].
+  ///
+  /// Returns the generated DID and the location at which the key was stored.
+  async fn did_create(
```
cycraig, line 57: I don't think we can mandate the type of error returned (we don't export error enum variants to bindings as far as I know), just that it returns some sort of error if the identity already exists.

We can throw that error in `Account` ourselves using `did_exists` prior to calling `did_create`.

The entire `Storage` interface doc-comments might need a pass to ensure internal consistency with the expected components. E.g. we mention `an index` here, it should be `the index` or `the DID index` and explain what that is in the doc-comment on `Storage` itself (along with other required components like the private key storage and Document/ChainState storage). Not sure if we should mention an index explicitly since that may just be an implementation detail.

Should we mention the exact functions to use like `KeyLocation::new` and `IotaDID::new`/`IotaDID::new_with_network`?

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +29,88 @@ mod storage_sub_trait {
...
+/// as well as key-value like storage of data structures, such as DID documents.
+///
+/// This interface works with [`AccountId`] and [`IotaDID`] as the top-level identifiers.
+/// `AccountId` is intended as a partition-key, i.e. everything related to an account
+/// can be stored in its own partition. Keys belonging to an account are identified by [`KeyLocation`]s
+/// in that partition.
+/// An `IotaDID` can be resolved to an `AccountId` using the index, which is a global data structure.
+/// Therefore, index operations need to be carefully synchronized per [`Storage`] instance.
+/// Other operations don't need to be synchronized globally, as it is a user error to create
+/// more than one `Account` for the same identity.
```
cycraig, line 44: Remove references to `AccountId`.

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +29,88 @@ mod storage_sub_trait {
...
-  async fn set_state(&self, did: &IotaDID, state: &IdentityState) -> Result<()>;
+  async fn document_set(&self, did: &IotaDID, state: &IotaDocument) -> Result<()>;
+
+  /// Returns `true` if the index contains the given `did`.
+  async fn index_has(&self, did: &IotaDID) -> Result<bool>;
 
-  /// Removes the keys and any state for the identity specified by `did`.
-  async fn purge(&self, did: &IotaDID) -> Result<()>;
+  /// Retrieves the list of [`IotaDID`]s stored in the index.
+  async fn index(&self) -> Result<Vec<IotaDID>>;
```
cycraig, line 112: Maybe call this `did_index` or `did_list`? Otherwise it stands as an outlier with no other common functions if `index_has` is renamed.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -213,66 +325,82 @@ impl Drop for Stronghold {
...
+  Ok(())
 }
 
-fn fmt_key(prefix: &str, location: &KeyLocation) -> Vec<u8> {
-  format!("{}:{}:{}", prefix, location.generation(), location.fragment_name()).into_bytes()
+impl From<&KeyLocation> for Location {
+  fn from(key_location: &KeyLocation) -> Self {
+    Location::generic(VAULT_PATH.to_vec(), key_location.to_string().into_bytes())
+  }
 }
```
cycraig, line 399: I don't think this should use `to_string` since the `Display` implementation could be changed without anyone thinking it affects the `Stronghold` locations.

identity-account-storage/src/types/key_location.rs
cycraig: So the only reason we need it to be a valid DID-Url fragment right now is because the signature uses the `KeyLocation` fragment to reconstruct the DIDUrl of the `VerificationMethod`. Even then a `Fragment` just adds a `#` which is removed in the `KeyLocation::fragment()`, so it's not doing any validation anyway as far as I can see.

My initial comment referenced previous conversations/PRs where it was discussed that we should ideally refactor `Fragment` to do the same validation as `DIDUrl` if we intend to use it independently or remove it altogether.

I vote removing it and using `String` again for now, apologies for the back-and-forth.

HenriqueNogara (approved):
LGTM

identity-account-storage/src/error.rs
```diff
@@ -12,6 +12,9 @@ pub enum Error {
   /// Caused by errors from the [identity_core] crate.
   #[error(transparent)]
   CoreError(#[from] identity_core::Error),
+  /// Caused by errors from the [identity_core] crate.
+  #[error(transparent)]
+  IotaCoreError(#[from] identity_iota_core::Error),
```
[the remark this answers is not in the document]
PhilippGackstatter: It's only used for conversion of the return value of `IotaDID::new_with_network` in the respective `Storage::did_create` implementations. Not sure what useful context I could add, other than the location where it happens, i.e. `did_create`. But is that useful enough to have a dedicated error variant, I'm leaning towards no.
In this particular case, we will have to re-do this for the `StorageError` refactor anyway, so perhaps we can leave it as-is for now?

identity-account-storage/src/stronghold/client_path.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use identity_iota_core::did::IotaDID;
+
+#[derive(Debug, Clone)]
+pub struct ClientPath(pub String);
```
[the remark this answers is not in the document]
PhilippGackstatter: It needs to be `pub` because the `stronghold` module is used from the `storage` module. Although that made me realize that we export our stronghold wrapper interface publicly, which is unnecessary and a lot of it will go away with the upcoming migration to the new version anyway:

```rs
// identity/src/lib.rs
pub use identity_account_storage::stronghold::*;
```

I think we should make this a private module, and only expose our `Stronghold` type from the `storage` module. Might as well do that breaking change with the next release. (I think that should be a separate PR and ideally after this PR was merged to avoid conflicts).

identity-account/Cargo.toml
```diff
@@ -27,6 +27,7 @@ zeroize = { version = "1.4", optional = true }
 
 [dev-dependencies]
 futures = { version = "0.3" }
+once_cell = { version = "1.10", default-features = false, features = ["std"] }
```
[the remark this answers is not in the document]
PhilippGackstatter: Thanks for the catch. It is used when adding the stronghold test from the PR description in `identity_account::tests::util`, but while that is not actually in the code base this can be removed.

identity-account-storage/src/error.rs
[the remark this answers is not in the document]
cycraig: Then there should be a `DIDCreationError` or something that just wraps the error string.

identity-account-storage/src/types/key_location.rs
```diff
@@ -6,60 +6,105 @@ use core::fmt::Display;
...
+      .map(char::from)
+      .collect();
+    let key_hash: u64 = (&mut thread_rng).gen();
+
+    Self {
+      key_type,
+      fragment: Fragment::new(fragment),
+      key_hash,
+    }
   }
```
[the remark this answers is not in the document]
PhilippGackstatter: You're right, probably not necessary. That also nicely enables us to feature-gate the `rand` crate to the `stronghold` feature, so no Wasm issues.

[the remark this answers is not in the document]
cycraig: There is a location where I used `rand` when `tesmode` is set to true in `Account`. There wasn't an elegant way to feature-gate it, but maybe we can go with an unelegant solution. It wasn't exposed to Wasm in my situation since `testmode` is internal-only.

identity-account-storage/src/stronghold/snapshot.rs
PhilippGackstatter: That would work just as well, and that's what I had initially, but then changed it to be super explicit about what the client path is. Since stronghold just uses `Vec<u8>`, almost anything goes as the client identifier. I mainly did this to not lose my sanity while trying to make sure to consistently use a fixed string or a did as the client path, which is important for forward compatibility. So I'll leave it as more explicit for now.

identity-account-storage/src/storage/traits.rs
PhilippGackstatter: True, as of now, we cannot specify the type we expect. With a potential `StorageError`, we could, as we would probably also export that to bindings, since it just becomes part of the interface at that point.

> just that it returns some sort of error if the identity already exists.

Will do that for now.

> We can throw that error in `Account` ourselves using `did_exists` prior to calling `did_create`.

Can't do that, because the DID is created inside `did_create`, so it's the storage's responsibility to check.

> Not sure if we should mention an index explicitly since that may just be an implementation detail.

Probably not, especially when we rename the `index_*` functions to `did_*`. Mentioning a "list" is probably good enough.

> Should we mention the exact functions

I'm hoping it would be fairly clear, also given example implementations. But I'm mainly trying to save ourselves some work of maintaining docs.

Perhaps a better thing we could do is write a storage test that users can easily plug their storage into, also from bindings, and then we could for example test, that a certain private key of a `KeyPair` returns the expected DID. I realize that would be (much) more work than just documenting it, but also better. Anyway, not trying to do this for this PR.

identity-account-storage/src/storage/traits.rs
PhilippGackstatter: Great suggestion, thanks. This makes the interface so much more consistent :rocket:

identity-account-storage/src/storage/stronghold.rs
PhilippGackstatter: That's fair. Each storage can derive a location on their own based on `KeyLocation::fragment` and `KeyLocation::key_hash`. In that case, I will remove the `Display` impl entirely, since there's no reason to have it then.

cycraig: It's not about removing `Display`, it's not using the implementation for logic. We can still use it in `DuplicateKeyLocation` and errors if needed. Error message strings are not intended to be stable in terms of API stability when considering breaking changes. Error variants are breaking changes, however.

PhilippGackstatter: It seems like good practice to me to have a stable `Display` implementation. If we don't commit to use `Display` as the canonical representation of a location, then I'm not sure if it still serves any purpose. If any storage is expected to use a different representation internally, a debug message containing our custom `Display` implementation doesn't help, `Debug` is better, even.

cycraig: `Display` should be as stable as the struct itself. As its namesake indicates, it should only be used for `Display` purposes, i.e. human-readable string representations. `Debug` string representations are generally considered not stable in `Rust`, even for standard library objects.

As previously suggested, we can offer a similar function for a canonical string representation suitable for `Storage` locations on `KeyLocation` that does not depend on `Display`.

eike-hass:
Sorry for chiming in so late, but I am not a fan of the `key_generate` addition to the interface. It bleeds stronghold specific logic into the interface, as you said for everyone to deal with. I thought maybe it could be a benefit rolling own crypto, but I believe this would be the wrong point to enable that. I would lean towards Option, but since this is all pretty far along I assume `key_generate` will become a thing. Just wanted to give my two cents.

PhilippGackstatter:
I assume you mean "you're not a fan of `did_create`"? `key_generate` is just the renamed `key_new`.

If we want to be able to derive locations from verification methods without keeping extra state, this is one of the trade-offs we have to make. Either it's `AccountId`, which is ultimately throwing some state at the problem to solve it, or give a little more responsibility to the interface to break the initial circular dependency of generating the DID. We could get away without `did_create` and just `key_generate` and `key_move` functions (generate the initial key at some random DID, then move it to the proper location later), but that's just a different trade-off and different weirdness.

> but would storage ever need to know about that? Like would storage store differently for different types, e.g.?

Excellent question. We could indeed get away with just passing the canonical representation of the key location as a string, e.g. "fragment:123456789" to storage. I don't think storage ever needs just the fragment of a location, for example. However, `KeyLocation` also includes the `KeyType` for convenient access to the type of the key at that location, and I've considered removing it, but it makes working with all of that quite a bit less convenient and requires another parameter for the functions that need it.

Still, your comment points out that it's probably better to just expose a string representation to storage rather than the `fragment` and `key_hash` fields. That's least confusing for implementers, I imagine.

eike-hass:
> I assume you mean "you're not a fan of `did_create`"? `key_generate` is just the renamed `key_new`.
> 
> If we want to be able to derive locations from verification methods without keeping extra state, this is one of the trade-offs we have to make. Either it's `AccountId`, which is ultimately throwing some state at the problem to solve it, or give a little more responsibility to the interface to break the initial circular dependency of generating the DID. We could get away without `did_create` and just `key_generate` and `key_move` functions (generate the initial key at some random DID, then move it to the proper location later), but that's just a different trade-off and different weirdness.
> 
> > but would storage ever need to know about that? Like would storage store differently for different types, e.g.?
> 
> Excellent question. We could indeed get away with just passing the canonical representation of the key location as a string, e.g. "fragment:123456789" to storage. I don't think storage ever needs just the fragment of a location, for example. However, `KeyLocation` also includes the `KeyType` for convenient access to the type of the key at that location, and I've considered removing it, but it makes working with all of that quite a bit less convenient and requires another parameter for the functions that need it.
> 
> Still, your comment points out that it's probably better to just expose a string representation to storage rather than the `fragment` and `key_hash` fields. That's least confusing for implementers, I imagine.

Indeed `did_create` my bad. Although not ideal, using a UUID as the key sounds reasonable to me, if it removes the need for `did_create`. Maybe similar to the ``KeyLocation` discussion: Is the storage actually concerned with the DID, or is it just a storage key, and what we see here is a hierarchy of two arbitrary storage keys? I think the question is does an implementer want to inspect it and act accordingly. 
Further regarding the "KeyLocation" if it is easier for us when reconstructing we can totally pre-/append the type to the fragment.

bindings/wasm/src/account/types/key_location.rs
```diff
@@ -4,56 +4,73 @@
...
-  /// Returns a copy of the fragment name of the key location.
-  #[wasm_bindgen]
-  pub fn fragment(&self) -> String {
-    self.0.fragment().clone().into()
+  /// Obtain the location of a verification method's key in storage.
+  #[wasm_bindgen(js_name = fromVerificationMethod)]
+  pub fn from_verification_method(method: &WasmVerificationMethod) -> Result<WasmKeyLocation> {
+    Ok(WasmKeyLocation(
+      KeyLocation::from_verification_method(&method.0).wasm_result()?,
+    ))
```
cycraig, line 40:
```suggestion
  pub fn from_verification_method(method: &WasmVerificationMethod) -> Result<WasmKeyLocation> {
    KeyLocation::from_verification_method(&method.0).map(WasmKeyLocation).wasm_result()
```

bindings/wasm/tests/wasm.rs
```diff
@@ -445,3 +446,24 @@ fn test_validations() {
...
+  let test_vector_2: [u8; 32] = [
+    125, 153, 99, 21, 23, 190, 149, 109, 84, 120, 40, 91, 181, 57, 67, 254, 11, 25, 152, 214, 84, 46, 105, 186, 16, 39,
+    141, 151, 100, 163, 138, 222,
+  ];
+
+  let location_1 = WasmKeyLocation::new(WasmKeyType::Ed25519, "".to_owned(), test_vector_1.to_vec());
+  let location_2 = WasmKeyLocation::new(WasmKeyType::Ed25519, "".to_owned(), test_vector_2.to_vec());
+
+  assert_eq!(location_1.to_string().split(':').last().unwrap(), "74874706796298672");
+  assert_eq!(location_2.to_string().split(':').last().unwrap(), "10201576743536852223");
```
cycraig, line 468: This should use `.canonical_repr` not `to_string`.

I would be fine testing `canonical_repr` as a whole to detect breaking changes to it by means of the unit tests failing.

bindings/wasm/src/account/types/key_location.rs
```diff
@@ -4,56 +4,73 @@
...
   #[wasm_bindgen(constructor)]
-  pub fn new(method: &WasmMethodType, fragment: String, generation: &WasmGeneration) -> WasmKeyLocation {
-    WasmKeyLocation(KeyLocation::new(method.0, fragment, generation.0))
-  }
-
-  /// Returns a copy of the method type of the key location.
-  #[wasm_bindgen]
-  pub fn method(&self) -> WasmMethodType {
-    self.0.method().into()
+  pub fn new(key_type: WasmKeyType, fragment: String, public_key: Vec<u8>) -> WasmKeyLocation {
```
cycraig, line 31: Use camelCase for Wasm parameters.

```suggestion
  #[allow(non_snake_case)]
  pub fn new(keyType: WasmKeyType, fragment: String, publicKey: Vec<u8>) -> WasmKeyLocation {
```

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -82,132 +101,226 @@ impl Stronghold {
...
+    &self,
+    network: NetworkName,
+    fragment: &str,
+    private_key: Option<PrivateKey>,
+  ) -> Result<(IotaDID, KeyLocation)> {
+    // TODO: Temporarily implemented via in-memory generation as opposed to using stronghold procedures.
+    let keypair: KeyPair = match private_key {
+      Some(private_key) => KeyPair::try_from_private_key_bytes(KeyType::Ed25519, private_key.as_ref())?,
+      None => KeyPair::new(KeyType::Ed25519)?,
+    };
```
cycraig, line 114: TODO? Are we fine shipping this?

identity-account/src/tests/util.rs
```diff
@@ -0,0 +1,52 @@
...
+      .collect::<String>(),
+  );
+  file.set_extension("stronghold");
+  file.to_str().unwrap().to_owned()
+}
+
+// TODO: This should also return a `Stronghold`.
+pub(super) async fn storages() -> [Arc<dyn Storage>; 1] {
+  [Arc::new(MemStore::new())]
+}
```
cycraig, line 52: TODO

identity-account-storage/src/types/key_location.rs
```diff
@@ -5,61 +5,163 @@ use core::fmt::Debug;
...
-  method: MethodType,
-  fragment: Fragment,
-  generation: Generation,
+  /// The [`KeyType`] of the key.
+  pub key_type: KeyType,
+  /// The fragment of the key.
+  fragment: String,
+  /// The hash of the public key.
+  #[serde(with = "key_hash_serialization")]
+  pub(in crate::types::key_location) key_hash: u64,
```
cycraig, line 33: Shouldn't we just use a `String` to store the hash if it's giving problems with serialization, or do we want to enforce it's always a number?

identity-account-storage/src/types/key_location.rs
```diff
@@ -5,61 +5,163 @@ use core::fmt::Debug;
...
   }
 
-  /// Returns the integration generation when this key was created.
-  pub fn generation(&self) -> Generation {
-    self.generation
+  /// Returns the canonical string representation of the location.
+  ///
+  /// This should be used as the representation for storage keys.
+  pub fn canonical_repr(&self) -> String {
+    format!("{}:{}", self.fragment, self.key_hash)
```
cycraig, line 73: Not a big fan of the abbreviation.

Maybe just `canonical`?

identity-account-storage/src/types/key_location.rs
```diff
@@ -5,61 +5,163 @@ use core::fmt::Debug;
...
-  fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-    f.write_fmt(format_args!("KeyLocation{}", self))
+// Custom Hash and Equality implementations to not include the key_type.
+
+impl Hash for KeyLocation {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    self.fragment.hash(state);
+    self.key_hash.hash(state);
+  }
+}
```
cycraig, line 90: I feel like this should use `canonical_repr` internally. This likely has the same platform-dependent bug with `hash`.

identity-account-storage/src/storage/traits.rs
```diff
@@ -27,45 +29,107 @@ mod storage_sub_trait {
...
-  /// Retrieves the public key at the specified `location`.
-  async fn key_get(&self, did: &IotaDID, location: &KeyLocation) -> Result<PublicKey>;
-
-  /// Deletes the keypair specified by `location`.
-  async fn key_del(&self, did: &IotaDID, location: &KeyLocation) -> Result<()>;
+  /// Creates a new identity for the given `network`.
+  ///
+  /// - Uses the given Ed25519 `private_key` or generates a new key if it's `None`.
+  /// - Returns an error if the DID already exists.
+  /// - Adds the newly created DID to an index which can be accessed via [`Storage::index`].
```
cycraig, line 73: Remove reference to index and `Storage::index`.

cycraig (approved):
Looks good to me overall. I'm sure I missed something but I think most internal issues could be addressed in patches. Still need to make Stronghold stuff private as discussed in previous comments.

Approving tentatively.

identity-account/src/types/method_content.rs
```diff
@@ -41,15 +40,16 @@ pub enum MethodContent {
 }
 
 impl MethodContent {
-  /// Returns the [`MethodType`] associated with the `MethodContent` variant.
-  pub(crate) fn method_type(&self) -> MethodType {
+  /// Returns the [`MethodType`](identity_did::verification::MethodType) associated with the `MethodContent` variant.
+  #[cfg(test)]
+  pub(crate) fn method_type(&self) -> identity_did::verification::MethodType {
```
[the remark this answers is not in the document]
cycraig: Unresolved.

bindings/stronghold-nodejs/js/stronghold.ts
```diff
@@ -16,76 +16,112 @@ export class Stronghold implements Storage {
...
 
-    public async keyNew(did: DID, keyLocation: KeyLocation) {
-        let napiDID = NapiDID.fromJSON(did.toJSON());
-        let napiKeyLocation = NapiKeyLocation.fromJSON(keyLocation.toJSON());
-        let publicKey = await this.napiStronghold.keyNew(napiDID, napiKeyLocation);
-        return Uint8Array.from(publicKey)
+    public async didExists(did: DID): Promise<boolean> {
+        const napiDID: NapiDID = NapiDID.fromJSON(did.toJSON());
+        return this.napiStronghold.didExists(napiDID);
     }
```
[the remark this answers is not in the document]
PhilippGackstatter: Mh, I thought it was better to return the promise directly since it's already of the same type. @eike-hass what's your opinion?

bindings/stronghold-nodejs/js/stronghold.ts
```diff
@@ -16,76 +16,112 @@ export class Stronghold implements Storage {
         return stronghold
     }
 
-    public async flushChanges() {
-        return this.napiStronghold.flushChanges()
+    public async didCreate(network: string, fragment: string, private_key: Uint8Array | undefined | null): Promise<[DID, KeyLocation]> {
```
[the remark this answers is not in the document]
cycraig: The `?` goes after the identifier, before the colon, not after the `type`.

`privateKey?: Uint8Array`

identity-account-storage/src/storage/stronghold.rs
PhilippGackstatter: It's either generate the KeyPair here or for every storage in the account. This way, only stronghold is temporary which is where the actual problem comes from. If we don't want to ship this, we need to migrate to the new stronghold first.

bindings/stronghold-nodejs/js/stronghold.ts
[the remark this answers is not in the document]
eike-hass: technically both are identical. The function being marked `async` would wrap the return `boolean` in a Promise. It wouldn't wrap a Promise though. I see no advantage either way in regards to typing. The return signature makes sure we don't mess up in both cases.  So it is a question of taste, awaiting the result makes it clearer that there is another async action, but I lean towards leaving it as is for simplicity.

identity-account-storage/src/types/key_location.rs
PhilippGackstatter: Good point, although the hash is not exposed to other languages right now (could be in the future I suppose), and only used so we can use `KeyLocation` as the key in `MemStore` rather than converting to strings.

identity-account/src/tests/util.rs
PhilippGackstatter: See Test section in PR description. Will be fixed with the stronghold migration.

identity-account-storage/src/types/key_location.rs
cycraig: Even used internally I think it may break in Wasm, similar to `KeyLocation::new`.

bindings/wasm/tests/wasm.rs
PhilippGackstatter: Extended the hash test to also test canonical.

identity-account-storage/src/types/key_location.rs
PhilippGackstatter: True that, I updated it to use `KeyLocation::canonical`.

PR #733, opened 2022-03-18
Remove `Storage::set_password`

abdulmth:
# Description of change
Remove the `set_password()` function from the account Storage interface.

fixes issue #733 

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/bindings/stronghold-nodejs/js/stronghold.ts
+++ b/bindings/stronghold-nodejs/js/stronghold.ts
@@ -16,6 +16,10 @@ export class Stronghold implements Storage {
         return stronghold
     }
 
+    public async setPassword(encryptionKey: Uint8Array) {
+        return this.napiStronghold.setPassword(Array.from(encryptionKey))
+    }
+
     public async flushChanges() {
         return this.napiStronghold.flushChanges()
     }
--- a/bindings/stronghold-nodejs/js/stronghold.ts
+++ b/bindings/stronghold-nodejs/js/stronghold.ts
@@ -16,10 +16,6 @@ export class Stronghold implements Storage {
         return stronghold
     }
 
-    public async setPassword(encryptionKey: Uint8Array) {
-        return this.napiStronghold.setPassword(Array.from(encryptionKey))
-    }
-
     public async flushChanges() {
         return this.napiStronghold.flushChanges()
     }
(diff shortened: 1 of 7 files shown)


bindings/stronghold-nodejs/js/stronghold.ts
```diff
@@ -16,10 +16,6 @@ export class Stronghold implements Storage {
         return stronghold
     }
 
-    public async setPassword(encryptionKey: Uint8Array) {
```
eike-hass, line 19: I assume we still need this to set the stronghold password, although it is not part of the interface anymore.

abdulmth: I agree, thanks Eike.

PhilippGackstatter: It's not required because we set the password once during construction. This does not really suffice in the current stronghold, because the password expires after some time. This is also the case in Rust, though. Once we have migrated to the refactored stronghold, it will be fixed. I still think we can remove `set_password` in the NAPI bindings already.

eike-hass: Right, because we set it in the constructor initially, which is enough for the new stronghold version. How do we coordinate this?

PhilippGackstatter: Coordinate the upgrade? Personally, I'd like to do it as soon as possible once the stronghold team gives the green light.

eike-hass: Or maybe worded differently: what is the effect if we merge it like this and do we need for a new stronghold version for a potential dev release?

PhilippGackstatter: > what is the effect if we merge it like this

Sorry for the delay, had to re-understand how password expiration works. Actually, the default is that the password does not expire at all. Merging it like this is therefore fine, as we set the password during construction, and then we don't need to set it again.



> do we need for a new stronghold version for a potential (dev) release?

I don't think we _need_ it. However, I can't guarantee that there is not going to be a breaking change during the migration to the new version. That said, from doing the `KeyLocation` refactor and from what I know about the new version, at least I cannot say that there will definitely be anything broken after the migration.

PhilippGackstatter (approved):
Looks good to me, but can't we remove `Stronghold::set_password` as well?

bindings/stronghold-nodejs/js/stronghold.ts
eike-hass: Thank you for clarifying. I was worried it would expire. SO we can totally merge and do a dev release.
Let's discuss potential breaking changes before 0.5.0 proper release.

eike-hass (approved):

Issue #721, opened 2022-03-16
[Task] Change keys in Wasm from Base58-BTC strings to `UInt8Array`


cycraig:
## Description
Change the public, private key representations used in the Wasm bindings from Base58-BTC encoded `String`s to `UInt8Array`.

## Motivation
Keys are not generally meant to be viewable and developers can encode them however they wish. `UInt8Array` is also the format expected by existing libraries such as `TweetNaCl`, so from an interoperability standpoint using `UInt8Array` would be more convenient.

May cause issues with the Stronghold Napi bindings though.

## Resources
This is a sub-task extracted from:
- https://github.com/iotaledger/identity.rs/issues/669

## To-do list

- [ ] Change keys in Wasm from Base58-BTC strings to `UInt8Array`

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

HenriqueNogara:
#734

PR #734, opened 2022-03-19
Change Wasm key types to `UInt8Array`

HenriqueNogara:
# Description of change
Change keys in Wasm from Base58-BTC strings to UInt8Array

#721 

## Type of change
- [ ] Documentation Fix

All examples and existing tests have been run

## Change checklist


PhilippGackstatter (approved):
Looks good to me, thanks!

bindings/stronghold-nodejs/src/account/storage/stronghold.rs
```diff
@@ -72,28 +70,31 @@ impl NapiStronghold {
...
-    Ok(encode_b58(&public_key))
+    let public_key: Vec<u8> = (&public_key).into();
+    Ok(public_key.into_iter().map(u32::from).collect())
   }
 
   /// Inserts a private key at the specified `location`.
   #[napi]
-  pub async fn key_insert(&self, did: &NapiDID, location: &NapiKeyLocation, private_key: String) -> Result<String> {
-    let private_key: PrivateKey = decode_b58(&private_key).napi_result()?.into();
+  pub async fn key_insert(&self, did: &NapiDID, location: &NapiKeyLocation, private_key: Vec<u32>) -> Result<Vec<u32>> {
```
cycraig, line 81: Can we open an issue with napi.rs regarding `Vec<u8>`/`UInt8Array` support? I don't understand the problem in enough detail to do so myself.

bindings/wasm/Cargo.toml
```diff
@@ -33,6 +33,7 @@ default-features = false
 features = ["account", "bindings-derive"]
 
 [dev-dependencies]
+json = { version = "0.12.4" }
 wasm-bindgen-test = { version = "0.3" }
```
cycraig, line 37: Do we absolutely need this `json` dev-dependency? Is it for convenience or is there something that `serde_json` cannot do?

bindings/wasm/examples/src/create_vc.js
```diff
@@ -38,8 +38,8 @@ async function createVC(clientConfig) {
     // Sign the credential with the Issuer's newKey
     const signedVc = issuer.doc.signCredential(unsignedVc, {
         method: issuer.doc.id().toString() + "#newKey",
-        public: issuer.newKey.public(),
-        private: issuer.newKey.private(),
+        public: Array.from(issuer.newKey.public()),
+        private: Array.from(issuer.newKey.private()),
     }, SignatureOptions.default());
```
cycraig, line 43: I feel like this is an unacceptable loss of ergonomics for developers. What type does `Array.from` return? What if developers forget to call `Array.from`, is it type-enforced? Is it just a problem with the duck-typed interface here or in general? Can we correct it if it's the former?

Ideally they should just have to call `.public()` or `.private()`. We may have to reconsider `UInt8Array` and instead use key wrapper types like `Public`/`Private` that can be converted to `UInt8Array` as-needed to avoid situations like this.

If we are forced to use `Array.from(uint8array)` or `[uint8array.buffer]` that's fine, but I'd like to have all options explored before that.

bindings/wasm/src/crypto/key_pair.rs
```diff
@@ -35,31 +31,28 @@ impl WasmKeyPair {
...
-  /// Parses a `KeyPair` object from base58-encoded public/private keys.
+  /// Parses a `KeyPair` object from the public/private keys.
   #[wasm_bindgen(js_name = fromBase58)]
-  pub fn from_base58(type_: KeyType, public_key: &str, private_key: &str) -> Result<WasmKeyPair> {
-    let public: PublicKey = decode_b58(public_key).wasm_result()?.into();
-    let private: PrivateKey = decode_b58(private_key).wasm_result()?.into();
-
-    Ok(Self((type_.into(), public, private).into()))
+  pub fn from_keys(type_: KeyType, public_key: Vec<u8>, private_key: Vec<u8>) -> Result<WasmKeyPair> {
+    Ok(Self((type_.into(), public_key.into(), private_key.into()).into()))
```
cycraig, line 37: Fix the name, it's still `fromBase58`, it should be `fromKeys`.

identity-core/src/crypto/key/key.rs
```diff
@@ -88,3 +88,9 @@ impl<V: private::Sealed> From<Vec<u8>> for Key<V> {
     other.into_boxed_slice().into()
   }
 }
+
+impl<V: private::Sealed> From<&Key<V>> for Vec<u8> {
+  fn from(other: &Key<V>) -> Self {
+    other.as_ref().to_vec()
+  }
+}
```
cycraig, line 96: This seems unnecessary, can't one just call `as_ref` to access the underlying `&[u8]` then call `to_vec` on that?

bindings/wasm/src/did/wasm_did.rs
```diff
@@ -22,23 +21,17 @@ impl WasmDID {
...
-  fn from_public_key(public: &[u8], network: Option<String>) -> Result<WasmDID> {
+  #[wasm_bindgen(js_name = fromPublicKey)]
+  pub fn from_public_key(public_key: &[u8], network: Option<String>) -> Result<WasmDID> {
     let did = if let Some(network) = network {
-      IotaDID::new_with_network(public, network)
+      IotaDID::new_with_network(public_key, network)
     } else {
-      IotaDID::new(public)
+      IotaDID::new(public_key)
     };
```
cycraig, line 35: We don't have `from_public_key` in the Rust library anymore, remove `from_public_key` and change the `new` constructor in the Wasm bindings to take a public key (`&[u8]`) instead of a `KeyPair`. I.e. change this to match the Rust interface for `IotaDID`.

cycraig (changes requested):

identity-core/src/crypto/key/key.rs
HenriqueNogara: We can. I just thought calling `into()` would be more convenient, but I'll remove this.

bindings/wasm/examples/src/create_vc.js
cycraig: If that's the only location causing problems (and we'll have to double-check that it is), then making `Args` a proper duck-typed interface with structural getters rather than going through serde would be the solution.

HenriqueNogara: As we discussed, since #755  removed this, it's not needed.

bindings/stronghold-nodejs/src/account/storage/stronghold.rs
cycraig: Totally not required, as long as we have a way around this.

HenriqueNogara: [issue](https://github.com/napi-rs/napi-rs/issues/1105).

cycraig (approved):

Issue #715, opened 2022-03-15
[Task] Remove `MerkleKeyCollection2021`


cycraig:
## Description
Remove `MerkleKeyCollection2021` and all associated structs.

## Motivation
The primary purpose of `MerkleKeyCollection2021` is to provide a means to issue and revoke verifiable credentials individually. Unfortunately, there are problems with it scaling due to having to store each individual private key in the collection in the current implementation. So to issue one million individually-revocable credentials, I need to store one million keys.

The intended use-case of revoking each credential individually means using each constituent key only once, and there are existing one-time-signature schemes that do not require storing each key but rather a single root key and associated state. Storing used private keys without being able to use them without losing the capability for individual revocation is unnecessary.

There are also [`credentialStatus`](https://www.w3.org/TR/vc-data-model/#status) specifications that enable revoking credentials individually, separate from the verification method and signature used, such as [Revocation List 2020](https://w3c-ccg.github.io/vc-status-rl-2020/). While one could argue that `MerkleKeyCollection2021` is useful for potentially revoking arbitrary signatures/messages, one can still achieve that by incorporating a scheme similar to `credentialStatus` for arbitrary signed data at scale. In general, `credentialStatus` is more flexible since it is not necessarily tied to a DID Document (which could be seen as a disadvantage in some situations but one can always create a status scheme that _is_ tied to a DID Document) and allows separating revocation from the signature, since it is useful to know whether a signed message was _ever_ valid or simply revoked later.

The argument could be made that the keys could be distributed to allow a compact verification method able to be used by multiple parties, but there also exist "k of n" multi-sig schemes that could be used equivalently in a "1 of n" setting.

Overall, while incredibly useful and acknowledging that improvements to it could be made, `MerkleKeyCollection2021` will eventually be replaced by one or more alternatives. To avoid a backwards-incompatible breaking change to DID Documents in the future, we should remove it completely now, even without an alternative in place. We could decide to add it back later in an improved and stable form with indefinite support.

## Resources
- https://wiki.iota.org/identity.rs/specs/did/merkle_key_collection
- https://www.w3.org/TR/vc-data-model/#status
- https://w3c-ccg.github.io/vc-status-rl-2020/

## To-do list

- [ ] Remove `MerkleKeyCollection2021` and all associated structs.
- [ ] Update the documentation, IOTA DID Method Specification.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

[...]

PR #735, opened 2022-03-20
Add X25519 key and verification method support

### LINKED ISSUE #671: [Request] Support `X25519KeyAgreementKey2019`
## Description
Add support for the `X25519KeyAgreementKey2019` verification method type.

- https://www.w3.org/TR/did-spec-registries/#x25519keyagreementkey2019

## Motivation
X25519 key exchange is used for setting up encrypted communication, so in anticipation of the DIDComm Actor we should support and expose creating X25519 keys and verification methods.

One argument against doing so is that Ed25519 keys can already be transformed to Curve25519 for use with X25519: https://en.wikipedia.org/wiki/Curve25519
Even so, we do not expose such functionality directly nor provide examples on how to do so. This means performing these operations securely via the Stronghold storage interface is not an option currently.

## Requirements

1. Add the `X25519` key type.
2. Add `X25519KeyAgreementKey2019` verification method type.
3. Add example demonstrating adding a `keyAgreement` verification method relationship and performing Diffie-Hellman key exchange between two parties.  (Optionally via DIDComm/Actor?)

## Open questions

1. Do we want to support X25519 directly or rely on Ed25519 transformations?
2. Should we implement this prior to a DIDComm Actor or as a prerequisite?
3. Does Stronghold already support all required functionality? See #625.
4. ~~For which milestone would this be slated?~~ 0.6, same as #625.
5. Does it matter that there is a warning that "`X25519KeyAgreementKey2019` has no normative definition" in the [specification registry](https://www.w3.org/TR/did-spec-registries/#x25519keyagreementkey2019), is it at risk?

## Are you planning to do it yourself in a pull request?

Depends on priorities.

cycraig:
# Description of change
Adds support for `X25519` cryptographic key pairs for performing Elliptic Curve Diffie-Hellman (ECDH) key exchange operations and the `X25519KeyAgreementKey2019` verification method type.

There are some open questions about the design of the `KeyExchange` trait which uses `PhantomData` rather than generics (in line with current conventions for `Sign` and `Verify`) and several TODOs related to removing `MerkleKeyCollections` in #715.

### Added

- Add `MethodType::X25519KeyAgreementKey2019`.
- Add `KeyType::X25519`.
- Add `MethodSecret::X25519`.
- Add `X25519` `KeyExchange` implementation.
- Add Diffie-Hellman key exchange example.
- Add `MethodData` Wasm bindings.

### Changed

- Change `KeyPair::try_from_ed25519_bytes` to `try_from_private_key_bytes`.
- Rename `MethodData::new_b58` to `new_base58`.
- Rename `VerificationMethod::key_type` to `type_`.
- Rename `VerificationMethod::key_data` to `data`.

### Removed

- Remove `KeyPair::new_ed25519`.

Fixes #671.

## Type of change

- [ ] Documentation Fix

Local tests and new examples work.

## Change checklist


identity-core/src/crypto/exchange/key_exchange.rs
```diff
@@ -0,0 +1,20 @@
...
+  type Private: ?Sized;
+  /// Corresponding public key type for the second party.
+  type Public: ?Sized;
+  /// Shared secret output type of the key exchange process.
+  type Output;
+  /// Error type on failure.
+  type Error;
+
+  /// Performs a cryptographic key exchange process (e.g. Diffie-Hellman) using the private key
+  /// of the first party with with the public key of the second party, resulting in a shared secret.
```
HenriqueNogara, line 18: duplicated with

HenriqueNogara (approved):
LGTM

identity-core/src/utils/ed25519.rs
```diff
@@ -11,6 +11,7 @@ use crate::error::Result;
 ///
 /// Note that the private key is a 32-byte seed in compliance with [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-3.2).
 /// Other implementations often use another format. See [this blog post](https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/) for further explanation.
+// TODO: move or remove
```
PhilippGackstatter, line 14: Can we inline this into `KeyPair::new`, like how it is with `X25519` now?

identity-did/src/verification/method_type.rs
```diff
@@ -11,16 +11,20 @@ use crate::error::Result;
 /// Supported verification method types.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
 #[repr(u32)]
+// TODO: use string representations to avoid breaking changes from reordering.
```
PhilippGackstatter, line 14: Aren't we already using string representations? In any case, I think we could remove the `#[repr(u32]` and `as_u32` and `from_u32` methods after #729, as it's only used in the current `KeyLocation`.

PhilippGackstatter (approved):
Looks good to me!

Some comments regarding the outstanding questions.

identity-core/src/utils/ed25519.rs
cycraig: It's used for `KeyCollection` I think, so it will be removed with the `MerkleKeyCollection` removal.

identity-did/src/verification/method_type.rs
cycraig: For `serde` yes we still use string representations since `repr(u32)` alone doesn't change the serialization. But `wasm-bindgen` uses numeric representations for flat enums in the Wasm bindings which is another thing I'm concerned about breaking (I don't think there's really a way around that though).

I think I'll go ahead and remove the `repr(u32)` and associated methods at least.

identity-core/src/crypto/key/pair.rs
```diff
@@ -12,9 +13,11 @@ use crate::crypto::PrivateKey;
 use crate::crypto::PublicKey;
 use crate::error::Result;
 use crate::utils::generate_ed25519_keypair;
-use crate::utils::keypair_from_ed25519_private_key;
 
 /// A convenient type for representing a pair of cryptographic keys.
+// TODO: refactor with exact types for each key type? E.g. Ed25519KeyPair, X25519KeyPair etc.
+//       Maybe a KeyPair trait with associated types? Might need typed key structs
+//       like Ed25519Public, X25519Private etc. to avoid exposing pre-1.0 dependency types.
```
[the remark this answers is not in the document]
cycraig: Yeah I'm hesitant to do a `KeyPair` refactor at this point since it has sweeping implications everywhere. One concern I have is that with introducing specific types we multiply the number of structs we have to make bindings for.

E.g. now we have `KeyPair` with `Public` and `Private` keys represented as strings/`UInt8Array` soon, so basically one type exposed to the bindings. With a refactor we might have 6 different structs:

- `Ed25519KeyPair`
- `Ed25519Private`
- `Ed25519Public`
- `X25519KeyPair`
- `X25519Private`
- `X25519Public`

And more for each new `KeyType` we choose to support. Note that the specific types for keys may be important for places where we only want/need the public key, such as creating a new verification method. We've largely moved away from requiring `KeyPairs` in function parameters where the private key is not required.

If we do go ahead with such a refactor I think a `KeyPair` trait would be useful for generalization, e.g.

```rust
trait KeyPair {
  type Public;
  type Private;

  fn public(&self) -> &Self::Public;
  fn private(&self) -> &Self::Private;
}
```

I don't think that will be in this PR, however.

identity-core/src/crypto/exchange/x25519.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use core::marker::PhantomData;
+
+use crypto::keys::x25519;
+
+use crate::crypto::KeyExchange;
+use crate::Error;
+
+/// An implementation of `X25519` Elliptic-curve Diffie-Hellman (ECDH) cryptographic key exchange.
+// TODO: this phantom data workaround sucks, maybe make key_exchange generic instead?
```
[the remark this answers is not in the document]
cycraig: Thanks for the suggestions!

> Do we know sufficient other key exchange types to design an abstraction for this now? 

The only other ones I'm aware of are RSA, which has broadly been phased out in favour of elliptic-curves which are faster with smaller key sizes, so I don't think we'll have to support it. 

The problem with using generics directly rather than associated types is that generics mostly prevent specifying a particular struct, which associated types do.

E.g. `PUB: AsRef<X25519Public>` only works is `AsRef` is implemented manually for `X25519Public` since it is not a default trait, while `PUB: X25519Public` throws an error. An associated type such as `type PUB = X25519Public` works fine, however. I believe this is why existing traits like `Sign`, for instance, use associated types for input parameters.

There are three option basically:

1. Keep the `KeyExchange` trait with associated types and `PhantomData` in `X25519`.
2. Change the `KeyExchange` trait to use generics for input parameters, removing the need for `PhantomData`.
3. Remove the `KeyExchange` trait entirely and implement `key_exchange` for `X25519` directly.

I'm still very undecided, any of the options would be acceptable in my opinion.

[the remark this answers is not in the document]
PhilippGackstatter: > I believe this is why existing traits like `Sign`, for instance, use associated types for input parameters.

Yeah, that make sense.

I'm in favor of option 3. Implementing an interface that we may have to break in the future, because its abstraction was based on just this case, is probably not worth that much. I'd go with the simple option for now and forgo the trait.
If you want to keep the trait, option 2 sounds somewhat preferable to option 1.

identity-core/src/crypto/key/pair.rs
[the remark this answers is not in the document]
PhilippGackstatter: Thanks for the explanations. It's fine to leave this for a future PR. Turning `KeyPair` into a trait is what I had in mind, too, that sounds good! I agree that we need to carefully consider the impact before going ahead, specifically for Wasm.

identity-did/src/verification/method_type.rs
cycraig: Removed the `u32` representation and methods.

identity-core/src/crypto/exchange/x25519.rs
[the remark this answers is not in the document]
cycraig: Removed the `KeyExchange` trait, moved `key_exchange` to `X25519`.

Issue #669, opened 2022-02-22
[Task] Wasm Bindings Improvements  


HenriqueNogara:
## Description
Remove exposure of Ed25519.sign, remove `Storage` `Send/Sync` ties to the `wasm` feature, and make keyDel in typescript compatible with the Rust code. Refactor the `Storage` trait to return better errors, and simplify its API.

## Motivation
1. Users should be able to sign data without relying on our code.  
2. We might need the Storage trait not to be Send/Sync for other bindings (e.g Python) as well, so we should not tie its bounds to the `wasm` feature. 
3. Js implementation of `MemStore` `keyDel` slightly differs from the Rust code regarding idempotency. 
4. `identity_account::types::signature::Signature::pkey` is not used in the Rust code.
5. `Storage::set_password` is no longer needed.
6. Improve error feedback. See: https://github.com/iotaledger/identity.rs/pull/597#discussion_r812013066

## To-do list
- [x] Assess the need to expose Ed25519.sign - @noble/ed25519 might provide the correct signature
- [ ] Remove Storage ties to `wasm` feature
  - [ ] Change the way the Account handles Storage
  - [x] Use the `cfg_attr` on Storage and change the name of the feature that enables it
- [x] Decide whether `Storage::key_del` should be idempotent or not. #729
- [x] Refactor `Account` `Signature` not to contain `PublicKey` https://github.com/iotaledger/identity.rs/issues/722
- [x] Remove `SetPassword` method from `Storage` https://github.com/iotaledger/identity.rs/issues/720
- [ ] Investigate simplification of state in the `Storage` interface, potentially reducing the number of state variables
- [ ] [Refactor  `Error` for `Storage` Interface](https://github.com/iotaledger/identity.rs/pull/597#discussion_r812013066)
  - [ ] Introduce a reasonably abstract error type that can accommodate many `Storage` implementations, including any type of file- or network-I/O-related failures
- [x] Prefer to use `UInt8Array` rather than Base58-BTC encoded strings for keys. https://github.com/iotaledger/identity.rs/issues/721
- [x] [Write macro for `napi` related types and impl](https://github.com/iotaledger/identity.rs/pull/660#discussion_r816896110) - #872
- [ ] [Assess `NapiResult` error messages](https://github.com/iotaledger/identity.rs/pull/660#discussion_r816933088)
- [x] [Avoid Serializing and Deserializing keys](https://github.com/iotaledger/identity.rs/pull/681#discussion_r826991675)

[...]

PR #738, opened 2022-03-21
Refactor Storage Signature

### LINKED ISSUE #722: [Task] Refactor `Account` `Signature`
## Description
Refactor the `Account` `Signature` struct to either an enum, or the raw bytes of the signature, or simply `Vec<u8>`/`UInt8Array`.

## Motivation
The `public_key` field in `Signature` is unused. Currently supported signing methods (`Ed25519`, `MerkleKeyCollection`) only require a single byte array to be embedded in the `proof` for the signature so we can just return that alone.

Note that future signature schemes may require extra information to be returned, so an enum might be more prudent to avoid breaking changes in the future, optionally.

Note that `UInt8Array` may cause problems with the Napi Stronghold bindings.

## Resources
This is a sub-task extracted from:
- https://github.com/iotaledger/identity.rs/issues/669

## To-do list

- [ ] Decide whether Signature should be a type alias for `Vec<u8>` or a newtype wrapper, and consider the effects on bindings in that decision.
- [ ] Replace `Account` `Signature` with a byte array or enum as appropriate.

## Change checklist

HenriqueNogara:
# Description of change
Change `account-core` `Signature` to new type struct, removing unused `PublicKey`. Remove `Serialize/Deserialize` for `Key` and `bindings-derive` feature

#669

fixes #722 

## Type of change
- [ ] Documentation Fix

All existing tests and examples have been run

## Change checklist


identity-account-storage/src/types/signature.rs
```diff
@@ -1,33 +1,20 @@
...
-  /// Returns the public key used to verify this signature.
-  pub fn pkey(&self) -> &PublicKey {
-    &self.pkey
-  }
-
-  /// Returns the signature data as a slice of bytes.
-  pub fn data(&self) -> &[u8] {
-    &self.data
+  /// Consumes the signature and returns it as bytes
+  pub fn as_bytes(self) -> Vec<u8> {
```
PhilippGackstatter, line 15: Not sure why clippy's [`wrong_self_convention`](https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention) isn't triggered here, this should be called `into_bytes`. Not sure about the JS method you're binding to, but I would probably go with `toBytes`.

identity-account-storage/src/types/signature.rs
```diff
@@ -1,33 +1,20 @@
...
-/// A digital signature and associated public key.
-#[cfg_attr(feature = "bindings-derive", derive(Clone, Deserialize, Serialize))]
-#[cfg_attr(not(feature = "bindings-derive"), derive(Clone))]
-pub struct Signature {
-  pub(crate) pkey: PublicKey,
-  pub(crate) data: Vec<u8>,
+/// A digital signature.
+#[derive(Clone, Deserialize, Serialize)]
+pub enum Signature {
+  Ed25519(Vec<u8>),
```
cycraig, line 10: I think this design is good for forwards-compatibility. Should we mark the enum as `non_exhaustive` so we can add variants without causing a breaking change?

PhilippGackstatter (changes requested):
It looks like you have a commit related to the release, that is not on `dev` somehow? Maybe you can check with @eike-hass what's going on.

Maybe I missed the conversation, but why is `Signature` an enum? I think we can just make the signature a wrapper around bytes.

identity-account-storage/src/types/signature.rs
```diff
@@ -1,33 +1,20 @@
...
-  }
-
-  /// Returns the signature data as a slice of bytes.
-  pub fn data(&self) -> &[u8] {
-    &self.data
+  /// Consumes the signature and returns it as bytes
+  pub fn as_bytes(self) -> Vec<u8> {
+    match self {
+      Signature::Ed25519(bytes) => bytes,
+    }
```
cycraig, line 18: Not too sure about the name, `as_*` usually doesn't consume `self`. Maybe `into_bytes` would be more appropriate? It's unclear whether all variants will be able to satisfy this though...

identity-account-storage/src/types/signature.rs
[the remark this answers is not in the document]
HenriqueNogara: We can, but if we want to add support to something else latter on, we would have to change it.

cycraig (changes requested):
This has a bunch of release changes, this may have to be rebased to remove them since I think `dev` was force-pushed.

Should we consider using `SignatureValue` from `identity-core` to avoid the new type altogether or does the `SignatureValue::None` variant make it unusable in this context?

identity-account-storage/src/types/signature.rs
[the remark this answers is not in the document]
PhilippGackstatter: I assume any signature can be represented as bytes, and what type of signature it is, is clear from context, e.g. when calling `key_sign` with a `KeyLocation` that contains a certain `MethodType`. So having to do the additional match seems unnecessary, unless the `Storage` implementation is simply wrong and returns something else. It's maybe slightly safer, but I don't really see a major advantage.
At least judging by `RemoteSign`, I think it's better not to use `SignatureValue`, so we don't accidentally return _that_ signature value, but have to first encode it as base58 and then return an instance of `SignatureValue`.

https://github.com/iotaledger/identity.rs/blob/1271d951cbb6a1ec2940e99d12015b6459a78aaa/identity-account-storage/src/crypto/remote.rs#L53-L56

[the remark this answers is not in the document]
cycraig: > I assume any signature can be represented as bytes

I'm not sure that's a strong assumption. For instance the [BBS+ Signatures 2020 spec requires a nonce value used to derive the proof](https://w3c-ccg.github.io/ldp-bbs2020/#nonce) as well, so it's probable that we will need more than just the signature/proof bytes in the future. One could argue that the `nonce` could be predetermined and passed to the `sign` function but that just moves the problem to the signing interface not being general enough to handle it.

> At least judging by RemoteSign, I think it's better not to use SignatureValue,

I agree, `SIgnatureValue` does not seem fully appropriate for this as it would require `Storage` implementers to also implement the signature schemes such as `JCSEd25519Signature2020` and handle encoding rather than just the raw key signing.

I'm fine using just `Vec<u8>` or a similar new-type wrapper, until we need to refactor it to handle a new method type, if we cannot come to a consensus on what a future-proof design looks like now.

cycraig (approved):
Looks good to me!

cycraig (commented):
Oh, maybe change the PR title to fit the changelog better. Something like "Refactor `Storage` `Signature`".

Edit: and add the `Wasm` label, it affects the bindings too.

bindings/wasm/src/account/types/signature.rs
```diff
@@ -12,22 +12,16 @@ pub struct WasmSignature(pub(crate) Signature);
...
   }
 
-  #[wasm_bindgen]
-  /// Returns a copy of the signature data as a vec of bytes.
-  pub fn data(&self) -> Vec<u8> {
-    self.0.data().to_vec()
+  /// Returns a copy of the signature as a `UInt8Array`.
+  #[wasm_bindgen(js_name = asBytes)]
+  pub fn as_bytes(&self) -> Vec<u8> {
+    self.0.as_bytes().to_vec()
```
PhilippGackstatter, line 24: Nit: I still think this should be `toBytes` in JS, mostly due to the Rust naming convention reasoning of "to" representing a conversion. But since I'm not familiar with JS naming conventions in that detail, I'll leave it up to you.

PhilippGackstatter (approved):

PR #764, opened 2022-03-24
Replace `MethodSecret` with `MethodContent` enum

### LINKED ISSUE #646: [Request] Allow creating methods by public key in `Account`, Wasm bindings
## Description
The low-level API allows creating an `IotaVerificationMethod` by a public key alone, namely the `from_did` method:

```rust
/// Creates a new [`IotaVerificationMethod`] from the given `did` and `keypair`.
pub fn from_did(did: IotaDID, key_type: KeyType, public_key: &PublicKey, fragment: &str) -> Result<Self> 
```

The `Account` and Wasm bindings should be updated to expose similar functionality.

In particular, the `WasmVerificationMethod::fromDID` method still uses a `KeyPair` rather than just the public key:
```rust
/// Creates a new `VerificationMethod` object from the given `did` and `key`.
#[wasm_bindgen(js_name = fromDID)]
pub fn from_did(did: &WasmDID, key: &KeyPair, fragment: String) -> Result<WasmVerificationMethod, JsValue>
```

The `Account` `CreateMethod` update builder also only allows inputting a secret, or generating one automatically:

```rust
CreateMethod {
  scope: MethodScope,
  type_: MethodType,
  fragment: String,
  method_secret: Option<MethodSecret>,
},
```

## Motivation
There are several use-cases for creating a `VerificationMethod` where one would not have access to the underlying private key to create a `KeyPair`:

- in a distributed setting where the private key is not on the device updating the DID Document, e.g. IoT devices, phone apps, controllers etc.
- when enabling another party to update your DID Document without accessing their private key, i.e. controller workaround.

Furthermore the private key in the `KeyPair` is usually completely unnecessary, and `KeyPair` is just used out of convenience since it wraps both the public key and the key type in one.

The `Account` does not actually need to store the private key in either of the above scenarios since the signer is another entity, and trying to sign with that method should just throw an error if the key is not present, which is fine.

## Requirements

1. ~~Update `WasmVerificationMethod::fromDID` method to match the `IotaVerificationMethod::from_did` parameters.~~ Done in #655 
2. Alter the `CreateMethod` update builder for the `Account` to allow creating methods with just a public key.

## Open questions (optional)
Is there a convenient design for `CreateMethod` to take either a `MethodSecret` or `PublicKey` (+ `KeyType`) without conflicting at runtime? Another enum level perhaps?

## Are you planning to do it yourself in a pull request?

If no-one else picks this up.

cycraig:
# Description of change
Replace `MethodSecret` with `MethodContent` to allow creating verification methods by public key, without generating or inserting a corresponding private key in the `Account`.

This also changes `Account` `create_identity` to take `Option<PrivateKey>` rather than `MethodSecret` since it will always require an Ed25519 private key and it does not fit the new `MethodContent`.

These changes also prevent `MethodType`/`MethodSecret` mismatches at runtime.

### Added

- Add `MethodContent` for `CreateMethod` updates in the `Account`, replacing `MethodSecret`.

### Changed

- Replace `IdentitySetup` `key_type` and `method_secret` with `private_key`.

### Removed

- Remove `MethodSecret`.

Fixes #646.

## Type of change

- [ ] Documentation Fix

Tests and examples pass locally.

## Change checklist


cycraig:
Stronghold tests in CI should be fixed by #767 but it raises questions about whether the current `Vec<u8>` inner representation for `MethodContent` will work with `into_serde` due to `Uint8Array` problems.

Will need to add more tests to confirm, this problem seems pervasive.

identity-account/src/types/method_content.rs
```diff
@@ -0,0 +1,66 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use identity_core::crypto::KeyType;
+use identity_core::crypto::PrivateKey;
+use identity_core::crypto::PublicKey;
+use identity_did::verification::MethodType;
+
+/// Method content for creating new verification methods.
+#[derive(Clone, Debug)]
+pub enum MethodContent {
```
PhilippGackstatter, line 11: Can you remind me what the problem with this was (as an effort to reduce how fast the enum needs to grow with future key types):

```rust
pub enum MethodContent {
  Generate(KeyType),
  PrivateKey(KeyType, PrivateKey),
  PublicKey(KeyType, PublicKey),
}
```

As it is now, we also don't have a guarantee that e.g. `PrivateEd25519(PrivateKey)` actually contains the correct number of bytes for an ed25519 key, so I'm wondering if we actually lose any validation with the above approach.

If we keep it as-is, shouldn't it be `#[non_exhaustive]`?

PhilippGackstatter (commented):
Looks good to me, just one question around `MethodContent`.

identity-account/src/updates/update.rs
```diff
@@ -305,72 +296,70 @@ impl Update {
...
 /// - `scope`: the scope of the method, defaults to [`MethodScope::default`].
 /// - `fragment`: the identifier of the method in the document, required.
-/// - `method_secret`: the secret key to use for the method, optional. Will be generated when omitted.
+/// - `content`: the key material to use for the method or key type to generate.
 CreateMethod {
-  @defaulte type_ MethodType = Ed25519VerificationKey2018,
   @default scope MethodScope,
   @required fragment String,
-  @optional method_secret MethodSecret
+  @required content MethodContent,
```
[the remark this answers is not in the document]
cycraig: I agree. It previously did default to generating an Ed25519 key but I didn't think that a default would achieve much here: the fragment is always required anyway (i.e. one couldn't omit all parameters in Javascript anyway) and explicitness seems preferred.

identity-account/src/types/method_content.rs
cycraig: It might be possible to go with a simpler enum if we refactor `PrivateKey` and `PublicKey` to be traits with concrete implementations for `X25519PublicKey`, `Ed25519PrivateKey` etc.

Will keep the current naming scheme.

PhilippGackstatter (approved):

HenriqueNogara (approved):

PR #771, opened 2022-03-26
Prepare for new wiki build system

Dr-Electron:
# Description of change

This is a first draft of the new wiki build system. It should make things a lot easier for devs. We will also provide a v2 of the wiki-cli to make things even easier ;)
I will keep this a draft until it is ready, but now if you want you can already look at what we are creating.
The wiki can be used by just running `yarn`, `yarn start` and navigating to http://localhost:3000/identity/introduction

TO-DOs:
- [x] Test
- [x] Add v2 of Wiki CLI

## Type of change

- [x] Documentation Fix


Dr-Electron:
As I already said, this is now really although there is still one little bug.
As of now `yarn iota-wiki start` doesn't work. We will fix it soon.
Until then you can use `yarn start` to test your docs

documentation/.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: .yarn/releases/yarn-3.2.0.cjs
```
eike-hass, line 1: Is this and the `yarn/releases` folder checked in to make sure people have yarn available? If so I think would be okay to have yarn a prerequisite, if it cannot be installed as (dev-)dependency.

Dr-Electron: I'm actually not surr anymore why I added this. I think it was because it doesn't work with older versions of yarn 🤔
I can test that again I guess 🤔

Dr-Electron: Ok. I can't figure out why I used yarn 3.2.0. I'm going to remove it 😅

jlvandenhout: Modern yarn puts the yarn distribution in the repo to be checked in. This has the benefit that everyone using this repo uses the exact same version.

Dr-Electron:
Ok, most of the things should work now :trollface: 
You can test the docs locally with `yarn start`
I also added a test build to check for broken links but I need to disable that until https://github.com/iotaledger/identity.rs/issues/804 is solved I guess, because as of now the api reference contains some "broken links" as you can see in the failed build 😅. All other broken links were fixed by me.
The parameter to break on broken linke is [this one](https://docusaurus.io/docs/api/docusaurus-config#onBrokenMarkdownLinks)

Dr-Electron:
> > I also added a test build to check for broken links but I need to disable that until #804 is solved I guess,
> 
> I think it was decided that #804 is blocked by this. Might have to merge with broken md links in the Wasm API reference and deal with it later.

Yeah, makes sense. That's why I disabled it for now and you can just enable it again once it is ready :)

Dr-Electron:
Also please inform the Wiki team (on discord) once you plan to merge as we need to update the wiki that day too :)

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,21 @@
+name: Test Docs Build
+
+on:
+  pull_request:
+    branches:
+      - support/v0.5
```
cycraig, line 6: Ideally we would want this to run on all PRs (will be merged back to `dev` from the `support/v0.5` branch). Anything else before this can be merged, @eike-hass ? I haven't tested anything.

```suggestion
on:
  push:
    branches:
      - main
      - dev
  pull_request:
    branches:
      - main
      - dev
      - epic/*
      - support/*
```

Dr-Electron: Maybe I haven't thought this through 100% either :trollface:

cycraig: That makes sense. Does the wiki automatically pull in those branches or is it pinned to `support/v0.5` on your side for now? 

I was just concerned about PRs breaking the wiki over time and only noticing during/after a release.

Dr-Electron: No. As soon as you merge this PR, I will add the `support/v0.5` branch as a git submodule to the wiki.

Eventually the new cli will have an easy version managing tool which allows you to add/remove versions in the wiki repo. We will also use gitpod as a tool there so that you won't need to clone the wiki repo and testing and making those changes is as less overhead for you as possible.

Dr-Electron: I mean we only pull from a specific set of branches, right? So if those branches are checked for braking links we will never have broken links in the wiki.
The problem is mainly if you are creating bigger feature branches that you won't to merge after months of coding on it and then you realise basically your whole branch contains broken links and you need to fix them before you merge into a release branch or create a new release branch :thinking:

Dr-Electron: Makes sense 👍

Dr-Electron:
I decided to add the new cli build command as it makes stuff easier. But I basically needed to remove the test action for now. I left a comment on what you need to do once broken links are fixed

.github/workflows/test-docs-build.yml
eike-hass: The workflow will only run on a branch if it is present on it. So we would merge the workflow together with the build system. I think @cycraig initial thought and suggestion is correct. Will commit the suggestion.

eike-hass (approved):
local build works. LGTM!

PR #787, opened 2022-04-01
Upgrade to new `Stronghold` interface

### LINKED ISSUE #785: [Task] Upgrade to new Stronghold interface
## Description

Upgrade Stronghold to the new client interface.

## Motivation

The new interface allows us to get rid of our password management solution and a big chunk or even all of our custom stronghold wrapper.

## To-do list

- [ ] Use stronghold's native password management.
- [ ] Use the new client interface.

## Change checklist

PhilippGackstatter:
# Description of change

Reimplements our `Stronghold` wrapper with the new stronghold interface.

Removes the old wrapper completely.

For now, the `Stronghold` wrapper takes care of creating the parent directories of a stronghold snapshot file before attempting to persist it. Stronghold should probably handle this itself, but it currently doesn't.

## TODO:
- change workflow include in `build-and-test.yml:207` back to `dev`

## Open Questions

- Should `Stronghold::new` only take path and password, without dropsave? Should we then add a second method `Stronghold::new_with_options` that takes a `StrongholdOptions` struct where users can set the dropsave? This would make things easily extendable in the future, but so far it doesn't look like we need a second option, so it's arguably premature optimization.
  - One possible second useful option might be for users to explicitly state whether the file path they provide is supposed to point to an existing stronghold, or if it should be created if it doesn't exist. Just like `std::fs::OpenOptions::create`. E.g. if it doesn't exist, then an error can be returned.
  - This would technically make this PR a breaking change.
- Do we have to roll back changes to a client if the entire did_create "transaction" doesn't complete successfully? Otherwise, someone re-trying to create an identity from an existing keypair might fail every time, because the DID was added to the index, but e.g. the client syncing later failed. Generally, we don't go to those lengths, so this might be fine?

fixes #785

## Type of change

- [ ] Documentation Fix

- The storage test suite was added for stronghold.
- Two tests were added to test the custom functionality of the `Stronghold` wrapper, such as `mutate_client` and an error case.
- A local test was executed with the current `dev` version and the current PR to ensure the storage layout from the `dev` stronghold snapshot files can continue to be used with the new `Stronghold` wrapper.
  - That test ran a modified example from `dev` that added a key and a service and wrote to a stronghold file. On the current PR, an example was run to ensure the DID could be loaded, sign with the added key, and the service could be removed.
- `identity_account::tests::util::storages` now also returns a `Stronghold`, so all update tests for the account run once with a `MemStore` and once with a `Stronghold`.

## Change checklist


PhilippGackstatter:
The multiple identities example using Napi Stronghold returns a lock acquisition error, that is tracked (in part) here: https://github.com/iotaledger/stronghold.rs/issues/353.

cycraig (approved):
The significant simplification and reduction of code and dependencies is much appreciated.

I'm still not that familiar with Stronghold, and the code is pretty meticulous, so my feedback is mostly nits.

identity-account-storage/src/stronghold/wrapper.rs
```diff
@@ -0,0 +1,143 @@
...
+  where
+    T: AsRef<Path> + ?Sized,
+  {
+    let stronghold: IotaStronghold = IotaStronghold::default();
+
+    let mut key: EncryptionKey = derive_encryption_key(&password);
+    password.zeroize();
+
+    let key_provider = KeyProvider::try_from(key.to_vec()).map_err(StrongholdError::Memory)?;
+    key.zeroize();
```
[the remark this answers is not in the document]
cycraig: This change would be breaking, so if this PR is really just a patch it should be done elsewhere.

cycraig:
> Should Stronghold::new only take path and password, without dropsave? Should we then add a second method Stronghold::new_with_options that takes a StrongholdOptions struct where users can set the dropsave? [...] 

I was opposed to changing to this because, as you noted, `dropsave` is the only optional parameter. I do like the idea of a `create` flag to indicate whether an existing snapshot is expected.

If this is up for debate, the broad options are:
- `Stronghold::new` and `Stronghold::new_with_options` with a `StrongholdOptions` (would have to be a builder to avoid breaking changes).
    ```rust
    Stronghold::new_with_options("./snapshot.hodl", "password", StrongholdOptions::new().dropsave(false)).await?;
    ```
- Introducing a `StrongholdBuilder`, e.g.,
    ```rust
    Stronghold::builder()
      .path("./snapshot.hodl")
      .password("password")
      .dropsave(false)
      .create(false)
      .build()
      .await?;
    ```
- Placing everything in a non-exhaustive `StrongholdOptions` struct, embracing the TypeScript approach:
    ```rust
    Stronghold::new(StrongholdOptions {
      path: "./snapshot.hold".into(),
      password: "some-password".into(),
      create: false,
      ..Default::default()
    }),await?;
    ```

> Do we have to roll back changes to a client if the entire did_create "transaction" doesn't complete successfully? [...]

Ideally, yes. We should probably restore the old index and purge both the temporary and DID clients if an error occurs. There are unfortunately many places that can error there...

> The multiple identities example using Napi Stronghold returns a lock acquisition error, that is tracked (in part) here: https://github.com/iotaledger/stronghold.rs/issues/353.

Looks like the Ubuntu example is throwing that error too, do we want to exclude that example as an edge-case for now or delay this PR until it's fixed?

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -201,82 +201,90 @@ impl Storage for Stronghold {
...
+
+      if !exists {
+        return Ok(exists);
+      }
+
+      client
+        .execute_procedure(procedures::RevokeData {
+          location: location.into(),
+          should_gc: true,
+        })
```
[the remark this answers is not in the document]
PhilippGackstatter: Agree with what you're saying, but if we don't check for existence first, then we cannot accurately return whether the function deleted the key or not (the bool return parameter). So to comply with our interface, I'd leave it as-is.

PhilippGackstatter:
Opened #836 to address the comments from @cycraig, as they are non-essential or constitute breaking changes. The motivation is to get this PR merged sooner than later.

Similarly, I've disabled the multiple identites examples in Rust and stronghold-nodejs, because of iotaledger/stronghold.rs#353, and disabled the musl build (opened #837 for the required changes). Both will need to be addressed before the next release.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -335,69 +337,112 @@ impl Drop for Stronghold {
...
-  let procedure: procedures::PublicKey = procedures::PublicKey {
-    ty: location_key_type(location),
-    private_key: location.into(),
-  };
+fn insert_private_key(client: &Client, mut private_key: PrivateKey, location: &KeyLocation) -> Result<()> {
+  let stronghold_location: Location = location.into();
+
+  let vault: ClientVault = client.vault(stronghold_location.vault_path());
+
+  let private_key_vec: Vec<u8> = private_key.as_ref().to_vec();
```
felsweg-iota, line 358: Is there a way to leverage the `KeyStore` or Stronghold's [NC](https://github.com/iotaledger/stronghold.rs/blob/629466da83a677925904b2e733ef7bfad5a42864/engine/runtime/src/memories/noncontiguous_memory.rs#L83) types in favor of `PrivateKey` ? It somehow feels wasted to convert the private key to a vec, to just be zeroized and dropped afterwards.

identity-account-storage/src/stronghold/wrapper.rs
```diff
@@ -0,0 +1,143 @@
...
+    })
+  }
+
+  /// Returns whether dropsave is enabled.
+  pub fn dropsave(&self) -> bool {
+    self.dropsave
+  }
+
+  /// Sets whether dropsave is enabled.
+  pub fn set_dropsave(&mut self, dropsave: bool) {
```
felsweg-iota, line 88: ultra-nit: 
```suggestion
  pub fn dropsave_mut(&mut self, dropsave: bool) {
```

felsweg-iota (approved):
Integration of Stronghold looks good to me. Check the remarks, if the `PrivateKey`-type can be replaced, or would require to much effort. The current solution is Ok.

identity-account-storage/src/stronghold/wrapper.rs
PhilippGackstatter: Mh, this is just a setter, and does not return a mutable reference to `dropsave`, so the naming matches Rust conventions. `clippy` would probably complain about `dropsave_mut`, too.

identity-account-storage/src/storage/stronghold.rs
PhilippGackstatter: Appreciate the idea! Unfortunately, this would "leak" the stronghold dependency to our `Account` which is a no-go for Wasm. If you could factor out these NC types in a separate small crate that is also Wasm compatible, we could reconsider :wink:. Perhaps this would improve Wasm memory security, too.

HenriqueNogara (approved):

identity-account-storage/src/stronghold/error.rs
```diff
@@ -1,37 +1,104 @@
...
-  #[error("Stronghold resource index malformed")]
-  InvalidResourceIndex,
+  #[error("failed to `{0}` stronghold client `{1}`")]
+  Client(ClientOperation, ClientPath, #[source] ClientError),
+  #[error("store `{0}` operation failed")]
+  Store(StoreOperation, #[source] ClientError),
+  #[error("vault operation `{0}` failed")]
+  Vault(VaultOperation, #[source] ClientError),
+  #[error("procedure `{0}` operating on locations {1:?} failed")]
+  Procedure(ProcedureName, Vec<KeyLocation>, #[source] ProcedureError),
```
HenriqueNogara, line 29: Not all procedures will be dealing with `KeyLocation`, since this is either an `Ed25519` or an `X25519` Key for now. For example, when dealing with the procedures related to shared secrets the type is neither. However, we are dealing with random locations in that case. Should we change `KeyLocation` or is it fine to have an empty location in these cases?

cycraig: Would changing it to `Vec<String>` account for shared secrets? I think `Location` alone doesn't have a string representation.

PhilippGackstatter: If the need arises to return errors that include non-`KeyLocation`s, then Craig's suggestion is great. Though, I think for debugging it should be helpful enough to include the `KeyLocation`s, as the other, temporary locations will probably be completely random or static, and we'll be able to trace the issue based on the procedure name alone, I imagine. From how I understand `execute_procedure_chained`, it will delete those temp locations by itself, so we shouldn't have to deal with potentially failing `RevokeData` procedures, either: https://github.com/iotaledger/stronghold.rs/blob/dev-refactor/client/src/types/client.rs#L247-#L254

HenriqueNogara: > temporary locations will probably be completely random or static, and we'll be able to trace the issue based on the procedure name alone, I imagine.

I thought so too. For now, I won't change the type, but will change it to `String` in case I need it.

Issue #757, opened 2022-03-23
[Task] Implement key deletion in the `Account`


PhilippGackstatter:
## Description

The `Storage::key_del` function can be used to delete the key belonging to a verification method, but it is currently not used from the `Account`. The task is to decide when to delete the key material of a verification method, and implement it.

## Motivation

The key material of deleted verification methods does not need to be retained, and so it should be deleted to keep the storage clean.

## When to delete keys?

In theory, keys can be deleted as soon as the verification method has been deleted. One caveat comes for methods that are used to sign updates. Rotating DID document signing keys means deleting an old method from the document, but the old method is still required to sign the update that removes itself. Thus, deleting keys immediately when processing `Update::DeleteMethod` is not possible in general.

In theory, publishing waits for milestone confirmation, so it is safe to delete keys afterwards.

One approach (I believe suggested by @cycraig) was to compare (diff?) the old and new document during publishing, and if publishing is successful, attempt to delete all keys that were removed. "Attempt", because in a distributed setup, not all verification methods may actually have local key material. Considering distributed setups, keys also need to be removed during `Account::fetch_document`, because methods may have been removed from a different machine. A nice feature of this approach is that it should not require additional state to be kept.

This approach may or may not have to be discussed some more.

## To-do list
Create a task-specific to-do list. Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ ] Implement key deletion in the account.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

cycraig:
> compare (diff?) the old and new document during publishing, and if publishing is successful, attempt to delete all keys that were removed.

Indeed that's one of the approaches we could use. I think I raised that we should only attempt that after resolving to ensure the latest state does not contain the keys (only resolving really confirms that a published update was successful), and resolving might be something we want to do for that reason anyway.

[...]

PR #791, opened 2022-04-04
Implement `Storage` test suite

### LINKED ISSUE #790: [Task] Implement a `Storage` test suite
## Description

Implement a test suite for the `Storage` interface.

## Motivation

With multiple internal implementations of our `Storage` interface, as well as the ability for users of our framework to implement their own storage, it makes sense to provide a generic test suite that tests the interface's implementation.

## To-do list

- [ ] Implement tests for the `Storage` test suite.
- [ ] Expose the tests in Rust and bindings.

## Change checklist

PhilippGackstatter:
# Description of change

With multiple internal implementations of our `Storage` interface, as well as the ability for users of our framework to implement their own storage, this PR adds a generic test suite that tests the interface's implementation.

Because we cannot use the `assert` macros in Rust for testing bindings, custom macros were added that return an `anyhow::Error` on failure. This also means that even in Rust we don't get exact line numbers for failure, so we need to add context to an error to help users identify the issue. Whether the current context is enough is debatable. Calls that include `anyhow::Context::context` do not add the test function's name to the error, while the `ensure!` macros do.

We could go the extra mile and return something like a test report that includes statements about what the test is doing and expecting, so users have _a lot_ more context to work with.

For consideration, here are the outputs:

Errors when an interface function returns an unexpected error in Rust:

```
thread 'storage::memstore::tests::test_memstore_did_create' panicked at 'called `Result::unwrap()` on an `Err` value: did_create returned an error

Caused by:
    Shared resource poisoned: read', identity-account-storage/src/storage/memstore.rs:269:8
```

and Wasm:

```
Err > did_create returned an error: javascript function threw an exception: JsValue(Error: custom exception
Error: custom exception
    at MemStore.<anonymous> (/home/morgan/git/identity.rs/bindings/wasm/examples-account/src/memory_storage.ts:27:19)
    at step (/home/morgan/git/identity.rs/bindings/wasm/examples-account/src/memory_storage.ts:35:23)
    at Object.next (/home/morgan/git/identity.rs/bindings/wasm/examples-account/src/memory_storage.ts:16:53)
    at /home/morgan/git/identity.rs/bindings/wasm/examples-account/src/memory_storage.ts:10:71
    at new Promise (<anonymous>)
    at __awaiter (/home/morgan/git/identity.rs/bindings/wasm/examples-account/src/memory_storage.ts:6:12)
    at MemStore.didCreate (/home/morgan/git/identity.rs/bindings/wasm/examples-account/src/memory_storage.ts:51:16)
    at module.exports.__wbg_didCreate_2f3e066022bb91f5 (/home/morgan/git/identity.rs/bindings/wasm/node/identity_wasm.js:7575:31)
    at null.<anonymous> (wasm://wasm/00ecefee:1:484460)
    at null.<anonymous> (wasm://wasm/00ecefee:1:1406135))
```

Errors when an assertion isn't met in Rust:

```
thread 'storage::memstore::tests::test_memstore_did_create' panicked at 'called `Result::unwrap()` on an `Err` value: [storage_did_create_test]: expected key at location `IWPKX3l5cwT2eCI21LKWNMVKclUdRDuX:5631958722719088083` to exist', identity-account-storage/src/storage/memstore.rs:269:8
```

and Wasm:

```
Err > [storage_did_create_test]: expected key at location `WMMjx4oYaRZejYllEHmV0ZxdzFCdenax:911318764047275679` to exist
```

fixes #790

## Type of change

- [ ] Documentation Fix

Tests are run in Wasm examples, stronghold-nodejs tests and for the `MemStore`.

`Stronghold` is left out and will be added with #787. The current stronghold is not easily testable because there is an issue with concurrent usage of `Stronghold`s. However, the storage tests were executed for stronghold locally and passed.

The `didPurge` test in stronghold-nodejs is skipped because it relies on unimplemented functionality, see #757.

## Change checklist


identity-account-storage/src/storage/tests.rs
```diff
@@ -0,0 +1,503 @@
...
+    "expected key_public to return `{:?}`, returned `{public_key:?}`",
+    keypair.public()
+  );
+
+  let network: NetworkName = Network::Devnet.name();
+  let (did, location): (IotaDID, KeyLocation) = storage
+    .did_create(network.clone(), &fragment, None)
+    .await
+    .context("did_create returned an error")?;
+
```
cycraig, line 118: One downside of having single test functions for e.g. `did_create` is that they each test multiple behaviours of those functions. Like here, if I got this error back as a `Storage` implementer I wouldn't know that this error was caused by the case where `None` is passed instead of a predefined `PrivateKey`.

Only options I can think of are: adding even more context to the errors, e.g. `.context("did_create with no private key failed")` or similar; or splitting the top-level test functions into e.g. `didCreateGeneratePrivateKeyTest` and `didCreateNoPrivateKeyTest`, or something.

identity-account-storage/src/storage/tests.rs
```diff
@@ -0,0 +1,503 @@
...
+  for (i, did) in dids.into_iter().enumerate() {
+    let purged: bool = storage.did_purge(&did).await.context("did_purge returned an error")?;
+
+    ensure!(purged, "expected the did `{did}` to be purged");
+
+    let exists: bool = storage.did_exists(&did).await.context("did_exists returned an error")?;
+    ensure!(!exists, "expected did `{did}` to no longer exist");
+
+    let list_len: usize = storage.did_list().await.context("did_list returned an error")?.len();
+    let expected_len: usize = NUM_IDENTITIES - (i + 1);
```
cycraig, line 283: This is kind of prevalent throughout the tests but calling other interface function makes it slightly more difficult for implementers to test functions independently (but I consider that mostly fine since these are meant to be comprehensive). I'm only mentioning it here because in this case `did_purge` should not be required for the `did_list` function, since `Storage` implementers may choose not to implement `did_purge` at all (for various reasons).

Suggestion: move these checks to the `did_purge` test rather.

identity/Cargo.toml
```diff
@@ -47,6 +47,8 @@ stronghold = ["identity-account/stronghold", "identity-account-storage/stronghol
 # Enables send and sync storage
 send-sync-storage = ["identity-account-storage/send-sync-storage"]
 
+storage-test-suite = ["identity-account-storage/storage-test-suite"]
```
cycraig, line 50: Document the `storage-test-suite` feature.

cycraig (commented):
Minor comments and naming gripes. Adding three dependencies isn't great, we could theoretically remove `hex` by decoding the test vector to bytes and `function_name` could be removed with some effort but probably not worth the maintenance cost of any alternative (I like the error messages).

I know my opinion is an outlier but I still prefer associated functions compared to relying on the namespace/module when exporting related functions in an API. So something similar to the `StorageTestSuite` struct in the Wasm bindings, just in Rust as well.

identity-account-storage/Cargo.toml
```diff
@@ -48,7 +51,7 @@ rusty-fork = { version = "0.3" }
 tokio = { version = "1.17.0", default-features = false, features = ["macros", "rt", "rt-multi-thread", "sync"] }
 
 [features]
-default = ["stronghold", "send-sync-storage"]
+default = ["stronghold", "send-sync-storage", "storage-test-suite"]
```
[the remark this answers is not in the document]
PhilippGackstatter: I agree. I only added this because `cargo test` without `--all-features` will fail otherwise (e.g. in CI). I assume it's fine to change CI in that case? Considering we already run `cargo check/clippy --all-features` it should be fine, even though we end up testing fewer permutations of features as a slight downside.

identity-account-storage/src/storage/memstore.rs
```diff
@@ -247,3 +247,70 @@ impl Default for MemStore {
...
+
+#[cfg(test)]
+mod tests {
+  use crate::storage::Storage;
+
+  use super::MemStore;
+
+  fn test_memstore() -> Box<dyn Storage> {
+    Box::new(MemStore::new())
+  }
```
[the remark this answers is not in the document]
PhilippGackstatter: Yes, our stronghold wrapper cannot be tested concurrently (there is an issue _somewhere_), so only with passing `--test-threads=1` to the test binary, and I don't think we can easily change that in CI for just `identity-account-storage` because we just run `cargo test` in the root, afaik. I ran the test suite locally and it passed for stronghold. Given that #787 is only waiting for some smaller changes from the stronghold team before it can be merged, that seems fine to me.

identity-account-storage/Cargo.toml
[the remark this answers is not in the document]
cycraig: Ideally we would be testing three times: with `--all-features`, with default features, and with `--no-default-features`. The only problem is that causes three compilations that cannot be cached properly.

I would be fine with just adding `--all-features` to the CI, under the`.github/workflows/build-and-test.yml` workflow.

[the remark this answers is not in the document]
olivereanderson: I would also be fine with just running `--all-features` in the CI and encourage contributors to run `cargo hack check --feature-powerset --no-dev-deps` locally (unless that turns out to take way too long) (see [cargo-hack](https://github.com/taiki-e/cargo-hack#--feature-powerset)).

identity-account-storage/src/storage/tests.rs
PhilippGackstatter: Split it in two, thanks for the suggestion!

abdulmth (approved):

bindings/wasm/src/account/storage/tests.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[wasm_bindgen]
+    impl StorageTestSuite {
+      #[wasm_bindgen(js_name = $js_name)]
+      pub fn $test_name(storage: WasmStorage) -> PromiseVoid {
+        let promise = future_to_promise(async move {
+          tests::$test_name(Box::new(storage))
+            .await
+            .map_err(|err| {
+              let mut errors: Vec<String> = err.chain().map(|error| error.to_string()).collect();
+              let output: String = AsMut::<[String]>::as_mut(&mut errors).join(": ");
```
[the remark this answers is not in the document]
PhilippGackstatter: I can only find `error_chain_fmt`, but that requires an `&impl std::error::Error` which `anyhow::Error` does not implement, and also a `Formatter`, which we don't have here, so I'll probably leave this as-is.

bindings/wasm/examples-account/src/node.ts
```diff
@@ -32,6 +34,8 @@ async function main() {
             return await unchecked();
         case "multiple_identities":
             return await multipleIdentities();
+        case "storage_test_suite":
+            return await storageTestSuite(async () => new MemStore())
```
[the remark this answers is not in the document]
PhilippGackstatter: No, you're right. I had it like that initially, but changed it so the example could be called from `stronghold-nodejs`. But since that ended up needing its own custom test calls (to exclude the `didPurge` test), combining it with `memory_storage` is better.

bindings/stronghold-nodejs/examples/src/tests/storage_test_suite.ts
```diff
@@ -2,17 +2,31 @@ import { stronghold } from '../stronghold';
...
+        await StorageTestSuite.didCreateGenerateKeyTest(await stronghold());
+    });
+    it("didCreatePrivateKey", async () => {
+        await StorageTestSuite.didCreatePrivateKeyTest(await stronghold());
+    });
+    it("keyGenerate", async () => {
         await StorageTestSuite.keyGenerateTest(await stronghold());
+    });
+    it("keyGenerate", async () => {
         await StorageTestSuite.keyDeleteTest(await stronghold());
```
cycraig, line 16: Should be `keyDelete`

```suggestion
    it("keyDelete", async () => {
        await StorageTestSuite.keyDeleteTest(await stronghold());
```

bindings/stronghold-nodejs/js/stronghold.ts
```diff
@@ -97,14 +97,14 @@ export class Stronghold implements Storage {
         return Signature.fromJSON(napiSignature.toJSON());
     }
 
-    public async chainStateGet(did: DID): Promise<ChainState | undefined | null> {
+    public async chainStateGet(did: DID): Promise<ChainState | undefined> {
         const napiDID: NapiDID = NapiDID.fromJSON(did.toJSON());
-        const napiChainState: NapiChainState | undefined | null = await this.napiStronghold.chainStateGet(napiDID);
+        const napiChainState: NapiChainState | undefined = await this.napiStronghold.chainStateGet(napiDID);
```
cycraig, line 102: I believe the `undefined | null` thing in parameters and return values is due to Napi `Options` always generating Typescript with `undefined | null`, not sure if it matters. I would say this should just match whatever the `Storage` interface from Wasm uses.

Same comment for `documentGet`.

bindings/wasm/examples-account/src/node.ts
```diff
@@ -34,8 +34,8 @@ async function main() {
             return await unchecked();
         case "multiple_identities":
             return await multipleIdentities();
-        case "storage_test_suite":
-            return await storageTestSuite(async () => new MemStore())
+        case "custom_storage":
+            return await storageTestSuite()
```
cycraig, line 38: Rename the file `memory_storage` to `custom_storage` (or whatever name is decided on) and add it to the README.md list of examples.

cycraig (approved):

Minor comments.

bindings/wasm/src/account/storage/tests.rs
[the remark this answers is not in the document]
cycraig: We need to consolidate all the errors in the Wasm bindings anyway, this can wait for that.

bindings/stronghold-nodejs/js/stronghold.ts
PhilippGackstatter: I changed it in `WasmStorage` to `ChainState | undefined` for `chainStateGet`, similarly for `documentGet`, so this should be okay as it is now.

PR #794, opened 2022-04-05
Fix Account `create_signed_*` function return types

abdulmth:
# Description of change
The return type of creating signed objects in the Account does not match the TypeScript types. 

For example 

```javascript
    // Created a signed credential by the issuer. 
    const signedVc = await issuer.createSignedCredential(
        "#newKey",
        unsignedVc,
        ProofOptions.default(),
    );
``` 
should return a `Credential` according to the method signature: 

```javascript
  createSignedCredential(fragment: string, credential: Credential, options: ProofOptions): Promise<Credential>;
```

but the actual return type is `Object` due to a mistake in using `unchecked_into` which should be tested manually every time it's used. 

This PR fixes this issues and the correct type is returned in the following fucntions:

`create_signed_credential`
`create_signed_document`
`create_signed_presentation`
`create_signed_data`

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/bindings/wasm/src/account/wasm_account/account.rs
+++ b/bindings/wasm/src/account/wasm_account/account.rs
@@ -9,8 +9,9 @@ use identity::account::Account;
 use identity::account::AccountBuilder;
 use identity::account::PublishOptions;
 use identity::account_storage::Storage;
+use identity::credential::Presentation;
 use identity::crypto::ProofOptions;
-use identity::crypto::SetSignature;
+
 use identity::did::verifiable::VerifiableProperties;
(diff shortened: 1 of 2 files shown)


cycraig (approved):
Looks good to me. We could probably use macros or generics with `create_signed_data` to deduplicate the code but this is more than fine.

I looked over other instances of `unchecked_into` in the Wasm bindings and didn't see anything wrong at a glance, so I assume these were the only functions affected by the bug.

Don't delete the branch, we can reapply it in a hotfix PR.

PR #795, opened 2022-04-05
Use Account for VC examples

abdulmth:
# Description of change
Refactor Verifiable Credentials examples to use the Account. 
The examples `create_vc`, `create_vp` and `revoke_vc` are now refactored moved from the low level examples to the account examples both in Rust and Wasm. 

The new examples have some code duplication, this was to make them self-contained in one file, since this is necessary for the documentation and the Replits. 

## Type of change

- [x] Documentation Fix

Locally ran the examples.


examples/account/create_vp.rs
```diff
@@ -81,26 +125,29 @@ async fn main() -> Result<()> {
   // Create an unsigned Presentation from the previously issued Verifiable Credential.
   let mut presentation: Presentation = PresentationBuilder::default()
     .id(Url::parse("asdf:foo:a87w3guasbdfuasbdfs")?)
```
PhilippGackstatter, line 127: @olivereanderson and I think we can remove this, because we don't have a sensible value to set. The `id` field is optional and since we don't need to refer to the presentation again, we don't need a unique identifier here.

While we're on the topic, we also agreed that it would be good to update the documentation of `PresentationBuilder::id` to `/// Sets the unique identifier of the presentation`, to be more expressive and in-line with the spec. Would you mind including that in this PR?

examples/account/create_vc.rs
```diff
@@ -0,0 +1,101 @@
...
+use identity::did::DID;
+use identity::iota::CredentialValidationOptions;
+use identity::iota::CredentialValidator;
+use identity::iota::FailFast;
+
+mod create_did;
+
+pub async fn create_vc() -> Result<String> {
+  // Create an identity for the issuer.
+  let mut issuer: Account = Account::builder().create_identity(IdentitySetup::default()).await?;
```
PhilippGackstatter, line 35: We should create one `AccountBuilder` and use it to create both `issuer` and `alice` here, unless that is intentional.

PhilippGackstatter (approved):
Looks good to me.

The only important change would be reusing the `AccountBuilder`, unless that is intentional.

examples/account/create_vp.rs
```diff
@@ -4,75 +4,119 @@
 //! A Verifiable Presentation (VP) represents a bundle of one or more Verifiable Credentials.
 //! This example demonstrates building and usage of VPs.
 //!
+//! Note: This example uses in-memory storage for simplicity.
+//! See `create_did` example to configure Stronghold storage.
+//!
```
cycraig, line 9: Move this to the `AccountBuilder` code for better visibility, if we really don't want to use Stronghold.

examples/account/revoke_vc.rs
```diff
@@ -0,0 +1,106 @@
...
+
+  // Sign the Credential with the issuer's verification method.
+  issuer.sign("#key-1", &mut credential, ProofOptions::default()).await?;
+
+  // ===========================================================================
+  // Revoke the Verifiable Credential.
+  // ===========================================================================
+
+  // Remove the public key that signed the VC from the issuer's DID document
+  // - effectively revoking the VC as it will no longer be able to verified.
```
cycraig, line 83:
```suggestion
  // Remove the public key that signed the VC from the issuer's DID document.
  // This effectively revokes the VC as it will no longer be able to be verified.
```

examples/account/revoke_vc.rs
```diff
@@ -0,0 +1,106 @@
...
+  issuer
+    .update_identity()
+    .delete_method()
+    .fragment("key-1")
+    .apply()
+    .await?;
+
+  // Check the verifiable credential
```
cycraig, line 91:
```suggestion
  // Check the verifiable credential is revoked.
```

bindings/wasm/examples/src/web.js
```diff
@@ -83,4 +63,4 @@ window.onload = async () => {
         .querySelector("#did_history_btn")
         .addEventListener("click", () => resolveHistory(CLIENT_CONFIG));
 
-};
+};
```
cycraig, line 66: Nit: add back the trailing newline at the end of files, configure the IDE to do it automatically.

bindings/wasm/examples/src/index.html
```diff
@@ -1,74 +1,62 @@
...
-        </button>
-
-        <h4 class="ma-md">Output:</h4>
-        <pre id="content" class="content"></pre>
-    </body>
-</html>
+<head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>WASM Examples Browser</title>
```
cycraig, line 7:
```suggestion
    <title>Wasm Examples Browser</title>
```

bindings/wasm/examples-account/src/create_vp.ts
```diff
@@ -12,58 +13,125 @@ import {
...
  A Verifiable Presentation is the format in which a (collection of) Verifiable Credential(s) gets shared.
  It is signed by the subject, to prove control over the Verifiable Credential with a nonce or timestamp.
-
- @param {{network: Network, explorer: ExplorerUrl}} clientConfig
  **/
-async function createVP(clientConfig) {
-    // An issuer creates and signs a Verifiable Credential and passes it to the credential subject `Alice`.
+async function createVP(storage?: Storage) {
+    // ===========================================================================
+    // Step1: Create identities for the issuer and the holder.
```
cycraig, line 29: Nit: formatting. Applies to all `StepN` comments.

```suggestion
    // Step 1: Create identities for the issuer and the holder.
```

bindings/wasm/examples-account/src/create_vp.ts
```diff
@@ -101,16 +169,19 @@ async function createVP(clientConfig) {
     const resolver = await Resolver
         .builder()
         .clientConfig({
-            network: clientConfig.network
+            network: Network.mainnet()
         })
```
cycraig, line 173: Remove the `clientConfig`, it defaults to the `mainnet` if omitted. Unless we configure the network for the accounts too.

bindings/wasm/examples-account/src/create_vp.ts
```diff
@@ -101,16 +169,19 @@ async function createVP(clientConfig) {
...
         .build();
 
     // Validate the presentation and all the credentials included in it according to the validation options
     await resolver.verifyPresentation(
         presentation,
         presentationValidationOptions,
-        FailFast.FirstError
+        FailFast.FirstError,
+        undefined,
+        undefined
```
cycraig, line 182: Why pass `undefined` here? Omitting the optional parameters is fine in Typescript.

cycraig (changes requested):
Minor changes.

examples/account/create_vp.rs
abdulmth: I think adding stronghold configuration is too much noise when the examples are already quite packed.
I added the note above the `AccountBuilder` creation.

bindings/wasm/examples-account/src/create_vp.ts
cycraig: We have a problem with async constructors in Javascript so we just use a blocking call internally (which is not ideal), the async builder is correct. I made the `Resolver` use a builder pattern due to the difficulties of passing around clients, making duck-typed parameter difficult/impossible, if I recall correctly.

bindings/wasm/examples-account/src/create_vp.ts
cycraig: Should we fix it in this PR since it's a non-breaking change?

cycraig: Fixed by #807.

Issue #534, opened 2021-12-08
[Task] Refactor error types 


olivereanderson:
## Description
Refactor the errors in this library to adhere to the following principles/goals: 

* **Unsurprising:** Our errors should have names that convey the contexts they are expected to be associated with, and vice-versa given a fallible function one should not be surprised by the name and/or properties of the corresponding error.  

* **Stable:** Our error types must avoid exposing details that are subject to change in our public API. This includes exposing underlying errors from external unstable crates either wrapped in a public enum variant directly or indirectly in implementations of the `From` trait. 

* **Simple:** Our error types should not introduce unnecessary complexity. When the user/caller is not expected to act on the error at run time, a good error message wrapped in an appropriately named type goes a long way.  

* **Flexible and ergonomic:** Our errors should be easy to construct and convert to/from.  This can be achieved by implementing the `From` trait and/or re-using already defined error types when this makes sense. 

These principles can (often) be in tension. When this is the case one is encouraged to favour those that appear first in the list.  
 
## Motivation
In [this discussion](https://github.com/iotaledger/identity.rs/discussions/460) we highlighted that our current error approach exposes  many details which both leads to problems in terms of stability as well as a high cognitive burden on the user.  After discussing ways in which we can improve our approach to error handling we decided to start refactoring our error types and function signatures. 

## To-do list
Create a task-specific to-do list . Please link PRs that match the To-do list item behind the item after it has been submitted.

- [ x ] identity-core
- [  ]  identity-account 
- [  ]  identity-credential 
- [  ]  identity-comm 
- [  ]  identity-did 
- [  ]  identity-diff 
- [  ]  identity-iota 
- [  ]  libjose 
- [  ]  bindings/wasm 

## How to contribute 
Pick a crate that has yet to have its errors refactored (see the checklist above) and submit a draft PR to [this epic branch](https://github.com/iotaledger/identity.rs/tree/epic/refactor-errors). 

### WASM/JS bindings: 
Is it possible to make the errors/error handling approach on the Javascript side more idiomatic for Node/Typescript/JS developers?

olivereanderson:
Here is an interesting question related to implementing the guidelines in the project: 

**Question:** The guidelines state the following: 
> Return source errors via Error::source unless the source error's message is included in your own error message in Display.

But are there any particular situations where one way is preferred over the other? 

**Answer:** As a rule of thumb one should prefer to return the source error via Error::source. The reason for this is that otherwise one can end up with a situation like the following:

[...]

PR #809, opened 2022-04-12
Add Diffie-Hellman key exchange to Account

### LINKED ISSUE #756: [Task] Add Diffie-Hellman key exchange to `Account`
## Description
Add support for Diffie-Hellman key exchange operations in `Account`.

## Motivation
Adding support for `X25519` keys included using them to perform Diffie-Hellman key exchange. We should expose this functionality through the `Account` too, since `Storage` prevents direct access to the private keys required to perform it. I.e. developers have no alternative if we do not support this feature directly.

The current function (not using the `Account`) looks as follows:
```rust
/// Performs a cryptographic key exchange process (e.g. Diffie-Hellman) using the private key
/// of the first party with the public key of the second party, resulting in a shared secret.
pub fn key_exchange<PRV, PUB>(private: &PRV, public: &PUB) -> Result<[u8; 32]>
where
  PRV: AsRef<[u8]> + ?Sized,
  PUB: AsRef<[u8]> + ?Sized,
{
  let private_key: x25519::SecretKey = x25519::SecretKey::try_from_slice(private.as_ref())?;
  let public_key: x25519::PublicKey = x25519::PublicKey::try_from_slice(public.as_ref())?;
  Ok(private_key.diffie_hellman(&public_key).to_bytes())
}
```

## Resources
- https://github.com/iotaledger/identity.rs/issues/671
- https://github.com/iotaledger/identity.rs/pull/735
- https://datatracker.ietf.org/doc/html/rfc7748#section-6.1
- https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange

## To-do list

- [ ] Add Diffie-Hellman key exchange function to `Account`/`Storage`.

## Change checklist

HenriqueNogara:
# Description of change
Add Diffie-Hellman key exchange to Account

closes #756 

## Type of change
- [ ] Documentation Fix

New key_exchange example.

## Change checklist


identity-account-storage/src/storage/memstore.rs
```diff
@@ -202,6 +206,85 @@ impl Storage for MemStore {
...
+    // Retrieves the PrivateKey from the vault
+    let mut vaults: RwLockWriteGuard<'_, _> = self.vaults.write()?;
+    let vault: &mut MemVault = vaults.get_mut(did).ok_or(Error::KeyVaultNotFound)?;
+    let key_pair: &KeyPair = vault.get(location).ok_or(Error::KeyNotFound)?;
+    // Encrypts the given data
+    let mut cipher_text: Vec<u8> = vec![0; data.len()];
+    let mut tag: Vec<u8> = [0; Aes256Gcm::TAG_LENGTH].to_vec();
+    Aes256Gcm::try_encrypt(
+      key_pair.private().as_ref(),
+      [0; Aes256Gcm::NONCE_LENGTH].as_ref(),
```
HenriqueNogara, line 255: What should we do here? Nonce has to be the same when encrypting and decrypting the data

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -365,6 +413,55 @@ async fn sign_ed25519(vault: &Vault<'_>, payload: Vec<u8>, location: &KeyLocatio
...
+  vault: &Vault<'_>,
+  location: &KeyLocation,
+  plaintext: Vec<u8>,
+  associated_data: Vec<u8>,
+) -> Result<EncryptedData> {
+  let aead_encrypt: procedures::AeadEncrypt = procedures::AeadEncrypt {
+    cipher: procedures::AeadCipher::Aes256Gcm,
+    associated_data,
+    plaintext,
+    nonce: [0; Aes256Gcm::NONCE_LENGTH].to_vec(),
```
HenriqueNogara, line 441: What should we do here? Nonce has to be the same when encrypting and decrypting the data

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -365,6 +407,56 @@ async fn sign_ed25519(vault: &Vault<'_>, payload: Vec<u8>, location: &KeyLocatio
...
+  associated_data: Vec<u8>,
+) -> Result<EncryptedData> {
+  let nonce: &[u8] = &Aes256Gcm::random_nonce()?;
+  let aead_encrypt: procedures::AeadEncrypt = procedures::AeadEncrypt {
+    cipher: procedures::AeadCipher::Aes256Gcm,
+    associated_data: associated_data.clone(),
+    plaintext,
+    nonce: nonce.to_vec(),
+    key: location.into(),
+  };
```
cycraig, line 438: Note that this uses the 32-byte/256-bit shared secret directly. This is basically the "Direct Key Agreement" mode referenced here: https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2

For DIDComm we need the "Key Wrapping" mode with the A256KW algorithm, as that is the only one the specification mentions for `X25519` keys: https://identity.foundation/didcomm-messaging/spec/#key-wrapping-algorithms

We either need to be more flexible and allow passing the "Key Wrapping" algorithm or mode to the encrypt/decrypt functions, or hardcode that ECDH-1PU always uses A256KW (less preferred).

identity-account/src/types/encryption_key.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use crate::Error;
+use crate::Result;
+
+/// Supported keys for encrypting data
+#[derive(Clone, Debug)]
+pub enum EncryptionKey {
+  Ed25519,
+  X25519(PublicKey),
+}
```
cycraig, line 22: I don't think `EncryptionKey` is necessary, it can usually be inferred from the verification method referenced by `fragment` and the public key can be part of a rich enum specifying with encryption algorithm to use.

cycraig (changes requested):
Preliminary review: I think we need to remove `key_exchange` and move the Diffie-Hellman operation into the encrypt/decrypt functions. We also need to be more explicit about which encryption algorithm is to be used in both `Account` and `Storage`. Lastly, we need to support the A256KW algorithm for ECDH-1PU required by DIDComm.

See:
- https://identity.foundation/didcomm-messaging/spec/#curves-and-content-encryption-algorithms
- https://identity.foundation/didcomm-messaging/spec/#key-wrapping-algorithms
- https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-04#section-2

identity-account-storage/src/storage/memstore.rs
cycraig: I think these comments can be resolved. For the reader: `crypto.rs` is used to generate an appropriate nonce and that nonce is returned in `EncryptedData`.

identity-account-storage/src/types/encrypted_data.rs
```diff
@@ -0,0 +1,41 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+
+/// The structure returned after encrypting data
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct EncryptedData {
+  nonce: Vec<u8>,
+  associated_data: Vec<u8>,
+  tag: Vec<u8>,
+  cypher_text: Vec<u8>,
```
cycraig, line 13: Nit: should be one word and I think `ciphertext` might be more common? One word either way.

```suggestion
  ciphertext: Vec<u8>,
```

identity-account-storage/src/storage/stronghold.rs
cycraig: This is implemented in the `encrypt_cek` function in `libjose`: https://github.com/iotaledger/identity.rs/blob/6c7d5a6b8d9f507839b9c1e92c3efb2609a72819/libjose/src/jwe/encoder.rs#L355

It uses the `aes-kw` feature of `crypto.rs`.

examples/account/key_exchange.rs
```diff
@@ -0,0 +1,109 @@
...
+    .content(MethodContent::GenerateX25519)
+    .fragment("kex-0")
+    .scope(MethodScope::key_agreement())
+    .apply()
+    .await?;
+
+  // Alice and Bob tell each other their DIDs. They each resolve the DID Document of the other
+  // to obtain their X25519 public key. Note that in practice, they would run this code completely
+  // separately.
+  let client: Client = Client::new().await?;
```
olivereanderson, line 65: I am wondering whether the `Resolver` is more fitting than the `Client` here as the only Client related operations performed is DID Document resolution. Maybe @cycraig has an opinion on this matter?

identity-account-storage/src/error.rs
```diff
@@ -12,6 +12,9 @@ pub enum Error {
   /// Caused by errors from the [identity_core] crate.
   #[error(transparent)]
   CoreError(#[from] identity_core::Error),
+  /// Caused by errors from the [crypto] crate.
+  #[error(transparent)]
+  CryptoError(#[from] crypto::error::Error),
```
olivereanderson, line 17: I might be wrong, but I was not able to find any reference to this variant in my IDE. If it is not needed then it can be removed. Regardless of whether it is actually in use or not I have a few concerns about the value of this variant.  It does not describe the context in which this error from an external crate occurred. Also not on the documentation of the variant hence it becomes very hard for those that want to handle errors of this type to understand when it may occur unless they are very familiar with the internals of this crate. Furthermore this is also a variant of `identity_core::Error` which is wrapped in the `CoreError` variant of this error enum. By exposing the error from the crypto crate we are assuming some consumers of identity.rs may want to handle errors of this type, but when it can appear without context in multiple places this task becomes increasingly hard.

identity-account-storage/src/error.rs
```diff
@@ -29,6 +32,13 @@ pub enum Error {
   /// [`KeyType`][identity_core::crypto::KeyType].
   #[error("Invalid Private Key: {0}")]
   InvalidPrivateKey(String),
+  /// Caused by attempting to perform Diffie-Hellman key exchange with the wrong key type
+  #[error("Invalid Key Type: {0}")]
+  InvalidKeyType(String),
+  /// Caused by providing bytes that cannot be used as a public key of the
+  /// [`KeyType`][identity_core::crypto::KeyType].
+  #[error("Invalid Public Key: {0}")]
+  InvalidPublicKey(String),
```
olivereanderson, line 41: I think this crate could present cryptography related errors in a better way. I tried removing the `CoreError` and `CryptoError` variants and pretty much all compilation errors were then the result of missing error types related to results of cryptographic operations. Hence I would suggest removing the `CoreError` and `CryptoError` variants and do one of the following things: 
**Option A:** Make an additional error in this crate `CryptographyError` (and also wrap it in the top level error enum) which you map cryptography related errors from `identity_core` and `crypto.rs` to, or 
**Option B:**  Keep the `InvalidKeyType` and `InvalidPublicKey` variants and add a few more variants to cover what was needed from  `CoreError` and `CryptoError`
**Option C:** Make a new error in the `identity_core::crypto` module covering various failures of cryptographic operations and introduce a new variant here that wraps said error. 

See https://github.com/iotaledger/identity.rs/issues/534 for more information on our error guidelines. I think this is a good opportunity to get better alignment with what our error guidelines state.

olivereanderson: Let me know if following up on my suggestion would be too much extra work for this PR. If that is the case I would be happy to help in a separate branch.

identity-account-storage/src/storage/stronghold.rs
cycraig: Opened https://github.com/iotaledger/stronghold.rs/issues/338 to request A256KW support in Stronghold.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -273,6 +280,70 @@ impl Storage for Stronghold {
...
+          public_key
+            .ok_or_else(|| Error::InvalidPublicKey("missing second party public key".to_owned()))?
+            .as_ref()
+            .try_into()
+            .map_err(|_| Error::InvalidPublicKey(format!("expected type: [u8, {}]", X25519::PUBLIC_KEY_LENGTH)))?,
+          &shared_secret,
+        )
+        .await?;
+        let encrypted_data: Result<EncryptedData> =
+          aead_encrypt(&vault, algorithm, &shared_secret, data, associated_data).await;
```
cycraig, line 309: I'm unsure how common or safe using the shared secret directly as the encryption key is. Most resources specify using a key-derivation function on the shared secret prior to using it for encryption. However, https://www.rfc-editor.org/rfc/rfc7518#section-4.5 does refer to using the shared secret directly, with security considerations regarding key lifetimes.

Basically, developers should be able to specify the key-wrapping algorithm used too.

identity-account-storage/src/types/encryption_algorithm.rs
```diff
@@ -0,0 +1,11 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Deserialize;
+use serde::Serialize;
+
+#[non_exhaustive]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum EncryptionAlgorithm {
+  Aes256Gcm,
+}
```
cycraig, line 11: Specifying `Aes256Gcm` alone might be insufficient. There are basically three configurable parts to encryption:

- Key exchange algorithm (e.g. RSA, ECDH), which is implicit by the type of key used.
- Encryption algorithm (e.g. AES256-GCM, AES256-CBC + HMAC-SHA512).
- Key wrapping algorithm (e.g. A256KW).

Unfortunately it's not so clear-cut. We have to deal with algorithms like ECDH-1PU which has additional steps (performs two Diffie-Hellman key exchanges using the ephemeral and static keys) and inputs.

We might want to follow the approach taken in `libjose`, where the entire algorithm is specified as an enum, e.g. `ECDH_1PU_A256KW`, except with a rich enum for additional parameters required.

https://github.com/iotaledger/identity.rs/blob/4287c9834de8dc42fd4ff0c0e8b3a2bc822d0ec4/libjose/src/jwe/algorithm.rs#L95

Note that this is separate from the encryption algorithm, which could be specified separately?
https://github.com/iotaledger/identity.rs/blob/4287c9834de8dc42fd4ff0c0e8b3a2bc822d0ec4/libjose/src/jwe/encryption.rs#L39

Basically, all this is saying is we need more flexibility/parameters/enums than just `EncryptionAlgorithm` to determine encryption behaviour.

cycraig (commented):
I'm not sure how useful this is until we get support for key wrapping and key concatenation in Stronghold:

- https://github.com/iotaledger/stronghold.rs/issues/338
- https://github.com/iotaledger/stronghold.rs/issues/339

We can only really support `ECDH-ES` with no key wrapping right now.

identity-account-storage/src/types/encryption_algorithm.rs
PhilippGackstatter: I made a similar addition to `identity-comm` for specifying the algorithm to determine the content encryption key (CEK) (called `CEKAlgorithm`) and that in addition to the `EncryptionAlgorithm` seems to work well so far, but I can't say yet whether that is the universally the best design. It seems like a good start though, here it is for reference:

https://github.com/iotaledger/identity.rs/blob/dbdda47160e8725424a31dfd0ff4a274f323b3f6/identity-comm/src/envelope/encrypted.rs#L131-L167

identity-account-storage/src/storage/stronghold.rs
olivereanderson: [Section 2.1 of the AEAD Interface spec](https://datatracker.ietf.org/doc/html/rfc5116#section-2.1) states that the encryption key:
> MUST be generated in a way that is uniformly random or pseudorandom.

Which is NOT the case for the shared secret produced by ECDH (as stated in for instance  [Section 3.3 of RFC 5869](https://datatracker.ietf.org/doc/html/rfc5869#section-3.3) and elsewhere). Hence we should probably indeed use a KDF (key-derivation function) on the shared secret. 

We could consider adding an additional optional parameter to this method specifying the KDF to use (possibly by introducing an enum of acceptable key-derivation functions (maybe with the possibility for the user to add a salt?)). Alternatively if there is some sufficiently common and recommended standard KDF to be used with ECDH we could decide to only support that.

cycraig: Thanks for the confirmation and sources.

To elaborate from a Slack conversation, [RFC 7518 Section 4.6](https://datatracker.ietf.org/doc/html/rfc7518#section-4.6), which defines `ECDH-ES`, specifies that the "Concat KDF" (which I believe uses a SHA digest per NIST recommendations) must be used to produce the content encryption key in both "Direct Key Agreement Mode" and "Key Wrapping Mode" (even though the wording of "key agreement result" is a bit misleading in my opinion):

> This section defines the specifics of key agreement with Elliptic
 Curve Diffie-Hellman Ephemeral Static [[RFC6090](https://datatracker.ietf.org/doc/html/rfc6090)], in combination with
  the **Concat KDF**, as defined in Section 5.8.1 of [[NIST.800-56A](https://datatracker.ietf.org/doc/html/rfc7518#ref-NIST.800-56A)].
>  The key agreement result can be used in one of two ways:
> 
> 1.  directly as the Content Encryption Key (CEK) for the "enc"
>    algorithm, in the Direct Key Agreement mode, or
> 
> 2.  as a symmetric key used to wrap the CEK with the "A128KW",
>    "A192KW", or "A256KW" algorithms, in the Key Agreement with Key
>    Wrapping mode.
> 
> A new ephemeral public key value MUST be generated for each key
> agreement operation.
> 
> **In Direct Key Agreement mode, the output of the Concat KDF MUST be a**
> **key of the same length as that used by the "enc" algorithm.** [...]

So basically without "Concat KDF" we would not technically have a spec-compliant ECDH-ES algorithm, which is a problem for interoperability if other libraries cannot decrypt it. (Aside: we definitely need to use their test vectors here.)

So we're stuck until https://github.com/iotaledger/stronghold.rs/issues/339 is resolved.

PhilippGackstatter: Opened iotaledger/stronghold#344 to request the concat KDF.

PhilippGackstatter (commented):
Preliminary comments.

bindings/wasm/src/account/types/cek_algorithm.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Supported CEK algorithms
+#[derive(Clone, Serialize, Deserialize)]
+#[wasm_bindgen(js_name = CEKAlgorithm, inspectable)]
+pub struct WasmCEKAlgorithm(CEKAlgorithm);
+
+#[wasm_bindgen(js_class = CEKAlgorithm)]
+impl WasmCEKAlgorithm {
+  /// ECDH-ES will be used as the content encryption key.
+  #[wasm_bindgen(js_name = ecdhES)]
+  pub fn ecdh_es(agreement: &WasmAgreementInfo) -> WasmCEKAlgorithm {
```
cycraig, line 22: Nit: would prefer `ECDH_ES` casing for consistency.

```suggestion
  #[wasm_bindgen(js_name = ECDH_ES)]
  pub fn ecdh_es(agreement: &WasmAgreementInfo) -> WasmCEKAlgorithm {
```

bindings/wasm/src/account/types/encryption_algorithm.rs
```diff
@@ -0,0 +1,48 @@
...
+/// Supported keys for encrypting data
+#[derive(Clone, Serialize, Deserialize)]
+#[wasm_bindgen(js_name = EncryptionAlgorithm, inspectable)]
+pub struct WasmEncryptionAlgorithm(EncryptionAlgorithm);
+
+#[wasm_bindgen(js_class = EncryptionAlgorithm)]
+impl WasmEncryptionAlgorithm {
+  /// Encrypts/Decrypts data using Aes256Gcm.
+  #[wasm_bindgen(js_name = aes256gcm)]
+  pub fn aes256gcm() -> WasmEncryptionAlgorithm {
```
cycraig, line 21: Q: do we intend to follow `JWA` naming schemes, should we use `Aes256Gcm` or `A256GCM` as per JWA? If we keep the former then we should probably use `EcdhEs`? Doesn't really matter but we should just be consistent.

bindings/wasm/src/account/types/encryption_options.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Object used for defining the algorithms used for encrypting/decrypting data
+#[derive(Clone, Serialize, Deserialize)]
+#[wasm_bindgen(js_name = EncryptionOptions, inspectable)]
+pub struct WasmEncryptionOptions(EncryptionOptions);
+
+#[wasm_bindgen(js_class = EncryptionOptions)]
+impl WasmEncryptionOptions {
+  /// Creates an `EncryptionOptions` object.
+  #[wasm_bindgen(js_name = new)]
+  pub fn new(
```
cycraig, line 23: Should use `constructor`.

```suggestion
  /// Creates an `EncryptionOptions` object.
  #[wasm_bindgen(constructor)]
  pub fn new(
```

identity-account-storage/src/storage/memstore.rs
```diff
@@ -245,6 +255,90 @@ impl Storage for MemStore {
...
+    public_key: PublicKey,
+  ) -> Result<EncryptedData> {
+    // Retrieves the PrivateKey from the vault
+    let vaults: RwLockReadGuard<'_, _> = self.vaults.read()?;
+    let vault: &MemVault = vaults.get(did).ok_or(Error::KeyVaultNotFound)?;
+    let key_pair: &KeyPair = vault.get(private_key).ok_or(Error::KeyNotFound)?;
+    // Encrypts the data
+    match key_pair.type_() {
+      KeyType::Ed25519 => Err(Error::InvalidPrivateKey(
+        "ED25519 keys are not suported for encrypting/decrypting data".to_owned(),
```
cycraig, line 274: @PhilippGackstatter flagged these error messages elsewhere so I'll highlight these for consistency:

```suggestion
        "Ed25519 keys are not supported for encryption".to_owned(),
```

identity-account-storage/src/storage/memstore.rs
```diff
@@ -245,6 +255,90 @@ impl Storage for MemStore {
...
+    public_key: PublicKey,
+  ) -> Result<Vec<u8>> {
+    // Retrieves the PrivateKey from the vault
+    let vaults: RwLockReadGuard<'_, _> = self.vaults.read()?;
+    let vault: &MemVault = vaults.get(did).ok_or(Error::KeyVaultNotFound)?;
+    let key_pair: &KeyPair = vault.get(private_key).ok_or(Error::KeyNotFound)?;
+    // Decrypts the data
+    match key_pair.type_() {
+      KeyType::Ed25519 => Err(Error::InvalidPrivateKey(
+        "ED25519 keys are not suported for encrypting/decrypting data".to_owned(),
```
cycraig, line 318:
```suggestion
        "Ed25519 keys are not supported for decryption".to_owned(),
```

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -270,6 +280,98 @@ impl Storage for Stronghold {
...
+            .await?;
+            let encrypted_data: Result<EncryptedData> = aead_encrypt(
+              &client,
+              &encryption_options.encryption_algorithm(),
+              concat_kdf.clone(),
+              data,
+              associated_data,
+            )
+            .await;
+            revoke_data(&client, shared_key, true).await?;
```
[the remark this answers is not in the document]
cycraig: Kind of ties into the fact that we lack transaction-like behaviour: https://github.com/iotaledger/identity.rs/issues/836

I rate we leave it for now, it's a pervasive issue.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -270,6 +280,98 @@ impl Storage for Stronghold {
...
+    did: &IotaDID,
+    data: EncryptedData,
+    encryption_options: &EncryptionOptions,
+    private_key: &KeyLocation,
+    public_key: PublicKey,
+  ) -> Result<Vec<u8>> {
+    let client: Client = self.client(&ClientPath::from(did))?;
+    match private_key.key_type {
+      KeyType::Ed25519 => Err(Error::InvalidPrivateKey(
+        "ED25519 keys are not suported for encrypting/decrypting data".to_owned(),
```
cycraig, line 341:
```suggestion
        "Ed25519 keys are not suported for decryption".to_owned(),
```

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -270,6 +280,98 @@ impl Storage for Stronghold {
...
+      )),
+      KeyType::X25519 => {
+        let public_key: [u8; X25519::PUBLIC_KEY_LENGTH] = public_key
+          .as_ref()
+          .try_into()
+          .map_err(|_| Error::InvalidPublicKey(format!("expected type: [u8, {}]", X25519::PUBLIC_KEY_LENGTH)))?;
+        match encryption_options.cek_algorithm() {
+          CEKAlgorithm::ECDH_ES { agreement } => {
+            let shared_key: Location = diffie_hellman(&client, private_key, public_key).await?;
+            let concat_kdf: Location = concat_kdf(
```
cycraig, line 351: @PhilippGackstatter are there chained procedures in Stronghold yet that we can use to avoid temporary locations?

PhilippGackstatter:
As @cycraig pointed out, `ECDH_ES` needs to return the ephemeral public key (epk) it generates, because the caller needs access to it. This is also true for `ECDH_ES+*`. However, e.g. `ECDH-ES+A256KW`, will also encrypt the cek internally and also needs to return that for the caller's use. (Both epk and encrypted cek need to be part of the JWE). The issue is, that with the current API we can't express that the key wrap variant will return an encrypted cek, while the non key wrap variant will not. Similarly, when decrypting, that epk (and the encrypted cek in key wrap variants) needs to be passed.

I'd probably just add a `encrypted_cek: Vec<u8>` in `EncryptedData` and leave it empty for `ECDH_ES`, (which is how JWA defines the `encrypted_key` value on the message as well [for reference](https://datatracker.ietf.org/doc/html/rfc7518#section-4.6)), which means it can also be omitted entirely in this PR, and we could add it when we add the key wrap variants. However, this makes the `Storage::encrypt_data` method quite specific to `ECDH_ES*` and perhaps we should consider renaming it to something more specific like `aead_encrypt` or something, so we can more easily add more encryption methods later?

There are still one (and a half) issues for use of this method in DIDComm.

## CEK generation

When generating a JWE for multiple recipients, the CEK is the same for the entire message. The current `encrypt_data` assume that there is a single recipient `public_key` parameter. Somehow it needs to be possible to generate a single CEK for multiple recipients.

## Ephemeral Key Pair

The way I interpret the examples in https://identity.foundation/didcomm-messaging/spec/#c3-didcomm-encrypted-messages, is that _ideally_ there is only a single `epk` for the entire message in the global protected header. Each individual recipient header only contains the `kid` which the recipient can use, together with the `epk` from the global header, to decrypt the `encrypted_key`. This means that when executing `ECHD_ES_A256KW`, we would want to generate a single ephemeral key pair for all recipients, but do the (DH + Concat KDF + CEK key wrapping) steps for each recipient.

I'm still in the process of understanding JOSE and I think that this is not a requirement, but just an optimization. It would also be possible to generate an `epk` for each individual recipient and include it in the per-recipient header (which is actually what `libjose` does). From the [message decryption procedure in JWE RFC](https://www.rfc-editor.org/rfc/rfc7516.html#section-5.2)](https://www.rfc-editor.org/rfc/rfc7516.html#section-5.2):

> ... let the JOSE Header be the union of the members of the JWE Protected Header, the JWE Shared Unprotected Header and the corresponding JWE Per-Recipient Unprotected Header, all of which must be completely valid JSON objects. During this step, verify that the resulting JOSE Header does not contain duplicate Header Parameter names.  When using the JWE JSON Serialization, this restriction includes that the same Header Parameter name also MUST NOT occur in distinct JSON object values that together comprise the JOSE Header.

As long as there is just a single `epk` present, it shouldn't matter where it is defined, as these are all merged together from how I understand it.

## Changes?

At least the CEK generation needs to be addressed. We can change the current `public_key` parameter to `Vec<PublicKey>` and make the whole function be aware, that it encrypts data for 1+ recipients. In my opinion, this is cramming too much into a single method. Alternatively, we redesign the entire API (a lot), which would also allow for the ephemeral key pair optimization (although if truly optional, I would consider that a bonus). @cycraig already mentioned:

> It is unclear how we should handle this from an API perspective, perhaps we should expose ephemeral key operations via the `Account`, and therefore `Storage` too? 

Building on that, I think one way to implement this would be by splitting `Storage::encrypt_data` into smaller methods (signatures are rough and in pseudo Rust - just tries to convey the idea):

- `generate_random_key(length) -> (Key)Location`
  - generates a random CEK suitable for a given cipher, return the location
- a method that can generate an ephemeral key pair
  - can use `Storage::key_generate(..., KeyType::X25519, ...)` as-is for that
- `diffie_hellman(secret_key_location, public_key_bytes, Some(agreement_info)) -> shared_key_location`
  - Diffie-Hellman with an optional Concat KDF step
- `aes_key_wrap(encryption_key_location, key_to_wrap_location) -> encrypted_key_bytes`
- `aead_encrypt(cipher, content_encryption_key_location, plaintext) -> ciphertext`
  - operation that is passed the location of the CEK (either one generated randomly (`ECDH_ES+*`), or one produced by DH + concat KDF (`ECDH_ES`)).

The `Account` can then glue together these functions to offer a single method that does what `Storage::encrypt_data` does now for convenience. But the actor / DIDComm implementations can directly call these operations on `Storage` and, in particular, generate a CEK once, and call DH + Concat KDF + Key Wrap multiple times to address both of the mentioned issues.

The major downside is that this adds significant complexity to the `Storage` API, although not to the internals, as `encrypt_data` already adds that complexity to all `Storage` implementations. This might be the point to explicitly allow `Storage` implementations to not support certain methods.

We likely have to do it this way, if we want to do it at all, because using generics will not translate well to Wasm or other bindings. We can also further consider a procedure-like API for this, which pollutes the interface itself less. Type-safely defining the return types continues to be an open problem without using generics, though.

Finally, sorry for bringing this up only now, but I only recently understood in more detail how `ECDH_ES` with key wrapping works.

identity-account-storage/src/storage/stronghold.rs
PhilippGackstatter: Looks like stronghold cannot do that, after all. I thought I remembered it did that once, but the procedures API has gone through more iterations since then, and this seems to have been dropped, or I misremembered completely. We will need to do the revocation ourselves, unfortunately.

identity-account/src/account/account.rs
```diff
@@ -303,6 +306,58 @@ where
     Ok(())
   }
 
+  /// Encrypts the given `data` with the specified `algorithm`
+  ///
+  /// Diffie-Helman key exchange will be performed in case an [`KeyType::X25519`] is given.
+  pub async fn encrypt_data(
```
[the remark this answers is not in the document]
PhilippGackstatter: That's a good idea! I suppose the issue is that we cannot feature gate it in `WasmStorage`, so I think you mean we will just unconditionally include it in `WasmStorage`.

Alternatively or specifically for bindings, we can explicitly allow implementations to return a `NotSupported` error or something, but the compile-time check is obviously preferred. It would be easier if we had a dedicated `StorageError` abstraction, too.

examples/account/key_exchange.rs
```diff
@@ -0,0 +1,119 @@
...
+  let message: &[u8] = b"This msg will be encrypted and decrypted";
+  let encrypted_data: EncryptedData = alice_account
+    .encrypt_data(
+      message,
+      b"associated_data",
+      &encryption_options,
+      "kex-0",
+      bob_public_key.into(),
+    )
+    .await?;
```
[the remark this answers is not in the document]
PhilippGackstatter: Just to confirm: I can't point to a specific RFC section, but I'm pretty sure the ephemeral key pair can be discarded afterwards. The sender has no use for the `ephemeral_secret` after the message was sent.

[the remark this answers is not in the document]
olivereanderson: > I'm actually uncertain whether Alice needs to retain said ephemeral X25519 keypair or if it is discarded and a new one generated for each anonymous-encrypted message they send? 

[Section 5.6.3.3 of NIST Special Publication 800-56A Revision 2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf) states the following: 

> An ephemeral private key shall be used in exactly one key-establishment transaction,
with one exception: an ephemeral private key may be used in multiple DLC key-transport
transactions that are transporting identical secret keying material simultaneously (or
within a short period of time; see the broadcast scenario in Section 7). **After its use, an
ephemeral private key shall be destroyed.**

However note that the NIST recommendation also states: 

> If a recipient of an ephemeral public key requires assurance that the claimed owner of
that public key has possession of the corresponding private key, then, in order to obtain
that assurance in compliance with this Recommendation, such assurance shall be
obtained as specified in Section 5.6.2.2.4. 

In the case of DID Comm I don't think it makes sense for Alice to retain said ephemeral key as those messages are assumed to be asynchronous, but for an ordinary key exchange it could be nice for the Account to provide a way for Alice to hold on to the ephemeral key in case Bob asks Alice to prove ownership of the corresponding private key.

identity-account-storage/src/storage/traits.rs
```diff
@@ -118,6 +120,31 @@ pub trait Storage: storage_sub_trait::StorageSendSyncMaybe + Debug {
   /// Returns `true` if a key exists at the specified `location`.
   async fn key_exists(&self, did: &IotaDID, location: &KeyLocation) -> Result<bool>;
 
+  /// Encrypts the given `data` with the specified `algorithm`
+  ///
+  /// Diffie-Helman key exchange will be performed in case an [`KeyType::X25519`] is given.
+  async fn encrypt_data(
```
olivereanderson, line 126: Please expand on the documentation here. Give some examples (in words) what the parameters `did` and `associated_data` might be used for.

identity-account-storage/src/storage/traits.rs
```diff
@@ -118,6 +120,31 @@ pub trait Storage: storage_sub_trait::StorageSendSyncMaybe + Debug {
...
+    associated_data: Vec<u8>,
+    algorithm: &EncryptionOptions,
+    private_key: &KeyLocation,
+    public_key: PublicKey,
+  ) -> Result<EncryptedData>;
+
+  /// Decrypts the given `data` with the specified `algorithm`
+  ///
+  /// Diffie-Helman key exchange will be performed in case an [`KeyType::X25519`] is given.
+  async fn decrypt_data(
```
olivereanderson, line 139: Again please write a bit more documentation here. What could the different parameters indicate/ be used for.

identity-account-storage/src/types/encrypted_data.rs
```diff
@@ -0,0 +1,41 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Deserialize;
+use serde::Serialize;
+
+/// The structure returned after encrypting data
```
olivereanderson, line 7: Perhaps also link to the method that returns this?

identity-account-storage/src/types/encrypted_data.rs
```diff
@@ -0,0 +1,41 @@
...
+
+  pub fn ciphertext(&self) -> &[u8] {
+    &self.ciphertext
+  }
+
+  pub fn tag(&self) -> &[u8] {
+    &self.tag
+  }
+
+  pub fn nonce(&self) -> &[u8] {
```
olivereanderson, line 38: Would be nice with some more documentation here. Since just slices of bytes are returned it would be nice if the documentation explained (or links to something that explains) what `associated_data`, `tag`, `ciphertext` and `nonce` is/could be.

identity-account/src/account/account.rs
olivereanderson, line 312: Again please explain the various parameters in the documentation.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -270,6 +280,98 @@ impl Storage for Stronghold {
       .map_err(Into::into)
   }
 
+  async fn encrypt_data(
```
olivereanderson, line 283: Please document this. Explain what `did` is used for in this case etc.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -394,6 +496,119 @@ fn sign_ed25519(client: &Client, payload: Vec<u8>, location: &KeyLocation) -> Re
...
+}
+
+async fn revoke_data<T: Into<Location>>(client: &Client, location: T, should_gc: bool) -> Result<()> {
+  let revoke_data: procedures::RevokeData = procedures::RevokeData {
+    location: location.into(),
+    should_gc,
+  };
+  client
+    .execute_procedure(revoke_data)
+    .map_err(|err| StrongholdError::Procedure(std::any::type_name::<procedures::RevokeData>(), vec![], err))?;
```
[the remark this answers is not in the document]
PhilippGackstatter: Yes, you can change it to `String`.

I think it's worth adding another location type then, next to `KeyLocation`. Maybe `TempLocation` that just wraps a `String` and has a method to produce a new random instance of itself, and also add a `From<&TempLocation> for Location` impl. Both location types should also be convertible to `String` so we can add it to the error message when needed.
Most functions that take a `KeyLocation` or `Location` should then be convertible to take a `&T` where `T: Into<Location>`.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -270,6 +280,98 @@ impl Storage for Stronghold {
...
+  async fn encrypt_data(
+    &self,
+    did: &IotaDID,
+    data: Vec<u8>,
+    associated_data: Vec<u8>,
+    encryption_options: &EncryptionOptions,
+    private_key: &KeyLocation,
+    public_key: PublicKey,
+  ) -> Result<EncryptedData> {
+    let client: Client = self.client(&ClientPath::from(did))?;
```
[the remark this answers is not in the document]
PhilippGackstatter: On second thoughts, we don't actually want to commit any modifications here, and without committing the changes explicitly, they are not written to the snapshot state. This way we don't even have to revoke the data, it's just dropped. Ideally, add a comment when loading the client and running mutating operations without committing them, that this is intentional.

identity-account-storage/src/storage/stronghold.rs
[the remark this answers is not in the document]
PhilippGackstatter: See https://github.com/iotaledger/identity.rs/pull/809/files#r872950555, which might be the way to work around this entirely.

.github/workflows/shared-build-and-test-stronghold-nodejs.yml
```diff
@@ -63,17 +63,18 @@ jobs:
...
+            #   target: x86_64-unknown-linux-musl
+            #   architecture: x64
+            #   setup: |
+            #    sudo apt-get install -y musl-tools
+            #   build: |
+            #    export NPM_BUILD_NAPI_ARGS=--target=x86_64-unknown-linux-musl
+            #    npm run build && strip dist/*.node
+            #   test-docker: |
+            #    docker build -f .github/Dockerfile.alpine -t tester .github
+            #    docker run --rm -v $(pwd):/build -w /build/bindings/stronghold-nodejs tester sh -c "ldd dist/identity-stronghold-nodejs.linux-x64-musl.node || true && npm run test:examples"
```
cycraig, line 77: Musl builds for Stronghold bindings were fixed in #845, pull from dev.

cycraig:
https://github.com/iotaledger/identity.rs/pull/809#issuecomment-1124704042
> The Account can then glue together these functions to offer a single method that does what Storage::encrypt_data does now for convenience. But the actor / DIDComm implementations can directly call these operations on Storage and, in particular, generate a CEK once, and call DH + Concat KDF + Key Wrap multiple times to address both of the mentioned issues.
>
> The major downside is that this adds significant complexity to the Storage API, although not to the internals, as encrypt_data already adds that complexity to all Storage implementations. This might be the point to explicitly allow Storage implementations to not support certain methods.

Ideally the `Account` should aim to provide all features required by the actor / DIDComm rather than going through `Storage`. As mentioned, decomposing `encrypt_data` into individual functions adds complexity to `Storage` implementers but I argue it does significantly affect their internals. For instance with the suggested decomposition a `Storage` has to manage temporary/ephemeral key locations, which introduces temporary storage management, garbage collection concerns etc. even though those only once (the use case given of multiple recipients can also be addressed by simply allowing multiple public keys as input). It would also mean implementers would not be able to call existing end-to-end implementations of such algorithms like `ECDH-ES`.

Another avenue would be to embrace JWA fully and use that for the interface (its well-specified headers, options etc.), however we would likely lose some type-safety in the process.

> We likely have to do it this way, if we want to do it at all, because using generics will not translate well to Wasm or other bindings. We can also further consider a procedure-like API for this, which pollutes the interface itself less. Type-safely defining the return types continues to be an open problem without using generics, though.

A procedure API would likely be appropriate but I agree the bindings are severely limiting. We could explore more creative uses of Typescript with wasm-bindgen for generics, however.

https://github.com/iotaledger/identity.rs/pull/809#discussion_r872483228
> [...] **After its use, an ephemeral private key shall be destroyed.**
> [...]
> In the case of DID Comm I don't think it makes sense for Alice to retain said ephemeral key as those messages are assumed to be asynchronous, but for an ordinary key exchange it could be nice for the Account to provide a way for Alice to hold on to the ephemeral key in case Bob asks Alice to prove ownership of the corresponding private key.

Thank you for the link. I would consider retaining the generated ephemeral key to be an atypical use-case then. Users can always generate the ephemeral key themselves, and we could perhaps support inputting a pre-generated keypair then. That would also sufficiently address the use-case of using a single `epk` with multiple recipients.

Overall the `Account` should focus on providing necessary cryptographic functionality for DID/DIDComm operations, not scope-creep to become a generic key management system.

olivereanderson:
> Users can always generate the ephemeral key themselves, and we could perhaps support inputting a pre-generated keypair then. That would also sufficiently address the use-case of using a single epk with multiple recipients.

Supporting a pre-generated keypair (as an `Option`) sounds like a really sensible thing to do!

bindings/stronghold-nodejs/examples/src/tests/multiple_identities.ts
```diff
@@ -4,6 +4,7 @@ import { stronghold } from '../stronghold';
 // Only verifies that no uncaught exceptions are thrown, including syntax errors etc.
 describe("Test Stronghold Node.js examples", function () {
     it("Multiple Identities", async () => {
-        await multipleIdentities(await stronghold());
+        // TODO: Temporarily disabled until iotaledger/stronghold.rs#353 is fixed.
```
eike-hass, line 7: should be okay again after a merging `dev`

cycraig: To clarify, it's disabled on `dev` too.

cycraig: Can we use an async `RwLock` over the Stronghold storage implementation to prevent this bug in the meantime then?

PhilippGackstatter: I guess we can, but this issue is supposed to be prioritized now, so perhaps @felsweg-iota can give an update on the current status.

felsweg-iota: The current problem with the implementation of the standard interface to Stronghold is, that it internally uses a std::sync::RwLock, since every function in the interface isn't supposed to be `async`. This actually brings in a problem while working with `async` code. 

There are two solutions to this, and i am just naming them here for reference, @PhilippGackstatter already pointed out the link to the issue: 

- we either change the current standard library locking mechanism to their `async` counterparts, which will effectively change the interface to operate as pseudo-`async`, 
- the STM based approach will be finalized and will replace all existing locks.

PhilippGackstatter: I think I'd prefer if we try the `RwLock` solution in our code to fix it temporarily (if that works), and that could avoid you having to do temporary work if the STM will eventually be the proper solution.

identity-account-storage/src/storage/traits.rs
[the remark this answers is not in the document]
HenriqueNogara: Since this sets the values of the ephemeral keys, we can't add it to the StorageTestSuite. I added a test for the stronghold type, but I feel it doesn't add much more than what we current assert in the `encryption` example.

identity-account/src/account/account.rs
```diff
@@ -307,6 +313,59 @@ where
...
+        encryption_options,
+        public_key,
+      )
+      .await
+      .map_err(Into::into)
+  }
+
+  /// Decrypts the given `data` with the specified `encryption_options`.
+  ///
+  /// Diffie-Helman key exchange with Concatenation Key Derivation Function will be performed to obtain the decryption
```
PhilippGackstatter, line 345:
```suggestion
  /// Diffie-Hellman key exchange with Concatenation Key Derivation Function will be performed to obtain the decryption
```

Probably best to do a search-replace for "Helman", this affects Wasm and NAPI as well.

identity-account/src/account/account.rs
```diff
@@ -307,6 +313,59 @@ where
...
+  /// secret.
+  /// 
+  /// Returns the decrypted text.
+  #[cfg(feature = "encryption-decryption")]
+  pub async fn decrypt_data(
+    &self,
+    data: EncryptedData,
+    encryption_options: &EncryptionOptions,
+    fragment: &str,
```
PhilippGackstatter, line 354: This method's docs should explain what the `fragment` is used for.

identity/Cargo.toml
```diff
@@ -50,6 +50,9 @@ send-sync-storage = ["identity-account-storage/send-sync-storage"]
 # Exposes Storage `test_suite` module.
 storage-test-suite = ["identity-account-storage/storage-test-suite"]
 
+# Enables encryption and decryption functionality
+encryption-decryption = ["identity-account/encryption-decryption"]
```
PhilippGackstatter, line 54: Change the name here as well, if you decide to change it for the `account` crate. Maybe it makes sense to use `account-encryption` for clarity? Not sure, I leave the decision to you.

examples/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 publish = false
 
 [dependencies]
-identity = { path = "../identity", features = ["account", "stronghold"] }
+identity = { path = "../identity", features = ["account", "stronghold", "encryption-decryption"] }
```
PhilippGackstatter, line 8: This seems unnecessary, because the encryption API is enabled by default in `identity-account`.

identity-account-storage/src/types/encrypted_data.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Deserialize;
+use serde::Serialize;
+
+/// The structure returned after encrypting data.
```
PhilippGackstatter, line 7:
```suggestion
/// The ciphertext together with supplementary data.
```

bindings/wasm/src/account/types/cek_algorithm.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::error::WasmResult;
+
+/// Supported CEK algorithms
+#[derive(Clone, Serialize, Deserialize)]
+#[wasm_bindgen(js_name = CekAlgorithm, inspectable)]
+pub struct WasmCekAlgorithm(CekAlgorithm);
+
+#[wasm_bindgen(js_class = CekAlgorithm)]
+impl WasmCekAlgorithm {
+  /// ECDH-ES will be used as the content encryption key.
```
PhilippGackstatter, line 20: I would suggest my suggestion for the documentation for `CekAlgorithm::EcdhEs` here.

bindings/wasm/examples-account/src/node.ts
```diff
@@ -36,6 +37,8 @@ async function main() {
             return await unchecked();
         case "multiple_identities":
             return await multipleIdentities();
+        case "key_exchange":
+            return await keyExchange();
```
PhilippGackstatter, line 41: Should match the Rust account example's name. I think "encryption" is better than "key_exchange".

bindings/stronghold-nodejs/package.json
```diff
@@ -14,8 +14,7 @@
         "aarch64-apple-darwin",
         "x86_64-apple-darwin",
         "x86_64-pc-windows-msvc",
-        "x86_64-unknown-linux-gnu",
-        "x86_64-unknown-linux-musl"
```
PhilippGackstatter, line 18: Was that removed on purpose? The musl target should work again, right?

PhilippGackstatter (commented):
Looks pretty good. The `MemStore` modifications need some changes. Other than that mostly quality of life stuff, docs, typos, etc. To solve the tuple problem in Wasm, I think you can just make the ephemeral public key part of `EncryptedData`.

Not sure about the naming, i.e. `EcdhEs` vs. `ECDH_ES`, but if it's not too much work to change, I'd prefer the latter, and we can just follow JWA naming in the future, preventing further naming debates :stuck_out_tongue:.

Edit: Extending the `StorageTestSuite` is probably the most important outstanding part so both `MemStore` and `Stronghold` are tested, ideally in Rust and Wasm.

identity-account-storage/src/storage/traits.rs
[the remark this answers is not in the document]
PhilippGackstatter: I think it's still valuable to have this test, because it proves (a large chunk of) our implementation is compatible with other implementations. With an extended storage test suite we can additionally ensure a working round-trip within our implementation.

identity-account-storage/src/storage/stronghold.rs
```diff
@@ -389,11 +465,146 @@ fn sign_ed25519(client: &Client, payload: Vec<u8>, location: &KeyLocation) -> Re
...
 
   Ok(Signature::new(signature.into()))
 }
 
+pub(crate) async fn diffie_hellman(
+  client: &Client,
+  private_key: &KeyLocation,
+  public_key: [u8; X25519::PUBLIC_KEY_LENGTH],
+) -> Result<Location> {
+  let location: [u8; 32] = OsRng.sample(rand::distributions::Standard);
```
PhilippGackstatter, line 478: As a performance improvement, we can switch all instances of `OsRng` back to `ThreadRng` now that we have a dedicated musl build container. Thanks @cycraig, for confirming.

identity-account-storage/src/types/encryption/cek_algorithm.rs
```diff
@@ -0,0 +1,65 @@
...
+    &self.apv
+  }
+
+  pub fn pub_info(&self) -> &[u8] {
+    &self.pub_info
+  }
+
+  pub fn priv_info(&self) -> &[u8] {
+    &self.priv_info
+  }
```
PhilippGackstatter, line 64: This should either be documented, or alternatively be removed, and the fields made `pub`. Since this is very unlikely to change, I think making the fields public is better.

identity/Cargo.toml
```diff
@@ -50,6 +50,9 @@ send-sync-storage = ["identity-account-storage/send-sync-storage"]
 # Exposes Storage `test_suite` module.
 storage-test-suite = ["identity-account-storage/storage-test-suite"]
 
+# Enables encryption and decryption functionality.
+encryption = ["identity-account/encryption"]
```
PhilippGackstatter, line 54:
```suggestion
# Enables encryption and decryption functionality.
# Breaking changes to types and functions behind this flag are not covered by semver.
unstable-encryption = ["identity-account/encryption"]
```

It's enough to make the flag `unstable` in the `identity` crate.

PhilippGackstatter (approved):
Great work, thanks for addressing all the previous comments so thoroughly!

A few more minor comments.

PhilippGackstatter:
One more thing: The `StorageTestSuite::encryption` test should also be exposed in Wasm (`wasm/src/account/storage/test_suite.rs`), and called in `stronghold-nodejs/examples/src/tests/storage_test_suite.ts`.

I'm fine with not implementing the encryption functions for the TS `MemStore` because that seems non-trivial to do without exposing all the primitives from the Rust library, and I'd rather get this PR merged soon. Realistically, this means it won't be implemented in the TS `MemStore` for quite a while.

identity-account-storage/src/types/encryption/cek_algorithm.rs
```diff
@@ -0,0 +1,49 @@
...
+use serde::Serialize;
+
+/// Supported algorithms used to determine and potentially encrypt the content encryption key (CEK).
+#[allow(non_camel_case_types)]
+#[non_exhaustive]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub enum CekAlgorithm {
+  /// Uses the result of a Diffie-Hellman key exchange between a recipient's public key and an ephemeral secret as the
+  /// content encryption key.
+  ECDH_ES(AgreementInfo),
```
cycraig, line 14:
```suggestion
  /// Elliptic Curve Diffie-Hellman Ephemeral Static key agreement using Concat KDF.
  ECDH_ES(AgreementInfo),
```

identity-account-storage/src/types/encryption/cek_algorithm.rs
```diff
@@ -0,0 +1,49 @@
...
+impl CekAlgorithm {
+  /// Returns the JWE algorithm as a `str` slice.
+  pub const fn name(&self) -> &'static str {
+    match self {
+      CekAlgorithm::ECDH_ES(_agreement) => "ECDH-ES",
+    }
+  }
+}
+
+/// Agreement information used as the input for the concat KDF.
```
cycraig, line 26:
```suggestion
/// Agreement information used as the input for the Concat KDF.
```

identity-account-storage/src/types/encryption/encrypted_data.rs
```diff
@@ -0,0 +1,67 @@
...
+/// The ciphertext together with supplementary data.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct EncryptedData {
+  associated_data: Vec<u8>,
+  nonce: Vec<u8>,
+  tag: Vec<u8>,
+  ciphertext: Vec<u8>,
+  encrypted_cek: Vec<u8>,
+  ephemeral_public_key: Vec<u8>,
+}
```
cycraig, line 16: Maybe just make these fields public too to avoid the accessor functions? We don't need to enforce any invariants here.

```suggestion
pub struct EncryptedData {
  pub associated_data: Vec<u8>,
  pub nonce: Vec<u8>,
  pub tag: Vec<u8>,
  pub ciphertext: Vec<u8>,
  pub encrypted_cek: Vec<u8>,
  pub ephemeral_public_key: Vec<u8>,
}
```

identity-account-storage/src/types/encryption/cek_algorithm.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct AgreementInfo {
```
cycraig, line 28: Derive `Default`.

```suggestion
#[derive(Clone, Debug, Default, Deserialize, Serialize)]
pub struct AgreementInfo {
```

identity-account-storage/src/types/encryption/encrypted_data.rs
```diff
@@ -0,0 +1,67 @@
...
+
+  /// Returns a reference to the encrypted content encryption key used in the encryption.
+  pub fn encrypted_cek(&self) -> &[u8] {
+    &self.encrypted_cek
+  }
+
+  /// Returns a reference to the ephemeral public key used for generating the shared secret.
+  pub fn ephemeral_public_key(&self) -> &[u8] {
+    &self.ephemeral_public_key
+  }
```
cycraig, line 66:
```suggestion
```

identity-account-storage/src/types/encryption/encryption_algorithm.rs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Deserialize;
+use serde::Serialize;
+
+/// Supported content encryption algorithms.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum EncryptionAlgorithm {
+  AES256GCM,
```
cycraig, line 11:
```suggestion
  /// AES GCM using 256-bit key.
  AES256GCM,
```

identity-account/src/account/account.rs
```diff
@@ -307,6 +315,62 @@ where
...
   }
 
+  /// Encrypts the given `plaintext` with the specified `encryption_algorithm` and `cek_algorithm`.
+  ///
+  /// Diffie-Hellman key exchange with Concatenation Key Derivation Function will be performed to obtain the encryption
+  /// secret.
+  ///
+  /// Returns an [`EncryptedData`] instance.
+  #[cfg(feature = "encryption")]
+  pub async fn encrypt_data(
```
cycraig, line 325:
```suggestion
  /// Encrypts the given `plaintext` with the specified `encryption_algorithm` and `cek_algorithm`.
  ///
  /// Returns an [`EncryptedData`] instance.
  #[cfg(feature = "encryption")]
  pub async fn encrypt_data(
```

identity-account/src/account/account.rs
```diff
@@ -307,6 +315,62 @@ where
...
+
+  /// Decrypts the given `data` with the key identified by `fragment` using the given `encryption_algorithm` and
+  /// `cek_algorithm`.
+  ///
+  /// Diffie-Hellman key exchange with Concatenation Key Derivation Function will be performed to obtain the encryption
+  /// secret.
+  ///
+  /// Returns the decrypted text.
+  #[cfg(feature = "encryption")]
+  pub async fn decrypt_data(
```
cycraig, line 355:
```suggestion
  /// Decrypts the given `data` with the key identified by `fragment` using the given `encryption_algorithm` and
  /// `cek_algorithm`.
  ///
  /// Returns the decrypted text.
  #[cfg(feature = "encryption")]
  pub async fn decrypt_data(
```

cycraig (approved):
Great to see how far this PR has come.

Most issues have already been addressed, minor comments.

PR #815, opened 2022-04-19
Add Wasm `Credential` and `Presentation` field getters and constructors

### LINKED ISSUE #813: [Task] Improve Credential, Presentation creation in Wasm
## Description
Use duck-typed interfaces and port relevant structs to the Wasm bindings to make constructing verifiable credentials and presentations more type-safe and ergonomic.

## Motivation
Currently, constructing a `Credential` in the Wasm bindings is done through pure JSON and does not follow the duck-typed constructor pattern used throughout the bindings:

E.g.
```ts
// Create a credential subject indicating the degree earned by Alice.
const credentialSubject = {
    id: alice.document().id().toString(),
    name: "Alice",
    degreeName: "Bachelor of Science and Arts",
    degreeType: "BachelorDegree",
    GPA: "4.0"
};

// Create an unsigned `UniversityDegree` credential for Alice
const unsignedVc = Credential.extend({
    id: "https://example.edu/credentials/3732",
    type: "UniversityDegreeCredential",
    issuer: issuer.document().id().toString(),
    credentialSubject,
});
```

where `extend` is one of the last places where the undesirable `any` type is used:

```ts
static extend(value: any): Credential;
```

Similarly, constructing a `Presentation` takes `any` and lacks the options for other fields available in Rust such as setting the types, context, refresh service or policies:

```ts
// Deserialize the credential.
const receivedVc = Credential.fromJSON(signedVcJson);

// Create a Verifiable Presentation from the Credential
const unsignedVp = new Presentation(alice.document(), receivedVc)
```

where the constructor again takes `any`:

```ts
  constructor(holder_doc: Document, credential_data: any, presentation_type?: string, presentation_id?: string);
```

### Proposed Syntax

The proposal is to replace the above with duck-typed interfaces for the constructors of `Credential` and `Presentation`. These interfaces would use exported types for things like refresh services and policies but it's unclear how to work with `credentialSubject` as that is entirely user-defined. We could create an interface which defines the minimum fields required (basically just an optional `id` URI), or leave it as `any`.

E.g.

```ts
/** Fields for creating a new `Credential`. */
interface ICredential {
  /** The JSON-LD context(s) applicable to the `Credential`. */
  readonly context?: string | Record<string, any> | Array<string | Record<string, any>>;

  /** A unique `URI` referencing the subject of the `Credential`. */
  readonly id?: string;
 
  /** One or more URIs defining the type of the `Credential`. */
  readonly type?: string | Array<string>;

  /** One or more objects representing the `Credential` subject(s). */
  readonly credentialSubject: TODO ISubject | Array<ISubject> ?;

  /** A reference to the issuer of the `Credential`. */
  readonly issuer: string | IIssuer;
  
  /** A timestamp of when the `Credential` becomes valid. Defaults to the current datetime. */
  readonly issuanceDate?: Timestamp; 

  /** A timestamp of when the `Credential` should no longer be considered valid. */
  readonly expirationDate?: Timestamp; 

  /** Information used to determine the current status of the `Credential`. */
  readonly credentialStatus?: Status | Array<Status>;

  /** Information used to assist in the enforcement of a specific `Credential` structure. */
  readonly credentialSchema?: Schema | Array<Schema>;

  /** Service(s) used to refresh an expired `Credential`. */
  readonly refreshService?: RefreshService | Array<RefreshService>;

  /** Terms-of-use specified by the `Credential` issuer. */
  readonly termsOfUse?: Policy | Array<Policy>;

  /** Human-readable evidence used to support the claims within the `Credential`. */
  readonly evidence?: Evidence | Array<Evidence>;

  /** Indicates that the `Credential` must only be contained within a {@link Presentation} with a proof issued from the `Credential` subject. */
  readonly nonTransferable?: boolean;

  /** Miscellaneous properties */
  readonly properties?: Record<string | any>;
}
```

## Resources
- Verifiable Credentials Data Model specification: https://www.w3.org/TR/vc-data-model

## To-do list

- [ ] Add a duck-typed constructor for `Credential`.
- [ ] Add a duck-typed constructor for `Presentation`.
- [ ] Update examples.

## Change checklist

cycraig:
# Description of change
Adds typed constructors for the `Credential` and `Presentation` structs in the Wasm bindings, and adds Typescript interfaces for the fields contained in them.

This changes several field names in Rust for consistency and also gets rid of (I think) the last `any` function parameters in the Wasm bindings.

**NOTE:** this makes breaking changes to the Wasm examples incompatible with 0.5.0. This may confuse developers using version 0.5.0 if released to the wiki prior to 0.6.0.

### Added

- Add Wasm `Credential` field getter functions.
- Add Wasm `Presentation` field getter functions.
- Add `Evidence` Typescript interface.
- Add `Issuer` Typescript interface.
- Add `Policy` Typescript interface.
- Add `RefreshService` Typescript interface.
- Add `Schema` Typescript interface.
- Add `Status` Typescript interface.
- Add `Subject` Typescript interface.
- Add Wasm `Credential` constructor.
- Add Wasm `Credential.BaseContext`.
- Add Wasm `Credential.BaseType`.
- Add Wasm `Presentation.BaseContext`.
- Add Wasm `Presentation.BaseType`.

### Changed

- Change Wasm `Presentation` constructor to use an interface.
- Change `WasmDocument.signCredential` parameter from `any` to `Credential`.
- Change `WasmDocument.signPresentation` parameter from `any` to `Credential`.
- Rename `Refresh` to `RefreshService`.
- Rename `CredentialBuilder::refresh` to `refresh_service`.
- Rename `CredentialBuilder::policy` to `terms_of_use`.
- Rename `PresentationBuilder::refresh` to `refresh_service`.
- Rename `PresentationBuilder::policy` to `terms_of_use`.

### Removed

- Remove Wasm `Credential.issue()`.
- Remove Wasm `Credential.extend()`.

Resolves #813.

## Type of change

- [ ] Documentation Fix

Added multiple Mocha unit tests that pass with `npm run test:unit:node`. Wasm examples also pass locally.

## Change checklist


HenriqueNogara (approved):

bindings/wasm/src/credential/presentation_builder.rs
```diff
@@ -0,0 +1,110 @@
...
+  holder: Option<String>,
+  /// Service(s) used to refresh an expired {@link Credential} in the `Presentation`.
+  #[typescript(name = "refreshService", type = "RefreshService | Array<RefreshService>")]
+  refresh_service: Option<OneOrMany<RefreshService>>,
+  /// Terms-of-use specified by the `Presentation` holder.
+  #[typescript(name = "termsOfUse", type = "Policy | Array<Policy>")]
+  terms_of_use: Option<OneOrMany<Policy>>,
+  /// Miscellaneous properties.
+  #[serde(flatten)]
+  #[typescript(optional = false, name = "[properties: string | symbol]", type = "unknown")]
```
abdulmth, line 108: Same thing here.

abdulmth, line 108: Same thing here.

bindings/wasm/src/credential/credential_builder.rs
```diff
@@ -0,0 +1,161 @@
...
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "ICredential")]
+  pub type ICredential;
+}
+
+/// Fields for constructing a new {@link Credential}.
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[typescript(name = "ICredential", readonly, optional)]
```
abdulmth, line 115: I would keep all the properties mutable, not sure if there's a reason to make them readonly?

abdulmth, line 115: I would keep all the properties mutable, not sure if there's a reason to make them readonly?

bindings/wasm/examples-account/src/create_vp.ts
```diff
@@ -87,11 +87,11 @@ async function createVP(storage?: Storage) {
     // The credential is then serialized to JSON and transmitted to the holder in a secure manner.
     // Note that the credential is NOT published to the IOTA Tangle. It is sent and stored off-chain.
     const signedVcJson = signedVc.toJSON();
-    console.log(`Credential JSON > ${signedVcJson}`)
+    console.log(`Credential JSON >`, JSON.stringify(signedVcJson, null, 2));
```
abdulmth, line 90: Thanks for fixing this, it was bugging me.

abdulmth, line 90: Thanks for fixing this, it was bugging me.

abdulmth (approved):
Looks good to me, minor comments, very readable proc-macro!

bindings/wasm/src/credential/credential_builder.rs
```diff
@@ -0,0 +1,161 @@
...
+  /// Human-readable evidence used to support the claims within the `Credential`.
+  #[typescript(type = "Evidence | Array<Evidence>")]
+  evidence: Option<OneOrMany<Evidence>>,
+  /// Indicates that the `Credential` must only be contained within a {@link Presentation} with a proof issued from the
+  /// `Credential` subject.
+  #[typescript(name = "nonTransferable", type = "boolean")]
+  non_transferable: Option<bool>,
+  /// Miscellaneous properties.
+  #[serde(flatten)]
+  #[typescript(optional = false, name = "[properties: string | symbol]", type = "unknown")]
```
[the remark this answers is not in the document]
cycraig: I don't get an error setting a `Symbol` key if I remove the `readonly` property, but it doesn't show up in the logs so I think it's best to remove it, thanks.

> It also can produce undesired behavior with serde since you can set 2 keys with the same string representation

We have this problem in Rust too with the `properties` map, it's just something we have to accept can happen.

bindings/wasm/src/credential/credential_builder.rs
cycraig: Several of the TypeScript interfaces are used as return values, where it's important to note that the fields are immutable since any changes to them wouldn't reflect on the struct in Wasm memory.

In other places such as `ICredential` it's used to somewhat convey that these are only meant to be used as function parameters, not representing the `Credential` itself (missing fields like `proof` and in a real `Credential` more fields are required than here). Basically it's to discourage other, incorrect uses of the interface. If we had basic function parameters rather than an interface they wouldn't be mutable either (because they wouldn't be a separate object), so I don't think it's a major disadvantage to keep these `readonly`.

PR #834, opened 2022-05-04
Unpin iota-crypto version

olivereanderson:
# Description of change
This PR specifies the `iota-crypto` dependencies used by this library to be in the range `0.7` up to (but NOT including) `0.10`.  https://github.com/iotaledger/identity.rs/pull/822 pinned the version to `0.9.1` which was necessary because `0.9.2` introduced a breaking change causing builds to fail. Now `0.9.2` has been yanked from crates.io and pinning is therefore no longer necessary. 

## Open questions
Before #822, we specified version `^0.7` for `iota-crypto` in our crates except for libjose (the reason `0.9.2` still got pulled was because of transitive dependencies (which pinning to `0.9.1` then prevented)), but as #822 demonstrated our library also worked with `0.9.1` and I have also tested with `0.8`. Hence the question is do we use the range specified in this PR, or go back to `^0.7` or use `^0.9` which is now OK due to `0.9.2` being yanked? 

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/identity-account-storage/Cargo.toml
+++ b/identity-account-storage/Cargo.toml
@@ -31,7 +31,7 @@ tokio = { version = "1.17.0", default-features = false, features = ["sync"], opt
 zeroize = { version = "1.4" }
 
 [dependencies.iota-crypto]
-version = "=0.9.1"
+version = ">=0.7, <0.10"
 default-features = false
 features = ["blake2b", "hmac", "pbkdf", "sha", "std"]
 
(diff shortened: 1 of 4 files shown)


cycraig (approved):
Thanks, a version range is definitely the way this was meant to be done, rather than pinning.

Although we should probably stay with the latest version possible for security updates etc., I don't see any changes since 0.7.0 that should affect us, so the range is probably fine?

eike-hass (approved):

PR #847, opened 2022-05-08
Improve Napi-Wasm error conversion

PhilippGackstatter:
# Description of change

Improve the conversion of errors in Napi when going from Rust to JS. This simply includes the entire error chain in the error string, which is necessary because the new stronghold errors do not include the source errors in their string representation, as per our error guidelines.

As an example, when opening a stronghold with an incorrect password:

**Current dev**

```
Error: snapshot operation `read` on path `test_strongholds/test.stronghodl` failed
```

**This PR**

```
Error: snapshot operation `read` on path `test_strongholds/test.stronghodl` failed. Caused by: Inner error occured(corrupted file: Decryption failed: error in algorithm XCHACHA20-POLY1305).
```

For formatting, the Napi code uses a modified version of the `error_chain_fmt` function from Wasm.

[Our error guidelines](https://github.com/iotaledger/identity.rs/issues/534).

## Type of change

- [ ] Documentation Fix

Ran a local test to see if the error strings look as expected.

## Change checklist

### DIFF
--- a/bindings/wasm/src/error.rs
+++ b/bindings/wasm/src/error.rs
@@ -211,9 +211,15 @@ impl JsValueResult {
   /// Consumes the struct and returns a Result<_, AccountStorageError>
   pub fn account_err(self) -> StdResult<JsValue, AccountStorageError> {
     self.0.map_err(|js_value| {
-      // Using the debug format includes a backtrace, which is awkwardly formatted,
-      // but no other way of extracting the error, like serialization, works.
-      AccountStorageError::JsError(format!("{:?}", js_value))
+      let error_string: String = match wasm_bindgen::JsCast::dyn_into::<js_sys::Error>(js_value) {
+        Ok(js_err) => ToString::to_string(&js_err.to_string()),
+        Err(js_val) => {
+          // Fall back to debug formatting if this is not a proper JS Error instance.
+          format!("{js_val:?}")
+        }
+      };
+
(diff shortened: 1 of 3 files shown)


HenriqueNogara (approved):

bindings/wasm/src/error.rs
```diff
@@ -211,9 +211,15 @@ impl JsValueResult {
   /// Consumes the struct and returns a Result<_, AccountStorageError>
   pub fn account_err(self) -> StdResult<JsValue, AccountStorageError> {
```
olivereanderson, line 212: From the name `account_err` I would expect this to return `Option<AccountStorageError>` since `std::Result<T,E>::error` returns `Option<E>`. Would it be possible to just inline this method in the `From<JsValueResult>` for  `AccountStorageResult` implementations if that is its intended use case?

PhilippGackstatter: I wouldn't really want to inline the method into the two implementations, since that duplicates code.
How about renaming the method to `to_account_error`?

olivereanderson: `to_account_error` is better. If it is necessary for this to be a public method, then please also update the documentation to explain that the `Ok` variant is left untouched.

olivereanderson (approved):
Looks good to me. Thanks for the changes.

PR #848, opened 2022-05-08
After publication, overwrite `Account` document with published one

PhilippGackstatter:
# Description of change

In the account, after publication, we have created a version of the internal document with a signature, but we currently discard it. Instead, we should set it internally, so `Account::document` returns this document with a valid signature.

n/a, came up on Discord.

## Type of change

- [ ] Documentation Fix

A test was added to check the behavior.

## Change checklist

### DIFF
--- a/bindings/wasm/src/account/wasm_account/account.rs
+++ b/bindings/wasm/src/account/wasm_account/account.rs
@@ -65,9 +65,9 @@ impl WasmAccount {
 
   /// Returns a copy of the document managed by the `Account`.
   ///
-  /// Note that the returned document only has a valid signature after publication.
-  /// In general for use cases where the signature is required, it is advisable to resolve the
-  /// document from tangle, for instance by using `resolve_identity`.
+  /// Note: the returned document only has a valid signature after publishing an integration chain update.
+  /// In general, for use cases where the signature is required, it is advisable to resolve the
+  /// document from the Tangle.
   #[wasm_bindgen]
   pub fn document(&self) -> WasmDocument {
     let document: IotaDocument = self.0.borrow().document().clone();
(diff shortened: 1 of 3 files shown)


bindings/wasm/src/account/wasm_account/account.rs
```diff
@@ -64,6 +64,10 @@ impl WasmAccount {
   }
 
   /// Returns a copy of the document managed by the `Account`.
+  ///
+  /// Note that the returned document only has a valid signature after publication.
+  /// In general for use cases where the signature is required, it is advisable to resolve the
+  /// document from tangle, for instance by using `resolve_identity`.
```
cycraig, line 70: Comment. This is technically untrue since publishing with a diff update (which are still not removed so we have to assume they'll come back at some point) doesn't sign the document.

Function name casing is different in Wasm, removed for convenience.

```suggestion
  /// Note: the returned document only has a valid signature after publishing an integration chain update.
  /// In general, for use cases where the signature is required, it is advisable to resolve the
  /// document from the Tangle.
```

identity-account/src/account/account.rs
```diff
@@ -176,6 +176,10 @@ where
 
   /// Returns the DID document of the identity, which this account manages,
   /// with all updates applied.
+  ///
+  /// Note that the returned document only has a valid signature after publication.
+  /// In general for use cases where the signature is required, it is advisable to resolve the
+  /// document from tangle, for instance by using [`Account::resolve_identity`].
```
cycraig, line 182: Comment. This is technically untrue since publishing with a diff update (which are still not removed so we have to assume they'll come back at some point) doesn't sign the document.

Function name casing is different in Wasm, removed for convenience.

```suggestion
  ///
  /// Note: the returned document only has a valid signature after publishing an integration chain update.
  /// In general, for use cases where the signature is required, it is advisable to resolve the
  /// document from the Tangle.
```

cycraig (approved):

PR #850, opened 2022-05-09
Performance: avoid allocation when deserializing Timestamp

olivereanderson:
# Description of change
This PR is a performance improvement.

Deserialization of `Timestamp` will now go via `TryFrom::ProvisionalTimestamp` instead of `TryFrom::String`. Where `ProvisionalTimestamp<'a>(#[serde(borrow) Cow<'a, str>)` is used internally only. Benchmarks show that this makes `Timestamp` deserialization between `6%` and `10%` faster. 

  
## Type of change

- [ ] Documentation Fix

All tests run locally. Performance improvements have been tested with `cargo bench` (using `criterion`) locally. 

## Change checklist


PhilippGackstatter (approved):

identity-core/Cargo.toml
```diff
@@ -32,11 +32,16 @@ features = ["ed25519", "random", "sha", "x25519"]
...
 [dev-dependencies]
+criterion = { version = "0.3" }
 hex = { version = "0.4", default-features = false }
 proptest = { version = "1.0.0" }
 quickcheck = { version = "1.0" }
 quickcheck_macros = { version = "1.0" }
 
+[[bench]]
+name = "timestamp"
+harness = false
```
cycraig, line 43: Timestamp deserialization is not something likely to deteriorate over time nor a performance bottleneck, so I would argue against keeping a benchmark for it around.

olivereanderson:
> This breaks the Stronghold bindings:
> 
> ```
> [Example] Error: javascript function threw an exception: JsValue(Error: invalid type: string "2022-05-09T13:23:17Z", expected a borrowed string
> Error: invalid type: string "2022-05-09T13:23:17Z", expected a borrowed string
> ```

Will look into this now. If it causes too much trouble I will close the PR (thought this should be a very simple change).

cycraig:
If datetime parsing performance is important we could also consider [humantime](https://github.com/tailhook/humantime).

olivereanderson:
> If datetime parsing performance is important we could also consider [humantime](https://github.com/tailhook/humantime).

This PR is not related to any issues on our roadmap and can therefore not be considered important at this point. The only reason I started working on this is that I initially thought it would be as simple as a one line changing (changing `#[serde(try_from =String`)` to `#[serde(try_from = &str)]`) which would have been a very easy win with almost no overhead had that been the case. When that turned out to not be the case I was curious why and wanted to see how it could be fixed which is what this PR now does. 

For now I would suggest not going further down the rabbit hole. Hence we should either merge this or simply close this PR. I would personally be fine with either.

identity-core/src/common/timestamp.rs
```diff
@@ -144,6 +147,27 @@ impl TryFrom<&'_ str> for Timestamp {
...
+  }
+}
+
+impl<'a> TryFrom<ProvisionalTimestamp<'a>> for Timestamp {
+  type Error = Error;
+
+  fn try_from(value: ProvisionalTimestamp<'a>) -> Result<Self, Self::Error> {
+    Timestamp::parse(&value)
+  }
+}
```
cycraig, line 169: Nit: only used in one place so I wouldn't implement `Deref`.

```suggestion
impl<'a> TryFrom<ProvisionalTimestamp<'a>> for Timestamp {
  type Error = Error;

  fn try_from(value: ProvisionalTimestamp<'a>) -> Result<Self, Self::Error> {
    Timestamp::parse(value.0.borrow())
  }
}
```

identity-core/src/common/timestamp.rs
```diff
@@ -144,6 +147,27 @@ impl TryFrom<&'_ str> for Timestamp {
   }
 }
 
+// This struct is only used to (potentially) avoid an allocation when deserializing a Timestamp. We cannot deserialize
+// via &str because that breaks serde_json::from_value which we use extensively in the Stronghold bindings.
+// This approach is inspired by https://crates.io/crates/serde_str_helpers, but we only use a subset of the functionality offered by that crate.
+#[derive(Deserialize, Serialize)]
+struct ProvisionalTimestamp<'a>(#[serde(borrow)] Cow<'a, str>);
```
cycraig, line 154: Nit: only used for deserialising so maybe remove `Serialize` derive.

```suggestion
#[derive(Deserialize)]
struct ProvisionalTimestamp<'a>(#[serde(borrow)] Cow<'a, str>);
```

cycraig (approved):
The additional unit test and any time improvement or avoided memory allocations for Wasm are appreciated.

identity-core/src/common/timestamp.rs
cycraig: Would making this `repr(transparent)` or `serde(transparent)` be better?

cycraig: P.S. PR should have the `Wasm` label too since the performance improvement applies to all platforms, probably.

PR #854, opened 2022-05-12
Merge pull request #771 from Dr-Electron/docs/new-build-system

eike-hass:
Prepare for new wiki build system

# Description of change

## Type of change

- [ ] Documentation Fix

## Change checklist


documentation/docusaurus.config.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = {
+    plugins: [
+        [
+            '@docusaurus/plugin-content-docs',
+            {
+                id: 'identity-rs',
+                path: path.resolve(__dirname, 'docs'),
+                routeBasePath: 'identity.rs',
+                sidebarPath: path.resolve(__dirname, 'sidebars.js'),
+                editUrl: 'https://github.com/iotaledger/identity/edit/support/v0.5/documentation',
```
Dr-Electron, line 12: For dev you want to use `dev` here instead of `support/0.5`

documentation/package.json
```diff
@@ -1,18 +1,23 @@
 {
-  "name": "documentation",
+  "name": "identity",
```
Dr-Electron, line 2: Same thing with being unique here. We probably need to rename this to `identity-dev` for example. Because otherwise we have multiple configs with the same id in the wiki and yarn doesn't like that.
I hopes that just changing the version would solve that, but it expects a different name sadly.
Maybe someone here has a better solution

documentation/README.md
```diff
@@ -10,30 +10,20 @@ The documentation is built using [Docusaurus 2](https://docusaurus.io/). The dep
 ## Installation
 
 ```console
-npm i
-npm run setup
+yarn
 ```
 
-This command checks out a local copy of the wiki and creates links to the content.
+This command installs all needed dependencies.
```
cycraig, line 16: I'm getting the following error when trying to execute `yarn` locally, am I doing something wrong?

```
➤ YN0013: │ @iota-community/iota-wiki-cli@https://github.com/iota-community/iota-wiki-cli.git#commit=ef2f73d6464205353a1660413a667b6318f378a7 can't be found in the cache and will be fetched from the remote repository
➤ YN0018: │ @iota-community/iota-wiki-cli@https://github.com/iota-community/iota-wiki-cli.git#commit=ef2f73d6464205353a1660413a667b6318f378a7: The remote archive doesn't match the expected checksum
➤ YN0000: └ Completed in 1m 38s
➤ YN0000: Failed with errors in 1m 38s
```

Dr-Electron: Hmm, thats strange. You are already the second person with that problem, but I can't reproduce. I will check again

cycraig: Probably just a Windows thing, possibly because git checks out the repo with CRLF line endings while the source uses LF, leading to a different checksum? Running `yarn` in WSL works fine.

For Windows one can work around this (locally) by adding `checksumBehavior: update` to the `.yarnc.yml` file, but that does change the checksum in the `yarn.lock` file. Just need to make sure not to commit that, since it would probably break the Wiki itself.

cycraig (changes requested):
Errors for me, link needs to change:

```
Module not found: Error: Can't resolve '../../../bindings/wasm/examples-account/src/memory_storage.ts' in '/mnt/c/Work/iota/identity.rs/documentation/docs/advanced'
client (webpack 5.72.1) compiled with 1 error
```

Should point to `../../../bindings/wasm/examples-account/src/custom_storage.ts`.

Edit: changed.

cycraig (approved):

eike-hass:
@Dr-Electron do you have strong opinions about the package naming to make it unique and do we have to tackle this before we merge?

documentation/README.md
Dr-Electron: Yeah after having 3 people under windows with the same problem we came to the conclusion that may be the problem.
Maybe we can find a nice solution for that. In the meantime just use https://gitpod.io/ :trollface:

Dr-Electron:
> @Dr-Electron do you have strong opinions about the package naming to make it unique and do we have to tackle this before we merge?

Actually not really. I thought you may also give some feedback there. In the end I would probably just have appended the version. Like `identity-develop`, `identity-production` or `identity-v0.5`. 
And yes i would actually do that before merge. I meant there isn't really a better solution (at least not for me 😅 ). The only alternative I can come up with is having some logic in the wiki that replaces the string in the file on build. But I hate that 😅 

So in the end I think we should add a command to the cli which would prepare your branch for a specific version by replacing the id, the edit url and everything else which needs to be changed. Wdyt?

eike-hass:
> > @Dr-Electron do you have strong opinions about the package naming to make it unique and do we have to tackle this before we merge?
> 
> Actually not really. I thought you may also give some feedback there. In the end I would probably just have appended the version. Like `identity-develop`, `identity-production` or `identity-v0.5`. And yes i would actually do that before merge. I meant there isn't really a better solution (at least not for me 😅 ). The only alternative I can come up with is having some logic in the wiki that replaces the string in the file on build. But I hate that 😅
> 
> So in the end I think we should add a command to the cli which would prepare your branch for a specific version by replacing the id, the edit url and everything else which needs to be changed. Wdyt?

That would be very comfortable indeed! Will change it to `identity-develop` for now.

Dr-Electron:
> > > @Dr-Electron do you have strong opinions about the package naming to make it unique and do we have to tackle this before we merge?
> > 
> > 
> > Actually not really. I thought you may also give some feedback there. In the end I would probably just have appended the version. Like `identity-develop`, `identity-production` or `identity-v0.5`. And yes i would actually do that before merge. I meant there isn't really a better solution (at least not for me 😅 ). The only alternative I can come up with is having some logic in the wiki that replaces the string in the file on build. But I hate that 😅
> > So in the end I think we should add a command to the cli which would prepare your branch for a specific version by replacing the id, the edit url and everything else which needs to be changed. Wdyt?
> 
> That would be very comfortable indeed! Will change it to `identity-develop` for now.

Can you do that now so that I could use this branch for development now? Want to update the versioning branch and check if there is something else we need

eike-hass:
> > > > @Dr-Electron do you have strong opinions about the package naming to make it unique and do we have to tackle this before we merge?
> > > 
> > > 
> > > Actually not really. I thought you may also give some feedback there. In the end I would probably just have appended the version. Like `identity-develop`, `identity-production` or `identity-v0.5`. And yes i would actually do that before merge. I meant there isn't really a better solution (at least not for me 😅 ). The only alternative I can come up with is having some logic in the wiki that replaces the string in the file on build. But I hate that 😅
> > > So in the end I think we should add a command to the cli which would prepare your branch for a specific version by replacing the id, the edit url and everything else which needs to be changed. Wdyt?
> > 
> > 
> > That would be very comfortable indeed! Will change it to `identity-develop` for now.
> 
> Can you do that now so that I could use this branch for development? Want to update the versioning branch and check if there is something else we need

Added the unique names. Will merge now. Feel free to test with dev or create a branch form it :)

PR #855, opened 2022-05-12
Use relative links

Dr-Electron:
# Description of change
 Hi, it's me again 😅 
There is another problem with using the @site alias as it resolves to the directory of the room `docusaurus.config` which in our case is in the wiki. I could create a custom alias (like I did here: https://github.com/Dr-Electron/identity.rs/commit/7c60dd423b3accef59a35d26514629931ba69441) which would work, but later on with versioning I would need a unique custom alias for every version. Which isn't really maintainable 😅.

So I guess we just go for relative links. Maybe I can come up with something more clever somewhere down the road but for now it shouldn't really be a problem 😅 

## Type of change

- [x] Documentation Fix

Tested this locally and tried to add it to the wiki build.

## Change checklist

### DIFF
--- a/documentation/docs/libraries/wasm/api_reference.md
+++ b/documentation/docs/libraries/wasm/api_reference.md
@@ -9,7 +9,7 @@ keywords:
 ---
 
 ```mdx-code-block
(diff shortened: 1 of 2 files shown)


eike-hass (approved):
LGTM!

PR #861, opened 2022-05-30
Add `RevocationBitmap2022`, bump MSRV to 1.60

### LINKED ISSUE #853: [Task] Implement on-tangle Revocation Method
## Description

Following the discussion https://github.com/iotaledger/identity.rs/discussions/806, implement a scalable ad secure on tangle revocation method.

Things we have discussed and/or settled on:

- Working title: `SimpleRevocationList2022` or `RevocationBitSet2022`
- How does a credential link to such a service?
  - [`credentialStatus`](https://www.w3.org/TR/vc-data-model/#status)
- `serviceEndpoint` MUST follow the data URL scheme for embedding the list
  - We can later add support for other URIs for forwards compatibility if we want to permit lists to be hosted externally
- Determine and include hard and soft limits
  - hard limit of maximum message size on tangle (`max_revokable_credentials`)
  - Soft limit given by POW and byte cost
  - Another soft limit can described as the trade-off for issuers on how many credentials they expect to be revoked. It's debatable whether this needs to be part of the spec, though, this will ultimately be an issuer's decision.
    - E.g. if they need to be able to revoke 100% of their issued credentials, then they cannot issue more than `max_revokable_credentials`.
    - But if they only need to be able to revoke 10% of all issued credentials, they can issue `max_revokable_credentials * 10` credentials.
- Deserialization produces a `Service` struct, which can lazily be decompressed into a bitset, thus only incurring costs for those needing to access the revocation list.
- Decide on the serialization format
  - Decide on zlib or brotli for compression.
  - Decide on bitset or roaring bitmap for byte format.
- Bonus: Possibly describe minimum bitset size for herd privacy?

## Motivation

We should consider an on-tangle scheme to have the guarantees towards immutability and availability as the DIDs themself. The storage on the tangle needs to be efficient to reduce the PoW or Bytecost requirements.

## Resources

Further details can be found in the discussion  https://github.com/iotaledger/identity.rs/discussions/806.

## Change checklist

HenriqueNogara:
# Description of change
Add EmbeddedRevocationList for allowing managing credential status.

Closes #853 

## Type of change

- [ ] Documentation Fix

Unit tests + revoked_vc example.

## Change checklist


identity-iota/src/credential/credential_validator.rs
```diff
@@ -170,6 +175,72 @@ impl CredentialValidator {
...
+        Err(ValidationError::InvalidService(ServiceError::InvalidServiceId(
+          format!("{} not found", credential_status.id),
+        )))
+      }
+    }
+  }
+
+  /// By deserializing the `Url` path into a `EmbeddedRevocationList`, checks if the index given in
+  /// `EmbeddedRevocationStatus` is revoked.
+  fn check_simple_revocation_list(
```
PhilippGackstatter, line 228: Should match the name of the revocation method.

identity-iota/src/credential/credential_validator.rs
```diff
@@ -170,6 +175,72 @@ impl CredentialValidator {
...
+    credential_status: &EmbeddedRevocationStatus,
+    data_url: &EmbeddedRevocationEndpoint,
+  ) -> ValidationUnitResult {
+    let revocation_list: EmbeddedRevocationList = data_url
+      .clone()
```
PhilippGackstatter, line 233: Can we avoid this clone by taking an owned `EmbeddedRevocationEndpoint`?

identity-iota/src/credential/errors.rs
```diff
@@ -53,6 +53,18 @@ pub enum ValidationError {
...
   MissingPresentationHolder,
+  /// Indicates that status is invalid.
+  #[error("invalid service: {0}")]
+  InvalidStatus(#[source] identity_iota_core::credential::CredentialStatusError),
+  /// Indicates that revocation list is invalid.
+  #[error("invalid service: {0}")]
+  InvalidRevocationList(#[source] identity_iota_core::revocation::RevocationMethodError),
+  /// Indicates that the service is invalid.
+  #[error("invalid service: {0}")]
+  InvalidService(#[source] identity_iota_core::service::ServiceError),
```
PhilippGackstatter, line 64: Don't include a `source` error in the error message.

identity-iota/src/tangle/traits.rs
```diff
@@ -41,16 +41,16 @@ impl TangleRef for DiffMessage {
...
   }
 
   fn previous_message_id(&self) -> &MessageId {
     self.previous_message_id()
   }
 
-  fn set_previous_message_id(&mut self, message_id: MessageId) {
-    self.set_previous_message_id(message_id);
+  fn set_previous_message_id(&mut self, _message_id: MessageId) {
+    self.set_previous_message_id(_message_id);
```
PhilippGackstatter, line 53: Is that a leftover temporary change or something?

examples/account/revoke_vc.rs
```diff
@@ -59,11 +79,17 @@ async fn main() -> Result<()> {
     "GPA": "4.0",
   }))?;
 
-  // Build credential using subject above and issuer.
+  // Create a credential status pointing verifiers to the endpoint that states if it has been revoked.
+  let service_url = IotaDIDUrl::parse(format!("{}#my-revocation-service", issuer.did()))?;
+  let credential_index: u32 = 5; // choosen arbitrarily
```
PhilippGackstatter, line 84:
```suggestion
  // The issuer determines the index of the credential in the revocation list.
  // We choose one arbitrarily here.
  let credential_index: u32 = 5;
```

PhilippGackstatter (changes requested):
Only reviewed Rust for now. Will review Wasm when the Rust design is more settled on.

identity-core/Cargo.toml
```diff
@@ -11,6 +11,7 @@ repository = "https://github.com/iotaledger/identity.rs"
 description = "The core traits and types for the identity-rs library."
 
 [dependencies]
+base64 = { version = "0.13", default-features = false, features = ["std"] }
```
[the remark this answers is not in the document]
cycraig: Depends on whether or not we want parity with `RevocationList2020` in the future, since `multibase` uses a sentinel character prefix to indicate the encoding.

https://github.com/multiformats/multibase#multibase-table

In this case `base64url` with no padding would be prefixed with `u`.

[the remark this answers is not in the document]
PhilippGackstatter: > Depends on whether or not we want parity with RevocationList2020

I think since we use roaring bitmap, our revocation method will not be compatible with `RevocationList2020` anyway, if that's what you mean. But maybe I misunderstand.

> In this case base64url with no padding would be prefixed with u.

In that case, I don't think we want to change our spec to multibase since it doesn't seem to have a particular advantage for our case (?), so I guess we can leave this as-is.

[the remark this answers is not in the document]
cycraig: That's news to me then, I was still under the impression we were going with a compressed bitstring serialization with a roaring bitmap internal representation.

> In that case, I don't think we want to change our spec to multibase since it doesn't seem to have a particular advantage for our case (?), so I guess we can leave this as-is.

The benefit of `base64` over `multibase` is better interoperability since Base64 is more common, but implementers will need `multibase` regardless since it's used in `VerificationMethod`.

The other argument is fitting the [Data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme). I originally used `data:,` since we don't need to decode the contents and so the media type defaults to `text/plain;charset=US-ASCII`. We could instead use something like `data:application/octet-stream;base64,` and then use proper `base64` encoding for the bitmap.

identity-iota/src/tangle/traits.rs
cycraig: Clippy incorrectly (hopefully) thinks the function is recursive, maybe it should be called explicitly as `DiffMessage::set_previous_message_id(self, message_id)`?

examples/account/revoke_vc.rs
```diff
@@ -59,11 +79,17 @@ async fn main() -> Result<()> {
...
+  let status: EmbeddedRevocationStatus = EmbeddedRevocationStatus::new(service_url, credential_index);
+
+  // Build credential using subject above, status, and issuer.
```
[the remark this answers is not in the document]
cycraig: Leave my Oxford commas alone :P

identity-credential/src/credential/status.rs
```diff
@@ -14,31 +13,21 @@ pub struct Status {
...
-  where
-    T: Into<OneOrMany<String>>,
-  {
-    Self {
-      id,
-      types: types.into(),
-      properties,
-    }
+  pub fn with_properties(id: Url, types: Option<String>, properties: Object) -> Self {
+    Self { id, types, properties }
```
cycraig, line 30: Nit: rename this `new_with_properties` for consistency.

identity-credential/src/credential/status.rs
```diff
@@ -14,31 +13,21 @@ pub struct Status {
   pub id: Url,
   /// The type(s) of the credential status.
   #[serde(rename = "type")]
-  pub types: OneOrMany<String>,
+  pub types: Option<String>,
```
cycraig, line 16: I believe `type` is a required field for `credentialStatus`?

identity-iota-core/src/revocation/embedded_revocation_list.rs
```diff
@@ -0,0 +1,211 @@
...
+use super::error::Result;
+use super::error::RevocationMethodError;
+use crate::service::EmbeddedRevocationEndpoint;
+
+pub const EMBEDDED_REVOCATION_METHOD_NAME: &str = "EmbeddedRevocationList";
+const REVOCATION_LIST_INDEX: &str = "revocationListIndex";
+
+/// A compressed bitset for managing credential revocation.
+#[derive(Clone, Debug, PartialEq)]
+pub struct EmbeddedRevocationList(RoaringBitmap);
```
cycraig, line 32: Nit: should just be named `RevocationBitmap`.

identity-iota-core/src/revocation/embedded_revocation_list.rs
```diff
@@ -0,0 +1,211 @@
...
+
+impl EmbeddedRevocationList {
+  /// Returns the name of the revocation method.
+  pub fn name() -> &'static str {
+    EMBEDDED_REVOCATION_METHOD_NAME
+  }
```
cycraig, line 38: The containing service should specify the name/type, this is a low-level struct with different concerns.

identity-iota-core/src/revocation/embedded_revocation_list.rs
```diff
@@ -0,0 +1,211 @@
...
+
+  // Returns the name of the property that contains the index of the credential to be checked.
+  pub fn credential_list_index_property() -> &'static str {
```
cycraig, line 41: Similar to `name`, this function should be placed at a higher level, in the containing service if at all.

identity-iota-core/src/revocation/embedded_revocation_list.rs
```diff
@@ -0,0 +1,211 @@
...
+  pub fn revoke(&mut self, index: u32) -> bool {
+    self.0.insert(index)
+  }
+
+  /// Given the index of multiple credentials, revoke all.
+  pub fn revoke_multiple(&mut self, indexes: &[u32]) {
+    for i in indexes {
+      self.revoke(*i);
+    }
+  }
```
cycraig, line 65: I don't think this needs a dedicated method, the caller just just as easily run a for-loop over `revoke`.

identity-iota-core/src/revocation/embedded_revocation_list.rs
```diff
@@ -0,0 +1,211 @@
...
+  /// The credential at the given `index` will be set to valid.
+  pub fn undo_revocation(&mut self, index: u32) -> bool {
+    self.0.remove(index)
+  }
+
+  /// Serializes and compressess the [`EmbeddedRevocationList`] and returns its data url representation
+  pub fn to_embedded_service_endpoint(&self) -> Result<EmbeddedRevocationEndpoint> {
+    let data_url: String = format!("data:,{}", self.serialize_compressed_b64()?);
+    EmbeddedRevocationEndpoint::parse(&data_url).map_err(|_| RevocationMethodError::InvalidUrlRepresentation(data_url))
+  }
```
cycraig, line 76: Nit: this should probably be a `TryFrom` implementation and specified in `EmbeddedRevocationEndpoint`.

identity-iota-core/Cargo.toml
```diff
@@ -12,9 +12,13 @@ description = ""
 
 [dependencies]
 bee-message = { version = "0.1.6", default-features = false, features = ["serde"] }
+data-url = { version = "0.1.1", default-features = false }
+flate2 = { version = "1.0.23" }
 identity-core = { version = "=0.5.0", path = "../identity-core", default-features = false }
+identity-credential = { version = "=0.5.0", path = "../identity-credential", default-features = false }
```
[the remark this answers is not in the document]
PhilippGackstatter: True, `identity-did` makes more sense, thanks! Feature-gating also seems like a good idea.

cycraig (commented):
Brief comments for now.

The main requests are related project layout: move implementations to `identity-did` and `identity-credential`.

This also raises how we should handle concrete implementations of abstractions like `Service` and `credentialStatus`. There are several approaches:

## Option 1: Properties

This was the approach used by `MerkleKeyCollection` and previously `IotaDocument`. Since things like `VerificationMethod` and `Service` are generic over the properties type (defaulting to `Object = BTreeMap<String, Value>`), we can instead choose another struct like 

```rust
struct RevocationProperties {
  revocationListIndex: Option<u32>,
}
```

And use `Service<RevocationProperties>` instead, or use `From/TryFrom` implementation to convert between `Service<Object>` and `Service<RevocationProperties>`.

**Note that this only works to represent additional fields, not replace or restrict existing fields.**

## Option 2: Structs

This is the approach taken by the current PR where a specific `Service` type is implemented as a separate struct, in this case `EmbeddedRevocationService`. Conversions use `From/TryFrom` as-needed.

This is useful for adding fields as well as placing restrictions on existing ones.

## Option 3: Structs + Enums

The last approach moves away from `From/TryFrom` implementations and instead leverages `serde` to deserialize to the correct type.

For example, we can use tagged enum variants to deserialize the intended type directly.

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "type")]
enum Service {
    EmbeddedRevocationService(EmbeddedRevocationService),
    DIDCommMessaging(DIDCommService),
    ...
}
```

This would mean less explicit conversions but more matching in code.

Unfortunately, variant tags need to be exhaustive since `serde` does not support a "catch-all" tag. Otherwise we need to use an `untagged` enum and write custom deserializer functions to check the tag manually.

```rust
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
enum Service {
    EmbeddedRevocationService(EmbeddedRevocationService),
    DIDCommMessaging(DIDCommService),
    Service(GenericService),
}

fn deserialize_embedded_revocation_service_type(...) {}
fn deserialize_didcomm_service_type(...) {}

#[derive(Serialize, Deserialize)]
#[serde(untagged)]
enum EmbeddedRevocationService {
  #[serde(rename = "type", deserialize_with = "deserialize_embedded_revocation_service_type")]
  pub(crate) type_: String,
  ...
}
```

The `deserialize_*_tag` functions would reject unmatching tags and fallthrough to `GenericService`.

---

If we think we might support more concrete `Service` implementations in the future then I would argue for `Option 3`, otherwise the current concrete struct approach will work.

cycraig:
> And indeed, we will have at least one more concrete service type for DIDComm in the future, so this would lay some good groundwork for that implementation.

I think we should leave that for another PR then, to avoid delaying the 0.6.0. It requires refactoring quite a lot of code and some manual workarounds for the Wasm bindings.

identity-did/src/service/bitmap_revocation_service.rs
```diff
@@ -0,0 +1,144 @@
...
+  pub(crate) type_: String,
+  pub(crate) service_endpoint: BitmapRevocationEndpoint,
+}
+
+impl<D: DID + Sized> BitmapRevocationService<D> {
+  /// Creates a new [`BitmapRevocationService`].
+  pub fn new(id: DIDUrl<D>, service_endpoint: BitmapRevocationEndpoint) -> Result<Self> {
+    if id.fragment().unwrap_or_default().is_empty() {
+      return Err(Error::InvalidService("invalid id - missing fragment"));
+    }
```
cycraig, line 29: I'm not sure how useful a `BitmapRevocationService` is anymore. It's not used to enforce any (de)serialization invariants and `BitmapRevocationEndpoint` could be replaced by `RevocationBitmap` directly in most places.

identity-did/src/revocation/bitmap.rs
```diff
@@ -0,0 +1,158 @@
...
+  pub fn revoke(&mut self, index: u32) -> bool {
+    self.0.insert(index)
+  }
+
+  /// The credential at the given `index` will be set to valid.
+  ///
+  /// Returns ture is the value was present in the set.
+  pub fn undo_revocation(&mut self, index: u32) -> bool {
+    self.0.remove(index)
+  }
```
cycraig, line 52: Fix doc-comment, don't mention `valid`, that has a specific, different meaning in terms of verifiable credentials.

Suggest renaming to `unrevoke` for brevity.

```suggestion
  /// Mark the `index` as not revoked.
  ///
  /// Returns true if the `index` was present in the set.
  pub fn unrevoke(&mut self, index: u32) -> bool {
    self.0.remove(index)
  }
```

identity-did/src/revocation/bitmap.rs
```diff
@@ -0,0 +1,158 @@
...
+      .map_err(Error::DeserializationError)
+      .map(Self)
+  }
+
+  /// Serializes a [`RevocationBitmap`] as a vector of bytes.
+  pub fn serialize_vec(&self) -> Result<Vec<u8>> {
+    let mut output: Vec<u8> = Vec::with_capacity(self.0.serialized_size());
+    self.0.serialize_into(&mut output).map_err(Error::SerializationError)?;
+    Ok(output)
+  }
```
cycraig, line 79: Not sure if we should be exporting many of these serialize/deserialize methods as public functions.

identity-did/src/revocation/bitmap.rs
```diff
@@ -0,0 +1,158 @@
...
+  pub fn is_revoked(&self, index: u32) -> bool {
+    self.0.contains(index)
+  }
+
+  /// Revokes the credential at the given `index`.
+  ///
+  /// Return whether the value was absent from the set.
+  pub fn revoke(&mut self, index: u32) -> bool {
+    self.0.insert(index)
+  }
```
cycraig, line 45:
```suggestion
  /// Mark the given `index` as revoked.
  ///
  /// Returns true if the `index` was absent from the set.
  pub fn revoke(&mut self, index: u32) -> bool {
    self.0.insert(index)
  }
```

identity-did/src/revocation/bitmap.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// A compressed bitmap for managing credential revocation.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct RevocationBitmap(RoaringBitmap);
+
+impl RevocationBitmap {
+  /// Creates a new [`RevocationBitmap`].
+  pub fn new() -> Self {
+    Self(RoaringBitmap::new())
+  }
```
cycraig, line 33:
```suggestion
  /// Constructs a new empty [`RevocationBitmap`].
  pub fn new() -> Self {
    Self(RoaringBitmap::new())
  }
```

identity-did/src/error.rs
```diff
@@ -54,4 +54,18 @@ pub enum Error {
...
+  #[error("unable to compress data")]
+  CompressionError(#[source] std::io::Error),
+  #[error("unable to decompress data")]
+  DecompressionError(#[source] std::io::Error),
+  #[error("revocation list could not be deserialized")]
+  DeserializationError(#[source] std::io::Error),
+  #[error("invalid bitmap revocation endpoint: {0}")]
+  InvalidBitmapEndpoint(&'static str),
+  #[error("revocation list could not be serialized")]
+  SerializationError(#[source] std::io::Error),
```
cycraig, line 70: Not a fan of the extensive error variant bloat. 

`Compression`/`Serialization` can be combined, as can `Decompression`/`Deserialization`.

identity-did/Cargo.toml
```diff
@@ -12,10 +12,13 @@ description = "An implementation of the Decentralized Identifiers standard."
 
 [dependencies]
 async-trait = { version = "0.1", default-features = false }
+data-url = { version = "0.1.1", default-features = false }
 did_url = { version = "0.1", default-features = false, features = ["std", "serde"] }
+flate2 = { version = "1.0.23" }
 form_urlencoded = { version = "1.0.1", default-features = false }
 identity-core = { version = "=0.5.0", path = "../identity-core" }
 indexmap = { version = "1.7", default-features = false, features = ["std", "serde-1"] }
+roaring = { version = "0.9.0", default-features = false }
```
cycraig, line 21: Would be nice to feature-gate `RevocationBitmap` functionality.

identity-credential/src/credential/bitmap_revocation_status.rs
```diff
@@ -0,0 +1,127 @@
...
+use identity_core::common::Value;
+use identity_did::did::DIDUrl;
+use identity_did::did::DID;
+use identity_did::service::BITMAP_REVOCATION_METHOD;
+
+use super::Status;
+use crate::error::Error;
+use crate::error::Result;
+
+const REVOCATION_LIST_INDEX: &str = "revocationListIndex";
```
cycraig, line 17: Suggestion: rename to `revocationBitmapIndex`?

```suggestion

const INDEX_PROPERTY_NAME: &str = "revocationBitmapIndex";
```

identity-credential/src/credential/bitmap_revocation_status.rs
```diff
@@ -0,0 +1,127 @@
+// Copyright 2020-2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
cycraig, line 2:
```suggestion
// Copyright 2020-2022 IOTA Stiftung
// SPDX-License-Identifier: Apache-2.0
```

identity-credential/src/credential/bitmap_revocation_status.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use super::Status;
+use crate::error::Error;
+use crate::error::Result;
+
+const REVOCATION_LIST_INDEX: &str = "revocationListIndex";
+
+/// Information used to determine the current status of a [`Credential`][crate::credential::Credential].
+#[derive(Clone, Debug, PartialEq)]
+pub struct BitmapRevocationStatus<D: DID + Sized> {
```
cycraig, line 21: Rename to `RevocationBitmapStatus` to be consistent with `RevocationBitmap2022`?

identity-core/src/utils/base_encoding.rs
```diff
@@ -115,6 +115,22 @@ where
...
+}
+
+/// Encodes the given `data` as base64.
+pub fn encode_b64<T>(data: &T) -> String
+where
+  T: AsRef<[u8]> + ?Sized,
+{
+  base64::encode_config(data.as_ref(), base64::URL_SAFE)
+}
+
```
cycraig, line 133: Ideally we should have some basic unit tests for these functions, like the other base encodings.

Have we settled on whether to use multibase instead?

identity-account/src/error.rs
```diff
@@ -35,6 +35,9 @@ pub enum Error {
   UpdateError(#[from] crate::updates::UpdateError),
   #[error("method missing fragment")]
   MethodMissingFragment,
+  /// Caused by failing to revoke credentials.
+  #[error("credential revocation error")]
+  CredentialRevocationError(#[source] identity_did::Error),
```
cycraig, line 40: Could just be `RevocationError`? Is there anything else we can revoke?

examples/account/revoke_vc.rs
```diff
@@ -74,6 +101,23 @@ async fn main() -> Result<()> {
   // Revoke the Verifiable Credential.
   // ===========================================================================
 
+  // Update the service for checking the credential status
+  // When verifiers look for the index corresponding to the credential, it will be set to revoked.
+  issuer
+    .revoke_credentials("my-revocation-service", &[credential_index])
+    .await?;
+
+  let validation_result = CredentialValidator::validate(
```
cycraig, line 110:
```suggestion   
  // Update the RevocationBitmap service in the issuer's DID Document.
  // This revokes the credential's unique index.
  issuer
    .revoke_credentials("my-revocation-service", &[credential_index])
    .await?;

  // Credential verification now fails.
  let validation_result = CredentialValidator::validate(
```

cycraig (changes requested):
Still some large questions around the dedicated structs approach and smaller issues like `base64` vs `multibase`. Also naming needs to be consistent across the library and bindings.

Overall it's unclear whether we should go with dedicated structs or continue with the `VerificationMethod` approach used elsewhere in the codebase, where supported types get their own constructors (and consider adding e.g. `ServiceType`, `StatusType` enums).

E.g.
```rust
impl Status<D> {
  new_revocation_bitmap_status(id: DIDUrl<D>, index: u32)
}

impl Service<D> {
  new_revocation_bitmap_service(id: DIDUrl<D>, bitmap: RevocationBitmap);
}
```

cycraig:
We can get rid of `base64` since `multibase` exposes encoding, decoding functionality for all its bases by default (without the prefix character).

E.g.
```rust
let encoded: String = multibase::Base::Base64Url.encode(&input);
let decoded: Vec<u8> = multibase::Base::Base64Url.decode(&encoded)?;
```

identity-core/Cargo.toml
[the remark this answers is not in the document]
cycraig: https://github.com/iotaledger/identity.rs/pull/861#issuecomment-1149822713

identity-iota/src/credential/errors.rs
```diff
@@ -53,6 +53,18 @@ pub enum ValidationError {
...
+  InvalidStatus(#[source] identity_credential::Error),
+  /// Indicates that the credential's enpoint is invalid.
+  #[error("invalid credential endpoint")]
+  InvalidEndpoint(#[source] identity_did::Error),
+  /// Indicates that the the credential's service is invalid.
+  #[error("invalid service")]
+  InvalidService(#[source] identity_did::Error),
+  /// Indicates that the credential has been revoked.
+  #[error("credential at index {0} has been revoked")]
+  RevokedCredential(u32),
```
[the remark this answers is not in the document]
PhilippGackstatter: The presentation validator re-uses the `ValidationError` type from here, so if we don't include it here, that information is lost, I believe.

identity-iota/src/credential/credential_validator.rs
```diff
@@ -170,6 +175,70 @@ impl CredentialValidator {
     }
   }
 
+  /// If the credential has a status of type [`BitmapRevocationStatus`], checks if the credential has been revoked.
+  pub fn check_revoked<T, D: AsRef<IotaDocument>>(credential: &Credential<T>, issuers: &[D]) -> ValidationUnitResult {
+    match &credential.credential_status {
+      Some(status) => {
+        if status.type_ != BITMAP_REVOCATION_METHOD {
+          return Ok(());
+        }
```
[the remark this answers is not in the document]
PhilippGackstatter: Changed it to error by default on unrecognized status types.

identity-iota/src/credential/errors.rs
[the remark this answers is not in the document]
cycraig: The index information is in `CompoundPresentationValidationError`, as far as I know. None of the other errors record the index explicitly.

```rust
/// An error caused by a failure to validate a Presentation.
pub struct CompoundPresentationValidationError {
  pub credential_errors: BTreeMap<usize, CompoundCredentialValidationError>,
  pub presentation_validation_errors: Vec<ValidationError>,
}
```

PhilippGackstatter:
I think I addressed all comments. Got rid of some types and (hopefully) simplified things overall. The only weirdness is now that `RevocationBitmapStatus` wraps a `Status<CoreDID>` instead of being generic, because that would mean also making `Credential` generic over `DID`, which seems unnecessary overall. Let me know what you think of that.

`Status::id` was changed from `Url` to `DIDUrl<D> where D: DID`, mostly because that makes it easier to build on top of. It's similar to `Service` now. I'm not entirely sure if this is spec compliant, as the specs point mandates that both `id` fields are `URI`s and not `URL`s.

cycraig:
> `Status::id` was changed from `Url` to `DIDUrl<D> where D: DID` mostly because that makes it easier to build on top of. I'm not entirely sure if this is spec compliant, as the specs point mandates that both id fields are URIs and not URLs.

Nope, change it back to `Url`. Many of our previous decisions have been based purely on following the specification, even if it's inconvenient. Deviating from the specification means losing potential interoperability with other implementations and any deviations means we cannot claim to be compliant.

https://www.w3.org/TR/vc-data-model/#status

identity-account/src/account/account.rs
```diff
@@ -315,6 +316,36 @@ where
     Ok(())
   }
 
+  /// If the document has a [`RevocationBitmapService`] identified by `fragment`,
+  /// revokes all credentials identified by the given `credential_indices`.
+  pub async fn revoke_credentials(&mut self, fragment: &str, credential_indices: &[u32]) -> Result<()> {
+    // Finds the service to be updated
+    let mut service_id: IotaDIDUrl = self.did().clone().into_url();
+    service_id.set_fragment(Some(fragment))?;
```
cycraig, line 324:
```suggestion
    // Find the service to be updated.
    let mut service_id: IotaDIDUrl = self.did().to_url();
    service_id.set_fragment(Some(fragment))?;
```

identity-account/src/account/account.rs
```diff
@@ -315,6 +316,36 @@ where
     Ok(())
   }
 
+  /// If the document has a [`RevocationBitmapService`] identified by `fragment`,
+  /// revokes all credentials identified by the given `credential_indices`.
+  pub async fn revoke_credentials(&mut self, fragment: &str, credential_indices: &[u32]) -> Result<()> {
```
cycraig, line 321: Do we need an equivalent `unrevoke_credentials` function in the `Account`? Should we move the bulk of the logic to `IotaDocument` instead so it can be used from the low-level bindings too?

identity-credential/src/credential/revocation_bitmap_status.rs
```diff
@@ -0,0 +1,116 @@
...
+use identity_did::did::DIDUrl;
+use identity_did::did::DID;
+
+use super::Status;
+use crate::error::Error;
+use crate::error::Result;
+
+/// Information used to determine the current status of a [`Credential`][crate::credential::Credential].
+#[derive(Clone, Debug, PartialEq)]
+pub struct RevocationBitmapStatus(Status);
```
cycraig, line 18: This could use the now-generic properties on `Status` to make things a bit easier? Or does it complicate things?

```suggestion
struct RevocationBitmapProperties {
  #[serde(rename = "revocationBitmapIndex")]
  pub revocation_bitmap_index: u32,
  #[serde(flatten)]
  pub properties: Object,
}

#[derive(Clone, Debug, PartialEq)]
pub struct RevocationBitmapStatus(Status<RevocationBitmapProperties>);
```

identity-did/src/error.rs
```diff
@@ -54,4 +54,11 @@ pub enum Error {
 
   #[error("signature verification failed: {0}")]
   InvalidSignature(&'static str),
+
+  #[error("unable to decode base64 string: `{0}`")]
+  Base64DecodingError(String, #[source] identity_core::error::Error),
+  #[error("revocation list could not be deserialized or decompressed")]
+  BitmapDecodingError(#[source] std::io::Error),
+  #[error("revocation list could not be serialized or compressed")]
+  BitmapEncodingError(#[source] std::io::Error),
```
cycraig, line 63: Nit: error messages.

```suggestion
  #[error("revocation bitmap could not be deserialized or decompressed")]
  BitmapDecodingError(#[source] std::io::Error),
  #[error("revocation bitmap could not be serialized or compressed")]
  BitmapEncodingError(#[source] std::io::Error),
```

cycraig (approved):
Thanks to @HenriqueNogara and @PhilippGackstatter for all the hard work. Minor changes left.

Not a big fan of the extra bytes taken up by using `application/octet-stream;base64` but it's probably more correct.

Edit: I appreciate the simplified approach. I think `RevocationBitmapStatus` could also be removed (or only used internally) in favour of something like `Status::new_revocation_bitmap_status`, but the struct could prove useful for developers?

Just needs documentation/formal specification on the wiki.

identity-credential/Cargo.toml
```diff
@@ -12,7 +12,7 @@ description = "An implementation of the Verfiable Credentials standard."
 
 [dependencies]
 identity-core = { version = "=0.5.0", path = "../identity-core" }
-identity-did = { version = "=0.5.0", path = "../identity-did" }
+identity-did = { version = "=0.5.0", path = "../identity-did", default-features = false, features = ["revocation-bitmap"] }
```
cycraig, line 15: I don't think the `revocation-bitmap` feature is ever actually used in `identity-credential`.

Maybe it needs to be a top-level feature in `identity` and `identity-account`/`identity-iota` (if the `(un)revoke_credential` logic is moved there too)?

identity-credential/src/credential/revocation_bitmap_status.rs
PhilippGackstatter: Then the constructor takes a `Status<Object>` but needs to transform it to a `Status<RevocationBitmapProperties>`, and that's not easily possible without going through `serde_json::Value`, or is it?

cycraig: The constructor would use `Status::new_with_properties` instead, passing a new `RevocationBitmapProperties` to it. Any conversion from `Status<RevocationBitmapProperties>` to `Status<Object>` might need to go through `serde` though?

Probably not worth the hassle, I don't think this would affect developers much, I think it's mostly internal.

identity/Cargo.toml
```diff
@@ -51,7 +51,11 @@ send-sync-storage = ["identity-account-storage/send-sync-storage"]
 storage-test-suite = ["identity-account-storage/storage-test-suite"]
 
 # Enables revocation with `RevocationBitmap2022`.
-revocation-bitmap = ["identity-account?/revocation-bitmap", "identity-iota/revocation-bitmap"]
+revocation-bitmap = [
+  "identity-account?/revocation-bitmap",
```
cycraig, line 55: This optional feature syntax is only available in Rust 1.60+, need to bump the MSRV.

cycraig: Should also add `revocation-bitmap` to the `default` feature sets.

bindings/wasm/src/revocation/bitmap.rs
```diff
@@ -0,0 +1,80 @@
...
+
+  /// Return the bitmap as a data url embedded in a service endpoint.
+  #[wasm_bindgen(js_name = toEndpoint)]
+  pub fn to_enpdoint(&self) -> Result<UServiceEndpoint> {
+    Ok(service_endpoint_to_js_value(&self.0.to_endpoint().wasm_result()?))
+  }
+
+  /// Construct a `RevocationBitmap` from a data `url`.
+  #[wasm_bindgen(js_name = fromEndpoint)]
+  pub fn from_endpoint(url: &str) -> Result<WasmRevocationBitmap> {
```
cycraig, line 63: This should actually take a `UServiceEndpoint` since that's what `WasmService::serviceEndpoint` returns.

identity-account/Cargo.toml
```diff
@@ -37,3 +37,6 @@ async = ["identity-iota/async"]
 send-sync-storage = ["identity-account-storage/send-sync-storage"]
 # Enables encryption and decryption functionality.
 encryption = ["identity-account-storage/encryption"]
+
+# Enables revocation with `RevocationBitmap2022`.
+revocation-bitmap = ["identity-iota/revocation-bitmap"]
```
cycraig, line 42: Should be `default`.

identity/Cargo.toml
PhilippGackstatter: This seems like a nice application of the new syntax, but I'm also fine not using it.

PhilippGackstatter (approved):
Thanks for adding the `StatusCheck` and the comprehensive test. Looks good to me.

identity/Cargo.toml
cycraig: Changed MSRV to 1.60.

PR #877, opened 2022-06-12
Change `remove_service` to return boolean

cycraig:
# Description of change
Small change to make `remove_service` consistent with `insert_service`.

## Type of change

- [ ] Documentation Fix

Unit tests pass locally.

## Change checklist


abdulmth (approved):
Looks good, could the other way around be better, though? To force people to deal with the error if inserting/removing fails?

cycraig:
> Looks good, could the other way around be better, though? To force people to deal with the error if inserting/removing fails?

Well, we never returned an error in the first place, only `Ok(())`, so there was no information on whether or not the operation actually succeeded. Most times deletion is meant to be idempotent anyway, and not error on missing values.

We moved away from errors on most collection operations since handling them is cumbersome (particularly in other languages such as JavaScript) and it pollutes the error enums. Particularly here where the only two errors would be `DuplicateServiceId` on `insert_service` and `MissingServiceId` on `remove_service`.

Issue #869, opened 2022-06-08
[Task] Release to crates.io


eike-hass:
## Description
In order to release to crates.io we need rename our packages and setup the automatic release process.

## To-do list

- [x] Rename crates - #895
- [x] Exclude `examples` (see https://github.com/iotaledger/identity.rs/runs/6837111260?check_suite_focus=true where examples would be published) - #895
- [x] Publish crates.
- [ ] setup CI to push automatically

cycraig:
Still need to rename packages and actually release.

cycraig:
I've added `release = false` to the `examples` package per the `cargo-release` FAQ: https://github.com/crate-ci/cargo-release/blob/c2f0d52a1c1a02cf1fdc6e68106954111f7e2c56/docs/faq.md#why-does-publishfalse-still-release

> ## Why does publish=false still release?
> publish only controls whether to publish to crates.io. Some programs might want version bumping, tagging, replacements, etc without publishing.
> 
> If you do want to disable everything, set release = false.

cycraig:
I don't think `cargo-release` was going to publish `examples` either way, it does distinguish `release` from `publish` specifically, and it says `release` in the logs:

```
Release
  identity-diff-derive 0.6.0
  identity-diff 0.6.0
  identity-core 0.6.0
  identity-did 0.6.0
  identity-iota-core 0.6.0
  identity-account-storage 0.6.0
  identity-credential 0.6.0
  identity-iota 0.6.0
  identity-account 0.6.0
  identity 0.6.0
  examples 0.6.0
```

eike-hass:
Let's look at at automatic publishing separately.

Issue #868, opened 2022-06-06
[Task] Improve Tests for the Encrypting/Decrypting Functionality 


HenriqueNogara:
## Description
Improve existing tests for the encrypting/decrypting Functionality. Perhaps allow the setting of ephemeral keys values. 

## Motivation
Make sure the functionality is implemented right by using existing tests vector.

## To-do list
- [ ] [Allow the Setting of ephemeral key values](https://github.com/iotaledger/identity.rs/pull/809#discussion_r889236329).
- [ ] [Test correctness beyond roundtrip encryption ](https://github.com/iotaledger/identity.rs/pull/809#discussion_r889251065)

PR #867, opened 2022-06-03
Implement `ECDH-ES+A256KW` for `Storage` encryption

PhilippGackstatter:
# Description of change

Implements `ECDH-ES+A256KW` in the `Storage` encryption API, for use in `DIDComm`.

## Links to any relevant issues

Part of #841

## Type of change
Add an `x` to the boxes that are relevant to your changes.

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

## How the change has been tested

`StorageTestSuite` was modified to also test the newly added `CekAlgorithm`.

## Change checklist
Add an `x` to the boxes that are relevant to your changes.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #881, opened 2022-06-13
Use `iota_stronghold` from crates.io

PhilippGackstatter:
# Description of change

Use the stronghold version on crates.io instead of pinning to GitHub.

required for #869.

## Type of change

- [ ] Documentation Fix

Built the crates locally.

## Change checklist

### DIFF
--- a/identity-account-storage/Cargo.toml
+++ b/identity-account-storage/Cargo.toml
@@ -19,6 +19,7 @@ hashbrown = { version = "0.11", features = ["serde"] }
 identity-core = { version = "=0.5.0", path = "../identity-core", default-features = false }
 identity-did = { version = "=0.5.0", path = "../identity-did", default-features = false }
 identity-iota-core = { version = "=0.5.0", path = "../identity-iota-core", default-features = false }
+iota_stronghold = { version = "0.5.1", default-features = false, features = ["std"], optional = true }
 once_cell = { version = "1.7", default-features = false, features = ["std"], optional = true }
 parking_lot = { version = "0.12" }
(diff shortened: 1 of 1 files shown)


cycraig:
Can we just merge https://github.com/iotaledger/identity.rs/pull/867 instead? it's going to cause problems after the crates are renamed otherwise.

eike-hass:
> Can we just merge #867 instead? It's going to cause problems after the crates are renamed otherwise.
The concern was, that we can't prove the correctness of the implementation as tracked in #868. Would you guys be in favor of merging anyways for the upcoming release?

cycraig:
> we can't prove the correctness of the implementation 

Basically the same problem with #809, I see no difference really.

eike-hass:
> > we can't prove the correctness of the implementation
> 
> Basically the same problem with #809 and that was merged, I see no difference really.

@PhilippGackstatter what do you think?

PhilippGackstatter:
I didn't have the time to check whether #867 is compatible with other crates in the ecosystem, so this was the quicker solution. Yeah, we also can't prove correctness like #809 just yet, not sure if that justifies adding more code that's not provably correct, though. But if this complicates the crate renaming somehow, I'm also fine with merging #867.

eike-hass:
> I didn't have the time to check whether #867 is compatible with other crates in the ecosystem, so this was the quicker solution. Yeah, we also can't prove correctness like #809 just yet, not sure if that justifies adding more code that's not provably correct, though. But if this complicates the crate renaming somehow, I'm also fine with merging #867.

Then let's go for that, sorry for the confusion.

PR #936, opened 2022-07-05
Fix brittle agent test

PhilippGackstatter:
# Description of change

Fixes an agent test that appears to be brittle on some systems or platforms or situations - it's unclear what exactly the problem is. 

The order of addresses and addresses themselves returned from `Agent::addresses` aren't the same across executions, so the occassional failures appear to come from times when the first address that is picked isn't dialable.
This is fixed by adding all addresses for an agent, which allows libp2p to attempt a dial for all of them.

Also fixes the crate description to match the README.

n/a

## Type of change

- [ ] Documentation Fix

The test was run on the affected system and appears to work consistently now.

## Change checklist


olivereanderson (approved):
Looks good to me. Shame we cannot find the exact cause of this now.

PR #946, opened 2022-07-16
Fix stack overflow in `CoreDID` `PartialEq` impl

PhilippGackstatter:
# Description of change

Fixes an infinite recursion when comparing `CoreDID`s. The following example currently causes a stack overflow:

```rust
let example1 = CoreDID::parse("did:example:0x0").unwrap();
let example2 = CoreDID::parse("did:example:0x1").unwrap();
println!("{}", example1 == &example2);
```

The Rust compiler ends up effectively calling:

```rust
<CoreDID as PartialEq<&CoreDID>>::eq(&example1, &&example2)
```

which recurses infinitely. The implementation is:

```rust
impl PartialEq<&CoreDID> for CoreDID {
  fn eq(&self, other: &&CoreDID) -> bool {
    self == other
  }
}
```

Two options for fixing the situation: 1) Using `*other` or 2) removing the impl. We don't use this implementation anywhere, and since this impl has been around for 9 months without anyone running into the problem, there's a good chance no one is using it. Note that this is an additional implementation next to the derived `PartialEq` impl on `CoreDID`. So this PR removes this impl. However, if we don't want this to be a breaking change or think it's a convenient impl, we can also go with option 1.

Edit: CI is failing due to a known ICE: https://github.com/rust-lang/rust/issues/99261

n/a

## Type of change

- [ ] Documentation Fix

Ran the above example before and after the change.

## Change checklist

### DIFF
--- a/identity_did/src/did/did.rs
+++ b/identity_did/src/did/did.rs
@@ -277,12 +277,6 @@ impl PartialEq<&'_ str> for CoreDID {
   }
 }
 
-impl PartialEq<&CoreDID> for CoreDID {
-  fn eq(&self, other: &&CoreDID) -> bool {
-    self == other
-  }
-}
-
 impl KeyComparable for CoreDID {
   type Key = CoreDID;
 


cycraig (approved):
Not sure why that `PartialEq` existed (my fault for adding it, just not sure why, apologies). It could have been from when unsafe was used to transmute transparent DID types but that's just a guess.

Thanks for fixing it! Fully agree to remove the `impl`, no loss of functionality since `CoreDID` still derives `PartialEq`.

Issue #942, opened 2022-07-12
[Task] DID document alias output serialization


PhilippGackstatter:
## Description

Design and implement how DID documents are stored in the state metadata of an alias output. Effects on performance and gas cost should be investigated, as well as ease of integration into the framework and by others implementing support for DIDs in alias outputs.
Moreover, consider whether the state metadata document needs to contain the `DID`.

## Motivation

DID documents are serialized with JSON according to the W3C spec. However, in the interest of easing interoperability from smart contracts, which are expected to build on top of the [various output types](https://github.com/iotaledger/tips/pull/38), a simpler serialization format might be justified. (De)serializing JSON in a smart contract might incur higher gas cost and reduce performance compared to a simpler byte layout serialization.

Since the identifier of an alias output is unknown until the output is published to the ledger, there is a need for a placeholder DID in the document. For consistency, we can consider to never store the `AliasId` (which is effectively the DID), in the state metadata, even after it is known, so we don't have the situation where the document in the state metadata sometimes contains the correct did, and sometimes the placeholder. This would also save a few bytes and therefore reduce the required storage deposit.

## Resources

Part of #908.

## To-do list
Create a task-specific to-do list. Please link PRs that match the TODO list item behind the item after it has been submitted.

- [ ] Investigate smart contract interoperability.
- [ ] Implement the serialization format, if any.
- [x] Consider a mechanism to remove the DID from the state metadata document #947

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

Issue #908, opened 2022-06-17
[Task] Stardust Method


cycraig:
## Description
This is the tracking issue for tasks required for the Stardust UTXO-based DID Method.

A small proof-of-concept is on the [`feat/stardust-poc`](https://github.com/iotaledger/identity.rs/tree/feat/stardust-poc) branch, which simply embeds an empty DID Document in the `stateMetadata` of an Alias Output. The `identity_stardust` crate on that branch is separate from the workspace, this is due to incompatible dependencies related to `iota-client`. I suggest continuing parallel development with the new and old DID Methods for the least friction, since the alternative would break all existing tests and examples that interact with the Tangle. Development should continue in isolation, with generic components incrementally being updated to support multiple DID Methods.

Note that the current intention is not to release `did:stardust` as a separate DID Method, but rather use it as a working title; the `did:iota` method (at least in name) will be replaced with this in the future.

## Notes

There were a number of issues encountered during development of the proof-of-concept, most minor.

1. The `AliasId` is inferred from the `Block`, so need to use a placeholder DID when creating a new DID Document, or publish twice: first to get the `AliasId`, then publish the DID Document. Publishing twice seems ill-advised and likely to cause more problems than fix them; the proof-of-concept resolution simply replaces all occurrences of the placeholder DID when extracting the DID Document.
2. Cannot get an `OutputId` (needed to resolve an `AliasOutput`) back from an `AliasId` (hash of `OutputId`), need to use Indexer API. This is fine since `iota-client` provides functions to do so, just needs a few steps and some type wrangling.
3. The response from the Indexer is an Output, not a `Block`, so cannot infer an `AliasId` from it. Fine since we use the `AliasId` from the DID to retrieve the `Output` in the first place. The `OutputDto` conversion is a bit annoying though.
4. The pieces needed to publish an update are fragmented (`OutputId` for input, amount, DID Document, etc.), bit annoying to reconstruct. Maybe use a holder struct like `Holder { AliasOutput, StardustDocument }` with convenience functions?
5. Inferred fields such as the `controller` and `governor` need to reflect in the (JSON) Document but excluded from the
`StardustDocument` serialization when published. Handle with a separate `pack` function like with current DID Messages?
[...]

PR #947, opened 2022-07-16
State metadata serialization for the stardust DID method

PhilippGackstatter:
# Description of change

Adds a `StateMetadataDocument` that represents the serialized DID document in an alias output for the stardust DID method.

The `AliasId`, which corresponds to the DID, is not computable before the alias output itself has been created. So a placeholder is required in absence of the DID. In order to avoid having to publish twice when creating the alias output, so that the DID document contained in the alias output has the correct DID set, the approach is to always use the placeholder. During resolution a transformation step that takes the block or alias output, and the contained DID document as inputs, produces the actual `StardustDocument` that contains the correct DID. This means that our solution to this issue is consistent no matter if the alias output existed or is newly created.

Since we'd like to enable smart contracts to build on top of the DID document in the alias output, we might not be able to use an expensive serialization format or even compression. Thus the shortest valid `CoreDID` is chosen as the placeholder, which is `did:0:0`. All references to the DID of the document itself are replaced by this DID. When using JSON, a simple string replace suffices to replace this string with the actual DID during the transformation. However, it's likely that we will not end up using JSON, due to determinism considerations and perhaps because other serialization formats offer tighter packing, so the lack of compression doesn't weigh in as much. This PR introduces a dedicated `StateMetadataDocument` that handles conversion to and from `StardustDocument`. `StateMetadataDocument` can eventually be used to implement the serialization format of choice. The trade-off is that the conversion requires quite a few lines of code.

In general, always using a small placeholder DID has significant effects on the size of the document. For example, a `StateMetadataDocument` with 10 verification methods is about half the size of the corresponding `StardustDocument`, when converted to a vector of uncompressed json bytes. That in turn saves byte costs.

part of #942 and #908.

## Type of change

- [ ] Documentation Fix

A transformation roundtrip (`StateMetadataDocument` <-> `StardustDocument`) test was added.

## Change checklist


identity_core/src/common/ordered_set.rs
```diff
@@ -108,6 +108,11 @@ impl<T> OrderedSet<T> {
     self.0
   }
 
+  /// Returns a mutable reference to the underlying vector.
+  pub fn as_vec_mut(&mut self) -> &mut Vec<T> {
+    &mut self.0
+  }
```
cycraig, line 114: This is unnecessary and can be removed.

If kept it should be `#[inline]` and be named `as_vec_mut_unchecked`, similar to `iter_mut_unchecked`.

```suggestion
```

identity_stardust/examples/create_did.rs
```diff
@@ -77,15 +83,64 @@ async fn main() -> anyhow::Result<()> {
...
+    .unwrap();
+
+  document
+    .tmp_add_service(
+      key_did.clone(),
+      "#my-foreign-service",
+      "RevocationList2030",
+      identity_did::service::ServiceEndpoint::One(Url::parse("https://example.com/f4c42e9da").unwrap()),
+    )
+    .unwrap();
```
cycraig, line 127: We should move this to a unit test to keep the example clean.

identity_stardust/src/did_or_placeholder.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// A [`CoreDID`] or a placeholder.
+#[derive(Debug, Hash, Clone, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum DIDOrPlaceholder {
+  #[serde(with = "placeholder_serde")]
+  Placeholder,
+  Core(CoreDID),
+}
```
cycraig, line 27: I'm hoping we can find a solution without needing `DIDOrPlaceholder`. It is more performant since it avoids creating many placeholder `did:0:0` instances but it also means lots more code to maintain, including the DID impl.

identity_stardust/src/did_or_placeholder.rs
```diff
@@ -0,0 +1,198 @@
...
+use identity_did::did::DIDError;
+use identity_did::did::DIDUrl;
+use identity_did::did::DID;
+use serde::Deserialize;
+use serde::Serialize;
+
+lazy_static::lazy_static! {
+  pub static ref PLACEHOLDER_DID: CoreDID =
+  CoreDID::parse("did:0:0").unwrap();
+}
```
cycraig, line 18: We should switch to `once_cell`. Apparently that's the preferred replacement for `lazy_static` now (news to me too).

identity_stardust/src/stardust_document.rs
```diff
@@ -51,8 +50,46 @@ impl StardustDocument {
...
+  pub fn tmp_id(&self) -> &CoreDID {
+    self.0.id()
+  }
+
   /// Returns the placeholder DID of newly constructed DID Documents,
-  /// `"did:stardust:0000000000000000000000000000000000000000000000000000000000000000"`.
+  /// `"did:0:0"`.
   // TODO: generalise to take network name?
   pub fn placeholder_did() -> &'static CoreDID {
     &PLACEHOLDER_DID
```
cycraig, line 95: This only works because `StardustDocument` is using `CoreDID`. When `StardustDID` is a thing, `did:0:0` will not pass validation checks on creation (which is fine, since it's only replaced during serialization).

identity_stardust/src/state_metadata_document.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    for service in self.0.service_mut().as_vec_mut().drain(..) {
+      builder = builder.service(service.map(|did| did.unwrap_or_else(original_did_ctor)));
+    }
+
+    let mut core_doc: CoreDocument = builder.build().expect("all required document fields should be set");
+
+    std::mem::swap(core_doc.also_known_as_mut(), self.0.also_known_as_mut());
+
+    StardustDocument(core_doc)
```
cycraig, line 104: All of this is unnecessary. This is why the `map` function exists.

 ```suggestion
     let mut core_doc: CoreDocument<DIDOrPlaceholder> = self.0;
    let core_document: CoreDocument = core_doc.map(
      // Replace placeholder identifiers.
      |did| match did {
        DIDOrPlaceholder::Placeholder => original_did.clone(),
        DIDOrPlaceholder::Core(core) => core,
      },
      // Do not modify properties.
      |o| o,
    );
    StardustDocument(core_document)
```

File: identity_stardust/src/state_metadata_document.rs [hunk 7]
```diff
@@ -0,0 +1,326 @@
...
+      builder = builder.service(service.map(|did| DIDOrPlaceholder::new(did, &original_did)));
+    }
+
+    let mut core_doc: CoreDocument<DIDOrPlaceholder, _, _, _> =
+      builder.build().expect("all required document fields should be set");
+
+    std::mem::swap(core_doc.also_known_as_mut(), doc.0.also_known_as_mut());
+
+    StateMetadataDocument(core_doc)
+  }
```
cycraig, line 195: Same here: use the `map` function.

```suggestion
  fn from(document: StardustDocument) -> Self {
    let core_document: CoreDocument = CoreDocument::from(document);
    let id = core_document.id().clone();
    let core_doc: CoreDocument<DIDOrPlaceholder> = core_document.map(
      // Replace self-referential identifiers with a placeholder, but not others.
      // TODO: can this just be a CoreDID like PLACEHOLDER_DID instead?
      |did| {
        if did == id {
          DIDOrPlaceholder::Placeholder
        } else {
          DIDOrPlaceholder::Core(did)
        }
      },
      // Do not modify properties.
      |o| o,
    );
    StateMetadataDocument(core_doc)
  }
```

identity_stardust/src/state_metadata_document.rs
```diff
@@ -0,0 +1,326 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::did_or_placeholder::DIDOrPlaceholder;
+use crate::did_or_placeholder::PLACEHOLDER_DID;
+use crate::StardustDocument;
+
+/// The DID document as it is contained in the state metadata of an alias output.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct StateMetadataDocument(CoreDocument<DIDOrPlaceholder>);
```
cycraig, line 18: This should be private and only used when `StardustDocument` is `pack`-ed into the state metadata.

identity_stardust/src/did_or_placeholder.rs
```diff
@@ -0,0 +1,198 @@
...
+
+/// A [`CoreDID`] or a placeholder.
+#[derive(Debug, Hash, Clone, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum DIDOrPlaceholder {
```
cycraig, line 23: Should be private.

cycraig (changes requested):
The code can be simplified but I agree with the overall approach and reasoning (i.e. why string replacement won't work with alternative serializations).

Can we avoid `DIDOrPlaceholder` somehow or is it necessary for performance?

There are also a few other missing pieces:
- We still need something similar to `DIDMessage` with a version flag, and maybe a format flag and magic bytes "DID" prefix.
- The `controllers` and all other fields like `id` that (should be) inferred from the `AliasOutput` should be removed from the serialization. Maybe. We might need to keep the `id` field if that's difficult to remove and to keep the `network` information.

Otherwise looking good, let's keep iterating.

identity_stardust/src/state_metadata_document.rs
cycraig: Requires a simple `From` impl on `StardustDocument`.

```rust
impl From<StardustDocument> for CoreDocument {
  fn from(document: StardustDocument) -> Self {
    // TODO: convert any additional properties (that cannot be inferred from the Output) into
    //       JSON Object or an alternative struct in CoreDocument.
    document.0
  }
}
```

identity_stardust/src/stardust_document.rs
PhilippGackstatter: For this case, we can introduce another `STARDUST_PLACHOLDER_DID` or something that is the minimally valid stardust DID. Since the replacement procedure doesn't care about the did of self, that should work. I'd leave it for now and leave this to the implementation of `StardustDID`.

PhilippGackstatter:
Thanks for the simplifications! Not sure how I ended up making it so unnecessarily complicated.

How would it be possible to remove the `controllers` from the serialization? Only in the case where the only controller is the `id` itself, as far as I can see. Is that what you mean? Otherwise we lose information if controller != id.

Agreed, we have to keep the network information in the serialization. Since it's unclear what serialization format we'll end up using, I'd leave this for the PR that eventually closes 942. Implementing removal of `id` now would require some custom serialization implementation and that may or may not be wasted work.

identity_stardust/examples/create_did.rs
PhilippGackstatter: I'd wait for the implementation of `StardustDocument` to be more mature, otherwise this requires moving almost the entire example into a unit test. The serialization transformation itself is tested already.

PhilippGackstatter: I'd wait for the implementation of `StardustDocument` to be more mature, otherwise this requires moving almost the entire example into a unit test. The serialization transformation itself is tested already.

cycraig: Unless the implementation of `StardustDocument` also generalises `CoreDocument` to always take `CoreDID` for `Service` and `VerificationMethod`, this example would not work after `StardustDID` is implemented anyway, since it takes foreign DID Methods.

```rust
  let example_did = CoreDID::parse("did:example:0xffffffffffffffff").unwrap();
  let key_did = CoreDID::parse("did:key:0xf43958394883939484848484").unwrap();
```

These should just be different `StardustDID` instances.

```rust
  let did1 = CoreDID::parse("did:stardust:0xffffffffffffffffffffffffffffffff").unwrap();
  let did2 = CoreDID::parse("did:stardust:0xf4395839488393948484848484848484").unwrap();
```

cycraig:
> How would it be possible to remove the controllers from the serialization? Only in the case where the only controller is the id itself, as far as I can see. Is that what you mean? Otherwise we lose information if controller != id.

The controller will be the state controller of the `AliasOutput`, so it's determined by the `AliasOutput` unlock conditions.

identity_stardust/src/stardust_document.rs
```diff
@@ -51,8 +50,52 @@ impl StardustDocument {
...
+  /// Temporary testing implementation.
+  pub fn tmp_id(&self) -> &CoreDID {
+    self.0.id()
+  }
+
+  /// Serializes the document for inclusion in an alias output's state metadata.
+  pub fn into_state_metadata_bytes(self, encoding: StateMetadataEncoding) -> Result<Vec<u8>> {
+    let state_metadata_doc = StateMetadataDocument::from(self);
+    state_metadata_doc.pack(encoding)
+  }
```
cycraig, line 95: This should probably just be `pack`, and `encoding` should be optional or moved to another function? Doesn't matter too much right now, will need an API pass later.

Aside: it would be nice to be able to use the [packable](https://docs.rs/packable/latest/packable/trait.Packable.html) like `bee`/`iota.rs` does, but that's beside the point.

identity_stardust/src/state_metadata/document.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/// Prepends the message flags and marker magic bytes to the data in the following order:
+/// `[version, marker, encoding, data]`.
+fn add_flags_to_message(mut data: Vec<u8>, version: StateMetadataVersion, encoding: StateMetadataEncoding) -> Vec<u8> {
+  let mut buffer: Vec<u8> = Vec::with_capacity(1 + DID_MARKER.len() + 1 + data.len());
+  buffer.push(version as u8);
+  buffer.extend_from_slice(DID_MARKER);
+  buffer.push(encoding as u8);
+  buffer.append(&mut data);
```
cycraig, line 102: `DID_MARKER` should still be first, before the version tag (if we keep it). Previous decision was wrong.
Doesn't have to be changed now.

cycraig (approved):

olivereanderson (approved):
LGTM

Issue #945, opened 2022-07-13
[Task] Implement StardustDID 


olivereanderson:
## Description
Implement the equivalent of the `IotaDID`, `CoreDID` types for Stardust. 

## Resources
Sub-task of https://github.com/iotaledger/identity.rs/issues/908. 

## To-do list
Create a task-specific to-do list. Please link PRs that match the TODO list item behind the item after it has been submitted.

- [ ] Make type convert to and from Strings of the form `did:stardust:<network>:<tag=AliasID>` 
- [ ] Make type compatible with the initial `AliasID := 0x000000000.....`. Determine if there needs to be one "independent copy" of this ID per supported network.  

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #949, opened 2022-07-18
StardustDID

olivereanderson:
# Description of change
Introduces a `StardustDID`struct that: 
- Has Constructors enforcing compliance with the new Stardust UTXO-based DID method. 
- Implements the `DID` trait. 
- Implements conversion traits for related structs such as `CoreDID`, `String`, etc. 
  

## What is NOT included in this PR
- Incorporation into the `StardustDocument`.
- `Network` to represent networks accepted by the `StardustDID` method specification. This should be incorporated in a separate PR (possibly together with the new `Client`).  **EDIT 22.07.2022: `NetworkName` has been added now.** 
 - Wrapper for the JS bindings .
 - Changes to `Base` and `BaseEncoding` to support the prefixed `0x` hex encoded format used to represent `AliasId` (not sure if this is needed). 
 - A port of the `Segments` struct from the current IOTA method. I felt it was easy enough to work without it and could not find any use outside of its defining crate in the `identity.rs` workspace. The only public constructor has hidden documentation so I believe it is safe to assume that there is also very little use of this API by library consumers. 
 
Fixes issue #945. 

## Type of change

- [ ] Documentation Fix

Runs under `cargo test` and `cargo test --all-features` locally. 

## Change checklist


identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,929 @@
...
+  fn from(id: AliasId) -> Self {
+    let did_str = format!("did:{}:{}", StardustDID::METHOD, id);
+    Self::parse(did_str).unwrap_or_else(|_| {
+      panic!(
+        "transforming an AliasId to a {} DID should be infallible",
+        StardustDID::METHOD
+      )
+    })
+  }
+}
```
cycraig, line 365: We may not want this due to the `network` requirement. I think something like a feature-gated `from_alias_id(alias_id, Option<network>)` function might be better?

cycraig (changes requested):
Few changes.

> A port of the Segments struct from the current IOTA method. I felt it was easy enough to work without it and could not find any use outside of its defining crate in the identity.rs workspace.
`Segments` was meant for internal use anyway, so that's fine. Its real benefit was reducing duplicate code and complexity. There are quite a few places where the iterator matching is duplicated or gets a bit complicated this PR.

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,929 @@
...
+
+  /// Creates a new [`StardustDID`] with the given network.
+  ///
+  /// # Errors
+  /// `Err` is returned if the network name does not satisfy the requirements of the [`StardustDID`] method
+  /// specification.
+  // TODO: consider refactoring to use `NetworkName` once that gets ported along with the `Client`.
+  pub fn new_with_network(network: &str) -> Result<Self> {
+    Self::check_network_str(network)?;
+    CoreDID::parse(&format!("did:stardust:{}:{}", network, INITIAL_ALIAS_ID))
```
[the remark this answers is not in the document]
olivereanderson: Thanks. I forgot to change it here :+1:

identity_stardust/src/did/stardust_did.rs
olivereanderson: I will create a new issue about `Network` related tasks that need to be addressed in `StardustDID` once the network related functionality has been ported.

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,929 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+
+use identity_did::did::BaseDIDUrl;
+use identity_did::did::CoreDID;
+use identity_did::did::DIDError;
+use identity_did::did::DIDUrl;
+use identity_did::did::DID;
+
+pub type Result<T> = std::result::Result<T, DIDError>;
```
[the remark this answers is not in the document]
olivereanderson: I thought to leave that for later when we need to propagate errors from `StardustDID`. We can then evaluate whether `DIDError` should be wrapped in the top level Error, or if another approach like mapping to a higher level error that is then wrapped in the crate level error would be more appropriate. Not saying that `DIDError` shouldn't be wrapped, just trying to get into the habit of not auto wrapping every error from every crate in the workspace.

identity_stardust/README.md
```diff
@@ -3,3 +3,5 @@
 This is a work-in-progress intended to replace the `did:iota` DID Method.
 
 `cargo run --example create_did`
+
+Requires `rustc >= 1.62`.
```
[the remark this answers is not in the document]
olivereanderson: > What new features of Rust 1.62 does this use? then_some?

Yes `then_some`. Nothing else as far as I know.

identity_stardust/src/did/stardust_did.rs
olivereanderson: Edit: Decided to introduce `NetworkName` now in this PR to deal with these things right away.

olivereanderson:
@cycraig thanks for your review. Most of your comments should be addressed now. I have now decided to introduce `NetworkName` in order to produce more convenient constructors here. The `Network` has still yet to be ported however and once that is done we might have to reconsider some of the method signatures of `StardustDID` to take that into account. 

The `NetworkName` in this PR is mostly a copy of the old version with a few tiny modifications.

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,927 @@
...
+  pub fn try_from_core(did: CoreDID) -> Result<Self> {
+    Self::check_validity(&did)?;
+
+    Ok(Self(Self::normalize(did)))
+  }
+
+  /// Constructs a new [`StardustDID`] from a byte representation of the tag and the given network name.
+  pub fn new<T: Into<NetworkName>>(bytes: &[u8; 32], network_name: T) -> Self {
+    let alias_id = AliasId::from(*bytes);
+    let did: String = format!("did:{}:{}:{}", Self::METHOD, network_name.into(), alias_id);
```
cycraig, line 68: The `AliasId` usage here makes it impossible to properly feature-gate the `iota-client` dependency. The options to avoid this are either: use the `AliasId` from the `bee-block` crate directly (less preferred since we would have to keep it in sync with the version `iota-client` uses); or use `prefix-hex` directly to format the tag string.

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,927 @@
...
+  ///
+  /// Returns `Err` if the input does not conform to the [`StardustDID`] specification.
+  pub fn try_from_core(did: CoreDID) -> Result<Self> {
+    Self::check_validity(&did)?;
+
+    Ok(Self(Self::normalize(did)))
+  }
+
+  /// Constructs a new [`StardustDID`] from a byte representation of the tag and the given network name.
+  pub fn new<T: Into<NetworkName>>(bytes: &[u8; 32], network_name: T) -> Self {
```
cycraig, line 66: Aside: a doc-test might be nice to show what it looks like. Optional.

```suggestion
  /// Constructs a new [`StardustDID`] from a byte representation of the tag and the given network name.
  ///
  /// See also [`StardustDID::placeholder`].
  ///
  /// # Example
  ///
  /// ```
  /// # use identity_did::did::DID;
  /// # use identity_stardust::NetworkName;
  /// # use identity_stardust::StardustDID;
  /// #
  /// let placeholder = StardustDID::new(&[1;32], NetworkName::try_from("smr").unwrap());
  /// assert_eq!(placeholder.as_str(), "did:stardust:smr:0x0101010101010101010101010101010101010101010101010101010101010101");
  /// ```
  pub fn new<T: Into<NetworkName>>(bytes: &[u8; 32], network_name: T) -> Self {
```

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,927 @@
...
+  ///
+  /// # Errors
+  ///
+  /// Returns `Err` if the input does not conform to the [`StardustDID`] specification.
+  pub fn parse(input: impl AsRef<str>) -> Result<Self> {
+    CoreDID::parse(input).and_then(Self::try_from_core)
+  }
+
+  /// Creates a new placeholder [`StardustDID`] with the given network name.
+  pub fn placeholder(network_name: NetworkName) -> Self {
```
cycraig, line 83: Aside: a doc-test might be nice to show what it looks like. Optional.

```suggestion
  /// Creates a new placeholder [`StardustDID`] with the given network name.
  ///
  /// # Example
  ///
  /// ```
  /// # use identity_did::did::DID;
  /// # use identity_stardust::NetworkName;
  /// # use identity_stardust::StardustDID;
  /// #
  /// let placeholder = StardustDID::placeholder(NetworkName::try_from("smr").unwrap());
  /// assert_eq!(placeholder.as_str(), "did:stardust:smr:0x0000000000000000000000000000000000000000000000000000000000000000");
  /// ```
  pub fn placeholder(network_name: NetworkName) -> Self {
```

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,927 @@
...
+    Self::new(&[0; 32], network_name)
+  }
+
+  #[cfg(feature = "alias-id")]
+  pub fn from_alias_id(alias_id: AliasId, network_name: NetworkName) -> Self {
+    Self::new(&alias_id, network_name)
+  }
```
cycraig, line 90: Not really necessary, it's just as easy to call `new` as done in the function body here.

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,927 @@
...
+
+  /// Checks if the given `DID` has a valid [`StardustDID`] `method_id`.
+  ///
+  /// # Errors
+  ///
+  /// Returns `Err` if the input does not have a [`StardustDID`] compliant method id.
+  // TODO: Is it correct to also validate the network here? The current IOTA DID method does NOT do that.
+  pub fn check_method_id<D: DID>(did: &D) -> Result<()> {
+    let (network, tag) = Self::denormalized_components(did.method_id());
+    NetworkName::validate_network_name(network)
```
cycraig, line 162: That's probably an oversight in `IotaDID` that should be fixed. It doesn't affect internal usage because it's always constructed with a valid `NetworkName`. So the network check is technically redundant during construction, but not when called otherwise.

```suggestion
  pub fn check_method_id<D: DID>(did: &D) -> Result<()> {
    let (network, tag) = Self::denormalized_components(did.method_id());
    NetworkName::validate_network_name(network)
```

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,927 @@
...
+    let new_method_id: String = format!("{}:{}", name, self.tag());
+    // unwrap is fine as we are only replacing the network
+    self.0.set_method_id(new_method_id).unwrap();
+  }
+
+  /// Replace the network name of this [`StardustDID`] leaving all other segments (did, method, tag) intact.  
+  pub fn with_network_name(mut self, name: NetworkName) -> Self {
+    self.set_network_name(name);
+    self
+  }
```
cycraig, line 214: `IotaDID` and `StardustDID` are not meant to be mutable in general, and I can't think of a use-case where changing the network of an existing IOTA DID would be necessary.

Leaving `with_network_name` might be fine.

```suggestion
  /// Replace the network name of this [`StardustDID`] leaving all other segments (did, method, tag) intact.  
  pub fn with_network_name(mut self, name: NetworkName) -> Self {
    self.set_network_name(name);
    self
  }
```

identity_stardust/src/network/network_name.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Network name compliant with the [`crate::StardustDID`] method specification.
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct NetworkName(Cow<'static, str>);
+
+impl NetworkName {
+  pub(crate) const MAX_LENGTH: usize = 6;
+  pub(crate) const DEFAULT_STR: &'static str = "main";
+  // Names of networks maintained by the IF that the stardust UTXO based DID method should support
+  const IOTA_NETWORK_NAMES: [&'static str; 4] = [Self::DEFAULT_STR, "dev", "smr", "rms"];
```
cycraig, line 27: This isn't really a concern of `NetworkName`, but whatever uses it. These names have to be mapped to endpoints to be useful.

```suggestion
```

identity_stardust/src/network/network_name.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::error::Error;
+use crate::error::Result;
+
+/// Network name compliant with the [`crate::StardustDID`] method specification.
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct NetworkName(Cow<'static, str>);
+
+impl NetworkName {
+  pub(crate) const MAX_LENGTH: usize = 6;
```
cycraig, line 24: Should we make `MAX_LENGTH` public? Might be useful?

identity_stardust/src/network/network_name.rs
```diff
@@ -0,0 +1,147 @@
...
+      return Err(Error::InvalidNetworkName);
+    }
+
+    Ok(())
+  }
+
+  /// Creates a [`NetworkName`] representing the main network.  
+  pub const fn main() -> Self {
+    Self(Cow::Borrowed(Self::DEFAULT_STR))
+  }
```
cycraig, line 63: Again, probably shouldn't be a function.

```suggestion
```

identity_stardust/src/network/network_name.rs
olivereanderson: Was mainly meant as a way to optimize the `validate_network_name` function, but I never measured the impact of doing this so it can be removed for now.

identity_stardust/Cargo.toml
```diff
@@ -42,6 +44,8 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
+# Brings `AliasId` from iota-client into the public API of this crate.
+alias-id = []
```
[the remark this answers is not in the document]
olivereanderson: Turns I misunderstood your earlier comment regarding feature gating `from_alias_id`. I thought avoiding a dependency on `iota-client` in internal code would be impossible and therefore implemented it in this way (so users could opt in to getting the `AliasId` type in the public API). Given that it is not very hard to convert `AliasId` to an array of 32 bytes (as you pointed out), I think it can be fine to remove all usage of `AliasId` from this PR (except from the tests).

identity_stardust/src/did/stardust_did.rs
olivereanderson: Opened a new issue to fix the bug in `IotaDID`.

olivereanderson:
@cycraig. I should have addressed all of your last comments now. Unfortunately I had to move the `StardustDIDUrl` type alias introduced in the `did` module into the tests module for now in order to avoid rustdoc warnings. The proper fix is of course to make the `document` module use the types introduced here, but I was not able to do that quickly now so I am wondering whether we can fix that in a separate PR?

PhilippGackstatter (approved):
Looks good to me. Just minor comments and nitpicks.

identity_stardust/src/network/network_name.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// Network name compliant with the [`crate::StardustDID`] method specification.
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct NetworkName(pub(crate) Cow<'static, str>);
+
+impl NetworkName {
+  pub(crate) const MAX_LENGTH: usize = 6;
+  pub(crate) const DEFAULT_STR: &'static str = "main";
+  // Names of networks maintained by the IF that the stardust UTXO based DID method should support
```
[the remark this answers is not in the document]
olivereanderson: Ah it was commenting something that was removed, but forgot to remove the comment apparently. Thanks!

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,919 @@
...
+        StardustDID::parse(format!("did:{}:test:foo:{}", StardustDID::METHOD, valid_alias_id)),
+        Err(DIDError::InvalidMethodId)
+      ));
+    };
+
+    execute_assertions(INITIAL_ALIAS_ID_STR);
+    execute_assertions(VALID_ALIAS_ID_STR);
+  }
+
+  #[cfg(feature = "alias-id")]
```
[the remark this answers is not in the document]
olivereanderson: Yes. Thanks!

identity_stardust/src/did/stardust_did.rs
```diff
@@ -0,0 +1,903 @@
...
+  pub fn check_network<D: DID>(did: &D) -> Result<()> {
+    let network_name = Self::denormalized_components(did.method_id()).0;
+    NetworkName::validate_network_name(network_name).map_err(|_| DIDError::Other("invalid network name"))
+  }
+
+  /// Checks if the given `DID` is syntactically valid according to the [`StardustDID`] method specification.
+  ///
+  ///
+  ///  This function does NOT check whether the `did-method-specific-id` corresponds to a Blake2b-256 hash of an
+  ///  `OutputID`, despite this being a requirement for the [`StardustDID`] method specification.
```
cycraig, line 144: This is misleading, the `check_tag_str` function does ensure the tag is well-formed (i.e. 'syntactically valid'), just not whether it is an existing `AliasId`.

```suggestion
```

identity_stardust/src/network/network_name.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::error::Result;
+
+/// Network name compliant with the [`crate::StardustDID`] method specification.
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct NetworkName(pub(crate) Cow<'static, str>);
+
+impl NetworkName {
+  pub(crate) const MAX_LENGTH: usize = 6;
+  pub(crate) const DEFAULT_STR: &'static str = "main";
```
cycraig, line 25: This should be removed, it's a concern of `StardustDID` and can be accessed from `StardustDID::DEFAULT_NETWORK`.

identity_stardust/src/network/network_name.rs
```diff
@@ -0,0 +1,124 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::error::Error;
+use crate::error::Result;
+
+/// Network name compliant with the [`crate::StardustDID`] method specification.
+#[derive(Clone, Hash, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+#[repr(transparent)]
+pub struct NetworkName(pub(crate) Cow<'static, str>);
```
cycraig, line 21: This should not be `pub(crate)`, that allows bypassing its invariants elsewhere in the code.

```suggestion
pub struct NetworkName(Cow<'static, str>);
```

cycraig (approved):

identity_stardust/src/did/stardust_did.rs
olivereanderson: Should I reword this to `This function does NOT check whether the `did-method-specific-id` corresponds to an existing `AliasId` or would it be better to simply remove this text altogether?

identity_stardust/src/network/network_name.rs
olivereanderson: If it is not accessible in the `stardust_did` module it will not be possible to construct `const StardustDID::DEFAULT_NETWORK : NetworkName` (unless we reintroduce the `const fn main_network() -> Self` function).

identity_stardust/src/network/network_name.rs
olivereanderson: The `MAX_LENGTH` constant is used in `Self::validate_network_name` so I don't see how that can be moved elsewhere.  I don't quite understand what we want `NetworkName` to be. Given your comments about this so far I am wondering whether it would be better to make this generic over the DID method?

cycraig: This comment was on `DEFAULT_STR`, a previous comment was about making `MAX_LENGTH` a public associated constant.

identity_stardust/src/network/network_name.rs
cycraig: We can still expose static instances for well-known networks like `dev` and `main`, possibly using `oncecell` if necessary.

Issue #950, opened 2022-07-19
[Task] Implement Client for Stardust


PhilippGackstatter:
## Description

Implement an interface or wrapper for iota.rs.

Given the increased complexity of the client with the stardust, it might make more sense to implement an interface rather than wrap and re-expose many of its options. We also have to consider Wasm compatibility, either through compilation of iota.rs to Wasm or by binding iota.js.

## Motivation

Ease publishing and resolving of DID Documents in alias outputs.

## Resources

part of #908

## To-do list
Create a task-specific to-do list. Please link PRs that match the TODO list item behind the item after it has been submitted.

- [ ] Decide on the approach for interface or wrapper.
- [ ] Implement interface/wrapper.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #963, opened 2022-07-26
Add Stardust Client Extension Trait

PhilippGackstatter:
# Description of change

Adds an extension trait for the `Client` that is supposed to ease publication and resolution of DID documents contianed in Alias Outputs.

## Added

- `StardustClientExt::new_did`: Create a new alias output given a document and some other parameters.
- `StardustClientExt::update_did`: Resolve and update an alias output with a given document and return it.
- `StardustClientExt::publish_did_output`: Publish an `AliasOutput` (e.g. one obtained from `new_did` or `update_did`).
- `StardustClientExt::deactivate_did_output`: Deactivate a DID (recoverable operation).
- `StardustClientExt::delete_did_output`: Permanently delete an alias output.
- `StardustClientExt::resolve_did`: Resolve a DID Document from an alias output.
- `StardustClientExt::resolve_alias_output`: Resolve an alias output from a DID.
- `StardustClientExt::network_name`: Get the name of the network (BECH32 HRP) the client is connected to.

Specialized some error types.

Note: Wasm bindings for the client extension will be done in a separate PR.

#950

## Type of change

- [ ] Documentation Fix

Tests and examples were added.

## Change checklist


identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,459 @@
...
+          documents.push(StardustDocument::unpack(&did, alias_output.state_metadata())?);
+        }
+      }
+    }
+
+    Ok(documents)
+  }
+
+  /// Returns all alias ids of the Alias Outputs contained in the payload's transaction, if any.
+  fn alias_ids_from_payload(payload: &Payload) -> Result<Vec<AliasId>> {
```
cycraig, line 107: I think this is unused, remove?

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    assert_eq!(document, resolved);
+  }
+
+  async fn publish_update(
+    client: &Client,
+    document: StardustDocument,
+    secret_manager: &SecretManager,
+  ) -> crate::error::Result<Block> {
+    let rent_structure: RentStructure = client.get_rent_structure().await.map_err(Error::ClientError)?;
```
cycraig, line 366: If we're managing updates at this level of abstraction, it might be convenient to allow specifying how to handle the `amount`?

E.g. an enum like

```rust
enum StorageAmount {
  Amount(u64), // try publish with the specified amount
  Minimum, // publish with the minimum deposit required
  #[default]
  Same/Previous, // do not alter the amount of tokens.
}
```

cycraig (commented):
Looking good so far. Some ideas in the comments below, mostly about whether to make the `Client` more or less convenient.

Some of the comments are scattered, apologies.

Aside: do we need a naming convention to distinguish the trait methods from the rest of the client? E.g. `resolve_did`, `publish_new_did`, or `did_resolve`, `did_publish_new`, or something?

identity_stardust/examples/create_did.rs
```diff
@@ -114,126 +88,83 @@ async fn main() -> anyhow::Result<()> {
...
-    // Set the updated DID Document.
-    .with_state_metadata(updated_document.pack()?)
-    // State controller updates must increment the state index.
-    .with_state_index(alias_output.state_index() + 1)
-    .finish_output()?;
 
-  println!("Updated output: {updated_alias_output:?}");
+  // Publish the output and get the published document.
+  let block: Block = client.publish_outputs(&secret_manager, vec![output]).await?;
+  let documents: Vec<StardustDocument> = client.documents_from_block(&block).await?;
```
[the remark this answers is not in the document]
PhilippGackstatter: The main thought was that publishing multiple identities together would not be possible with the current API if `publish_output` would just take a single output. And in the account we might want to support that. On the flipside, extracting a single document form a vector is easily possible.
However, I agree with your other comment that building the alias output is harder than publishing so if we go down the `let alias_output = client.new_identity(&doc)` route, publishing multiple outputs in one transaction would be easily possible, since we wouldn't handle publishing.

identity_stardust/examples/create_did.rs
```diff
@@ -28,79 +35,46 @@ use identity_stardust::StardustDID;
...
-  let mnemonic = "veteran provide abstract express quick another fee dragon trend extend cotton tail dog truly angle napkin lunch dinosaur shrimp odor gain bag media mountain";
-  println!("Mnemonic: {}", mnemonic);
-  let secret_manager = SecretManager::Mnemonic(MnemonicSecretManager::try_from_mnemonic(mnemonic)?);
-
-  // Create a client instance.
-  let client = Client::builder()
-    .with_node(endpoint)?
-    .with_node_sync_disabled()
-    .finish()?;
+pub async fn run() -> anyhow::Result<(Client, Address, SecretManager, StardustDocument)> {
```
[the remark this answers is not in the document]
PhilippGackstatter: Agreed. The only issue is that then we can't write `mod 0_create` in the other examples to depend on the create example, since identifiers in Rust cannot start with digits.

identity_stardust/examples/create_did.rs
```diff
@@ -28,79 +35,46 @@ use identity_stardust::StardustDID;
...
-// 1) Alias Id is inferred from the block, so we have to use a placeholder DID for creation.
-// 2) Cannot get an Output Id back from an Alias Id (hash of Output Id), need to use Indexer API.
-// 3) The Output response from the Indexer is an Output, not a Block, so cannot infer Alias ID from it (fine since we
-// use the ID to retrieve the Output in the first place).    The OutputDto conversion is annoying too.
-// 4) The pieces needed to publish an update are fragmented (Output ID for input, amount, document), bit annoying to
-// reconstruct.    Use a holder struct like Holder { AliasOutput, StardustDocument } with convenience functions?
-// 5) Inferred fields such as the controller and governor need to reflect in the (JSON) Document but excluded from the
-// StardustDocument serialization when published.    Handle with a separate `pack` function like before?
+static ENDPOINT: &str = "https://api.testnet.shimmer.network/";
+static FAUCET_URL: &str = "https://faucet.testnet.shimmer.network/api/enqueue";
```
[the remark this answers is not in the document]
PhilippGackstatter: I guess we want to get away from predefining node endpoints, at least on the main and shimmer network (i.e. not test networks), so probably not?
I'm still in favor of retrieving the HRP at runtime as already partially done in this PR.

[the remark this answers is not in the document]
cycraig: > I guess we want to get away from predefining node endpoints

Do we? I think at least having them hardcoded in the library for convenience (whether in a `Network` or just as constants) would be useful.
And again the `HRP` can conflict between networks/chains, so I don't think it's sufficient as a network identifier.

identity_stardust/src/client/client_ext.rs
PhilippGackstatter: I've added this method now:

```rust
/// Resolves the alias output associated to the DID in `document` and updates it with that document.
///
/// Returns the updated alias output for further customization and publication.
///
/// This method does not modify the on-ledger state.
async fn update_did(&self, document: StardustDocument, rent_structure: RentStructure)
-> Result<AliasOutput> { ... }
```

Do you think it's still necessary then? This method uses `with_minimum_storage_deposit` by default, but users can set the amount they want explicitly afterwards. This should overwrite the storage deposit I imagine:

```rust
let alias_output = client.update_did(...).await?;
AliasOutputBuilder::from(&alias_output).with_amount(1_000_000)?.finish()?;
```

This enables the functionality of your `Amount` variant, but not the `Same/Previous` variants, since the resolved output is hidden from the user so they cannot access the previous amount.

We could alternatively not use `with_minimum_storage_deposit` in `update_did`, so the deposit stays the same, and a user can use that method or `with_amount` to configure the functionalities you suggested. However, I don't think that's what the default should be, since that would oftentimes not cover the required deposit.

identity_stardust/examples/create_did.rs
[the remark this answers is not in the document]
PhilippGackstatter: > Do we?

I'd lean towards it, yes, to encourage people to run their own nodes and not use IF nodes.

> And again the HRP can conflict between networks/chains

Yeah, although it only seems to be a problem for the test networks not for main or shimmer net. So I'm not sure how much of a problem it really is in practice. E.g. when you want to resolve and `rms` DID you don't know which network it is so it's unclear what endpoint to use. Again, though, only an issue in test scenarios from my understanding.

We can alternatively use the `network_id` of the network (a `u64`) which should be unique, and hardcode mappings to our own names like `main`, `smr`, `smr_testnet`, whatever. Then it'll be unambiguous what network to resolve a `did:iota:smr:...` on.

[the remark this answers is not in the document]
cycraig: > I'd lean towards it, yes, to encourage people to run their own nodes and not use IF nodes.

What about users that just want to resolve identities?

> it only seems to be a problem for the test networks not for main or shimmer net

It seems like a problem for anyone using a private Tangle too, or forks, or potentially shards, or whatever else could be done in the future out of our control.

> We can alternatively use the `network_id` of the network

How stable is that, does it change between network updates? Will it change after Stardust hits the IOTA Mainnet?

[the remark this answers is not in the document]
PhilippGackstatter: > What about users that just want to resolve identities?

I honestly don't know if we want to encourage resolvers to just trust IF nodes. 

> How stable is that

Not as stable as I hoped:

> Network identifier. This field denotes whether the message was meant for mainnet, testnet, or a private net. It also marks what protocol rules apply to the message. Usually, it will be set to the first 8 bytes of the BLAKE2b-256 hash of the concatenation of the network type and the protocol version string.

https://github.com/iotaledger/tips/blob/f4e1daa2b830695a848f437de30a8ddfe25f4f45/tips/TIP-0006/tip-0006.md#serialized-layout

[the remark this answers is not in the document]
cycraig: > I honestly don't know if we want to encourage resolvers to just trust IF nodes.

Not maintaining a default list of endpoints just seems like a big barrier to entry for anyone looking to start out using IOTA Identity. Even regular IOTA users might only be familiar with what Firefly handles automatically for them.

> Not as stable as I hoped:

Unfortunate, thanks for looking up the information. It would be a problem if that network identifier changes and breaks backwards compatibility arbitrarily, so maybe not the way forward.

Might need to look at what other DID methods do again, I know some use explicit network identifiers but we should see.

identity_stardust/src/client/client_ext.rs
cycraig: Getting the `RentStructure` is annoying and should be handled by the method, or be optional to re-use a previous one.

I concur that `with_minimum_storage_deposit` should not be the default because it affects tokens. The `Same/Previous` behaviour seems the most sensible default, but updating to the minimum required storage deposit is very common and should be supported too. Again, it's just a case of how convenient we want the `Client` to be compared to the `Account`.

PhilippGackstatter: Sorry for being unclear. I actually meant `with_minimum_storage_deposit` should be the default, because that's usually what's needed for the state update to succeed. Whether it affects tokens or not doesn't change how the API is called, so not sure where you see the issue.

cycraig: > Whether it affects tokens or not doesn't change how the API is called, so not sure where you see the issue.

The default should be to NOT affect people's tokens, anything potentially "dangerous" should be explicitly opt-in. The function signature of "update_did" shows no indication that it might take more funds or remove funds from the output (and documentation is insufficient per previous design decisions), hence an explicit parameter is preferred.

identity_stardust/examples/create_did.rs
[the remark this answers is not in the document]
PhilippGackstatter: > Not maintaining a default list of endpoints just seems like a big barrier to entry

That is fair. I would leave adding a `Network` type for another PR, though.

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,484 @@
...
+      .map_err(Error::AliasOutputBuildError)
+  }
+
+  /// Resolves the alias output associated to the DID in `document` and updates it with that document.
+  ///
+  /// Returns the updated alias output for further customization and publication. The storage deposit
+  /// on the output is unchanged. If the size of the document increased, the amount has to be increased.
+  ///
+  /// This method does not modify the on-ledger state.
+  async fn update_did(&self, document: StardustDocument) -> Result<AliasOutput> {
```
cycraig, line 85:
```suggestion
  /// Returns the updated alias output for further customization and publication. The storage deposit
  /// on the output is unchanged. If the size of the document increased, the amount must be increased manually.
  ///
  /// NOTE: this does *not* publish the updated Alias Output. See [publish_did](StardustClientExt::publish_did).
  async fn update_did(&self, document: StardustDocument) -> Result<AliasOutput> {
```

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    alias_output_builder.finish().map_err(Error::AliasOutputBuildError)
+  }
+
+  /// Publish the given `alias_outputs` with the provided `secret_manager`
+  /// and returns the block they were published in.
+  ///
+  /// This method modifies the on-ledger state.
+  async fn publish_did(&self, secret_manager: &SecretManager, alias_output: AliasOutput) -> Result<StardustDocument> {
```
cycraig, line 103:
```suggestion
  /// Publish the given `alias_output` with the provided `secret_manager`
  /// and returns the block they were published in.
  ///
  /// This method modifies the on-ledger state.
  async fn publish_did(&self, secret_manager: &SecretManager, alias_output: AliasOutput) -> Result<StardustDocument> {
```

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,484 @@
...
+    )
+  }
+
+  /// Consume the Alias Output containing the given `did`, sending its tokens to a new Basic Output
+  /// unlockable by `address`.
+  ///
+  /// # WARNING
+  ///
+  /// This destroys the DID Document and the Alias Output and renders the DID permanently unrecoverable.
+  async fn delete_did(&self, secret_manager: &SecretManager, address: Address, did: &StardustDID) -> Result<()> {
```
cycraig, line 135:
```suggestion
  /// Consume the Alias Output containing the given `did`, sending its tokens to a new Basic Output
  /// unlockable by `address`.
  ///
  /// Note that only the governor of an Alias Output is allowed to destroy it.
  ///
  /// # WARNING
  ///
  /// This destroys the DID Document and the Alias Output and renders the DID permanently unrecoverable.
  async fn delete_did(&self, secret_manager: &SecretManager, address: Address, did: &StardustDID) -> Result<()> {
```

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    Ok(())
+  }
+
+  /// Resolve a [`StardustDID`] to a [`StardustDocument`].
+  ///
+  /// # Errors
+  ///
+  /// Returns an [`NotFound`](iota_client::Error::NotFound) if the associated alias output wasn't found.
+  async fn resolve(&self, did: &StardustDID) -> Result<StardustDocument> {
```
cycraig, line 160: This should be `resolve_did`?

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,484 @@
...
+use iota_client::Client;
+
+use crate::error::Result;
+use crate::Error;
+use crate::NetworkName;
+use crate::StardustDID;
+use crate::StardustDocument;
+
+/// An extension trait for a [`Client`] that provides helper functions publication and resolution of
+/// DID documents in Alias Outputs.
```
olivereanderson, line 37: Perhaps it should be mentioned that this trait is mainly meant to be used rather than implemented?

identity_stardust/examples/create_did.rs
```diff
@@ -1,239 +1,130 @@
...
+  let client: Client = Client::builder().with_primary_node(ENDPOINT, None)?.finish()?;
+
+  // Get the BECH32 HRP identifier of the network.
+  let network_bech32_hrp: NetworkName = client
+    .get_network_info()
+    .await?
+    .bech32_hrp
+    .ok_or(Error::InvalidNetworkName)
+    .and_then(NetworkName::try_from)?;
+
```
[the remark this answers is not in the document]
PhilippGackstatter: With the intention of it being temporary? I guess we won't need it anymore if or when we fix the network name situation. I'll add it for now, though.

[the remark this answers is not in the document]
cycraig: It doesn't seem like it will be temporary.

identity_stardust/examples/resolve_did.rs
```diff
@@ -0,0 +1,47 @@
...
+      &client,
+      "rms1qrjwn7gc9fu88amxssxnkzqkxn80sfc29lwxnm0l2690z8kgd6yeqqatfd8"
+    )
+    .await
+  );
+
+  Ok(())
+}
+
+async fn get_address_balance(client: &Client, address: &str) -> u64 {
```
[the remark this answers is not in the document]
PhilippGackstatter: Committed something that wasn't supposed to be committed, I'll update.

identity_stardust/examples/create_did.rs
```diff
@@ -1,239 +1,130 @@
...
+    .data(MethodData::new_multibase(b"#key-1"))
+    .build()?;
+
+  document.insert_method(method, MethodScope::VerificationMethod)?;
+
+  // Get an address with funds from the testnet faucet.
+  let (address, secret_manager): (Address, SecretManager) = get_address_with_funds(&client).await?;
+
+  // Construct an alias output containing the `document` and whose state and governance is controlled by `address`.
+  let alias_output: AliasOutput = client.new_did(address, document, None).await?;
```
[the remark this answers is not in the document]
PhilippGackstatter: I tried to use "Alias Output" everywhere but apparently failed to do so consistently. If that's preferred, I'll make it consistent.

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,484 @@
...
+#[async_trait::async_trait]
+pub trait StardustClientExt: Sync {
```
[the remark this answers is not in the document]
PhilippGackstatter: iota.rs can still be compiled to Wasm and there are no plans to drop Wasm support itself (they only stopped doing iota.rs Wasm bindings themselves), so the idea was that this trait should "just work" for Wasm as long as we get a `Client` instance. Now that you mention it, the `Sync` requirement here might be a problem...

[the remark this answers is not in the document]
cycraig: If we don't use `iota.js` in the bindings then we need to expose most of what `iota-client` does anyway, so we may as well revert to wrapping it. Including everything to do with the `AliasOutput` (for setting the amount, controllers etc. which `StardustClientExt` does not do) in order to port it to the Wasm bindings ourselves.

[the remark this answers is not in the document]
PhilippGackstatter: We probably want to support iota.js as well as iota.rs's node.js bindings, since the latter comes with Stronghold support. To support code-sharing for the Wasm bindings, We can consider reorganizing the current code. The current trait has these methods:

```rust
#[async_trait::async_trait]
pub trait StardustClientExt: Sync {
  async fn new_did(
    &self,
    address: Address,
    document: StardustDocument,
    rent_structure: Option<RentStructure>,
  ) -> Result<AliasOutput> { ... }

  async fn update_did(&self, document: StardustDocument) -> Result<AliasOutput> { ... }

  async fn deactivate_did_output(&self, secret_manager: &SecretManager, did: &StardustDID) -> Result<()> { ... }

  async fn publish_did_output(
    &self,
    secret_manager: &SecretManager,
    alias_output: AliasOutput,
  ) -> Result<StardustDocument> { ... }

  async fn delete_did_output(&self, secret_manager: &SecretManager, address: Address, did: &StardustDID) -> Result<()> { ... }

  async fn resolve_did(&self, did: &StardustDID) -> Result<StardustDocument> { ... }

  async fn resolve_alias_output(&self, did: &StardustDID) -> Result<AliasOutput> { ... }

  async fn network_name(&self) -> Result<NetworkName> { ... }
}
```

We can split it into two traits, based on whether the method needs a `SecretManager` or not, since it is not serializable or otherwise portable across FFI (Stronghold cannot be compiled to Wasm).

```rust
#[async_trait::async_trait]
pub trait StardustClientExtCrud: StardustIdentityClient + Sync {
  async fn new_did(
    &self,
    address: Address,
    document: StardustDocument,
    rent_structure: Option<RentStructure>,
  ) -> Result<AliasOutput> { ... }

  async fn update_did(&self, document: StardustDocument) -> Result<AliasOutput> { ... }

  async fn resolve_did(&self, did: &StardustDID) -> Result<StardustDocument> { ... }

  async fn resolve_alias_output(&self, did: &StardustDID) -> Result<AliasOutput> { ... }

  async fn network_name(&self) -> Result<NetworkName> { ... }
}

#[async_trait::async_trait]
pub trait StardustIdentityClient: Sync {
  async fn get_network_hrp(&self) -> Result<String>;

  async fn resolve_alias_output(&self, did: &StardustDID) -> Result<(AliasId, OutputId, AliasOutput)>;

  async fn get_rent_structure(&self) -> Result<RentStructure>;

  /* more helper methods that StardustClientExtCrud requires */
}

#[async_trait::async_trait]
impl StardustIdentityClient for Client { ... }

impl<T> StardustClientExtCrud for T where T: StardustIdentityClient {}
```

And the other part:

```rust
// Contains methods that need a SecretManager.
pub trait StardustClientExtPublishing: Sync {
  async fn deactivate_did_output(&self, secret_manager: &SecretManager, did: &StardustDID) -> Result<()> { ... }

  async fn publish_did_output(
    &self,
    secret_manager: &SecretManager,
    alias_output: AliasOutput,
  ) -> Result<StardustDocument> { ... }

  async fn delete_did_output(&self, secret_manager: &SecretManager, address: Address, did: &StardustDID) -> Result<()> { ... }
}
```

(Names temporary).

We should be able to implement `StardustIdentityClient` for `IotaJsClient` and/or `IotaRsNodeJsBindingClient` since all the parameters of the trait's methods are serializable, so we can re-use the code from `StardustClientExtCrud`. `StardustIdentityClient` is meant to be an abstraction over the necessery networking functionality the extension trait needs. `StardustClientExtPublishing` is only meant as an extension trait for the Rust API. Its functionality will have to be reimplemented from scratch for the Wasm binding, due to the lack of `SecretManager` portability.

For convenience in Rust we can implement a trait that looks like the initial `StardustClientExt` and just delegates to both traits.

Does that make sense?

In the interest of getting this PR merged soon, I would not do the refactoring in this one, though.

identity_stardust/examples/ex4_delete_did.rs
```diff
@@ -0,0 +1,38 @@
...
+  // This operation is *not* reversible.
+  // Deletion can only be done by the governor of the Alias Output.
+  client
+    .delete_did_output(&secret_manager, address, document.id())
+    .await?;
+
+  // Wait for the node to index the new state.
+  tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+
+  // Attempting to resolve a delete DID results in a `NotFound` error.
```
olivereanderson, line 29:
```suggestion
  // Attempting to resolve a deleted DID results in a `NotFound` error.
```

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,603 @@
...
+
+  /// Create a DID with a new Alias Output containing the given `document`.
+  ///
+  /// The `address` will be set as the state controller and governor unlock conditions.
+  /// The minimum required token deposit amount will be set according to the given
+  /// `rent_structure`, which will be fetched from the node if not provided.
+  /// The returned Alias Output can be further customized before publication, if desired.
+  ///
+  /// NOTE: this does *not* publish the Alias Output. See [`publish_did_output`](StardustClientExt::publish_did_output).
+  async fn new_did(
```
olivereanderson, line 53: Consider adding an `#Errors` section to the documentation explaining how this method may fail and what error variants to expect in these scenarios.

identity_stardust/examples/ex3_deactivate_did.rs
```diff
@@ -0,0 +1,32 @@
...
+/// Demonstrate how to deactivate a DID in an Alias Output.
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+  // Create a new DID in an Alias Output for us to modify.
+  let (client, _, secret_manager, document): (Client, Address, SecretManager, StardustDocument) =
+    ex0_create_did::run().await?;
+
+  // Deactivate the DID by publishing an empty document.
+  // This process can be reversed since the Alias Output is not destroyed.
+  client.deactivate_did_output(&secret_manager, document.id()).await?;
```
cycraig, line 21: State that only the state controller can perform updates and deactivations, similar to the remark in `ex4_delete_did`.

identity_stardust/src/error.rs
```diff
@@ -1,31 +1,45 @@
...
   #[error("{0}")]
   InvalidDoc(#[from] identity_did::Error),
   #[cfg(feature = "iota-client")]
-  #[error("{0}")]
-  ClientError(#[from] iota_client::error::Error),
+  #[error("DID update failed")]
+  DIDUpdateError(#[source] iota_client::error::Error),
+  #[cfg(feature = "iota-client")]
+  #[error("DID resolution failed")]
+  ResolutionError(#[source] iota_client::error::Error),
```
cycraig, line 22: Inconsistent use of the `DID` prefix.

identity_stardust/examples/ex3_deactivate_did.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Demonstrate how to deactivate a DID in an Alias Output.
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+  // Create a new DID in an Alias Output for us to modify.
+  let (client, _, secret_manager, document): (Client, Address, SecretManager, StardustDocument) =
+    ex0_create_did::run().await?;
+
+  // Deactivate the DID by publishing an empty document.
+  // This process can be reversed since the Alias Output is not destroyed.
```
cycraig, line 20: The example should also demonstrate how to re-publish the DID Document, "reactivating" the DID.

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,616 @@
...
+        .map_err(Error::DIDUpdateError)?
+    };
+
+    AliasOutputBuilder::new_with_minimum_storage_deposit(rent_structure, AliasId::null())
+      .map_err(Error::AliasOutputBuildError)?
+      .with_state_index(0)
+      .with_foundry_counter(0)
+      .with_state_metadata(document.pack()?)
+      .add_feature(Feature::Sender(SenderFeature::new(address)))
+      .add_immutable_feature(Feature::Issuer(IssuerFeature::new(address)))
```
cycraig, line 80: The issuer feature could be undesirable or seen as unnecessary. I suppose people could remove it themselves though, if possible.

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,616 @@
...
+    assert!(matches!(error, Error::ResolutionError(iota_client::Error::NotFound)));
+
+    let balance: u64 = get_address_balance(&client, &address_bech32).await;
+
+    assert_eq!(initial_balance, balance);
+  }
+
+  #[tokio::test]
+  async fn test_client_deactivate_and_reactivate() {
+    let client: Client = client();
```
cycraig, line 587: We usually don't include unit tests that perform the same actions as the examples, since they're mostly redundant and interacting with the Tangle is slow.

The Stardust examples should be hooked up in CI at this point I guess.

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,616 @@
...
+
+    for (index, output) in regular.outputs().iter().enumerate() {
+      if let Output::Alias(alias_output) = output {
+        let alias_id = if alias_output.alias_id().is_null() {
+          AliasId::from(
+            OutputId::new(
+              tx_payload.id(),
+              index.try_into().expect("the output count should not exceed u16"),
+            )
+            .expect("the index of any output in a RegularTransactionEssence should be a valid OutputIndex"),
```
cycraig, line 312: I can't imagine these invariants being broken but we should map them to errors in case the protocol changes and breaks them.

identity_stardust/src/client/client_ext.rs
```diff
@@ -0,0 +1,616 @@
...
+  /// `rent_structure`, which will be fetched from the node if not provided.
+  /// The returned Alias Output can be further customized before publication, if desired.
+  ///
+  /// NOTE: this does *not* publish the Alias Output. See [`publish_did_output`](StardustClientExt::publish_did_output).
+  ///
+  /// # Errors
+  ///
+  /// - Returns an [`Error::DIDUpdateError`] when retrieving the `RentStructure` fails.
+  /// - Returns an [`Error::AliasOutputBuildError`] when building the Alias Output fails.
+  async fn new_did(
```
cycraig, line 58: Both `new_did` and `update_did` return `AliasOutput`, should they be named `*_did_output` to be consistent with the other methods?

cycraig (approved):
Errors are getting worse but they can be addressed later.

olivereanderson (approved):
Looks good to me, 

Thanks for removing error variants that only give information about which crate the error originated from (`CoreError`, `ClientError` and `BlockError`).

identity_stardust/src/client/client_ext.rs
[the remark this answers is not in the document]
cycraig: Seems like a good approach. We will pursue this direction for the Wasm bindings once this PR is merged.

identity_stardust/src/client/client_ext.rs
PhilippGackstatter: Removed it.

According to TIP-18 it's primary use case is:

> [...] proving authenticity of NFTs. Whenever an NFT is minted as an NFT output, the creator (issuer) can fill the Issuer Feature with their address that they have to unlock in the transaction. Issuers then can publicly disclose their addresses to prove the authenticity of the NFT once it is in circulation.

PR #972, opened 2022-08-09
Update DID Method Specification

abdulmth:
# Description of change
Update the DID specs page in the wiki for the new Stardust method.

This is a first draft, please add your suggestions and comments. 

- [x] Documentation Fix


documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,29 +14,54 @@ keywords:
...
 
 ![Blockchain bottleneck](/img/blockchain-bottleneck.gif) | ![Tangle Bottleneck](/img/tangle-bottleneck.gif)
 
-For this method, the most important features of IOTA are: 
+## UTXO and Stardust
+
+The unspent transaction output ([UTXO](https://wiki.iota.org/IOTA-2.0-Research-Specifications/5.1UTXO)) model defines a ledger state where outputs are created by a transaction consuming outputs of previous transactions as inputs. Stardust introduced multiple output types including Basic Output for value transactions, NFT Output for storing NFTs on layer 1, and Alias Output mainly designed for smart contract support.
+
+All types of outputs must hold a minimum amount of IOTA coins in order to be stored on the ledger. For some output types that can hold data, for instance the Alias Output, the amount of IOTA coins held by the output must cover the byte cost of the data stored. This helps to control the ledger size from growing uncontrollably while guaranteeing the data is indefinitely stored on the ledger which is important for resolving DID Documents.
+This deposit is fully refundable and can be reclaimed when the output is destroyed.
```
cycraig, line 37:
```suggestion
All outputs must hold a minimum amount of coins to be stored on the ledger. For output types that can hold arbitrary data, for instance the Alias Output, the amount of coins held by the output must cover the byte cost of the data stored. This helps control the ledger size from growing uncontrollably while guaranteeing the data is stored on the ledger indefinitely, which is important for resolving DID Documents. This deposit is fully refundable and can be reclaimed when the output is destroyed.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,29 +14,54 @@ keywords:
...
+![Outputs](/img/utxo.png)
+
+**todo outdated image**
+
+### Alias Output
+
+The [Alias Output](https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#alias-output) is a specific implementation of the UTXO state machine. It includes specific properties including:
+
+* **Amount**: the amount of IOTA coins held by the output.
+* **Alias ID**: 32 Byte Array, a unique identifier of the alias, which is the BLAKE2b-256 hash
```
cycraig, line 50:
```suggestion
* **Alias ID**: 32 byte array, a unique identifier of the alias, which is the BLAKE2b-256 hash
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,29 +14,54 @@ keywords:
...
 
-* The lack of fees, requiring no cryptocurrency tokens to be owned in order to submit a message to the DLT.
-* The DLT has a public and permissionless network which runs the IOTA cryptocurrency.
-* Pure data messages are possible to be stored immutably. 
-* Few nodes store the entire Tangle, requiring additional logic to prove the immutability of data. 
+All outputs include an `Unlock Conditions` property. This feature defines how the output can be unlocked and spent. For the Alias Output there's two types of controllers that can be set as Unlock Conditions, state and governor controllers. Each of these controllers can be either an Ed25519 Address, Alias Address or an NFT Address. An Alias Output can have at most one of each controller.
+
+The state controller is responsible for a state transition. It is identified by an incremented `State Index` and can change the fields `IOTA Amount`, `State Index`, `State Metadata` among other properties.
```
cycraig, line 62:
```suggestion
The state controller is responsible for a state transition. It is identified by an incremented `State Index` and can change the fields `Amount`, `State Index`, `State Metadata` among other properties.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,29 +14,54 @@ keywords:
...
+
+The governor controller, on the other hand, is responsible for a governance transition indicated by an unchanged `State Index`. A governance transition can change the Addresses of state and governor controllers. It also allows a transition to an empty state resulting in destroying the Alias Output.
```
cycraig, line 64:
```suggestion
The governor, on the other hand, is responsible for a governance transition indicated by an unchanged `State Index`. A governance transition can change the addresses of the state controller and governor. It also allows destroying the Alias Output by consuming it.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,29 +14,54 @@ keywords:
...
+    * State Controller Address Unlock Condition
+    * Governor Address Unlock Condition
+
+Consuming an Alias Output in a transaction means that the alias is transformed into the next state. The current state is defined as the consumed alias output, while the next state is defined as the **alias output with the same explicit `AliasID` on the output side**. There are two types of transitions: `state transition` and `governance transition`.
 
-* The lack of fees, requiring no cryptocurrency tokens to be owned in order to submit a message to the DLT.
-* The DLT has a public and permissionless network which runs the IOTA cryptocurrency.
-* Pure data messages are possible to be stored immutably. 
-* Few nodes store the entire Tangle, requiring additional logic to prove the immutability of data. 
+All outputs include an `Unlock Conditions` property. This feature defines how the output can be unlocked and spent. For the Alias Output there's two types of controllers that can be set as Unlock Conditions, state and governor controllers. Each of these controllers can be either an Ed25519 Address, Alias Address or an NFT Address. An Alias Output can have at most one of each controller.
```
cycraig, line 60: This is frequent: it's just "governor".

```suggestion
All outputs include an `Unlock Conditions` property. This feature defines how the output can be unlocked and spent. The Alias Output supports two types of controllers that can be set as unlock conditions: the state controller and governor. Each of these can be either an Ed25519 Address, Alias Address or an NFT Address. An Alias Output can have at most one of each unlock condition.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
 
 ### IOTA-Network
 
+**TODO should be replaced by shimmer? **
```
cycraig, line 89: This should now be the network Bech32-HRP. E.g. "iota", "atoi", "smr", "rms".

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-The following steps MUST be taken to generate a valid tag:
-* Generate an asymmetric keypair using a supported verification method type.
-* Hash the public key using `BLAKE2b-256` then encode it using [Base58-BTC](https://tools.ietf.org/id/draft-msporny-base58-01.html).
-
-This public key MUST be embedded into the DID Document (see [CRUD: Create](#create)).
+The IOTA tag references an `Alias ID` identitfying the Alias Output where the DID Document is stored.
+The tag will not be known before the generation of the DID. It will be assigned when the Alias Output is created.
 
-## DID Messages
+### Anatomy of serialized document
```
cycraig, line 110:
```suggestion
### Anatomy of the State Metadata
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
 
-DID Documents associated with the `did:iota` method consist of a chain of data messages, called "DID messages", published to a Tangle. The Tangle has no understanding of DID messages and acts purely as an immutable database. The chain of DID messages and the resulting DID Document must therefore be validated on the client side. 
+In the `Metadata` of the Alias Output must be a JSON document containing two properties:
```
cycraig, line 112: This is incorrect, it's a byte-packed struct with header fields, like the DID Message. It's also `stateMetadata`, not `metadata`, which is a separate feature.

`DID<VERSION><ENCODING><PAYLOAD>`

This should conveyed by a Markdown table similar to the output specifications, e.g. https://hackmd.io/SULTqAt5Q7yQwWTghJtZCw#DID-Output

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-    "verificationMethod": "did:iota:GzXeqBXGCbuebiFtKo4JDNo6CmYmGbqxyh2fDVKadiBG#authentication",
-    "signatureValue": "3fLtv3KUU4T5bHNLprV3UQ2Te3bcRZ9uUYSFouEA7fmYthieV35NNLqbKUu8t2QmzYgnfp1KMzCqPzGNi3RjU822"
-}
 ```
 
-### Anatomy of the Diff object
+## Controllers
 
-The `diff` object MUST contain all the differences between the current and previous DID Document and its DID Document Metadata. The differentiation is formatted as a JSON object that includes the differences between the two DID Document objects and the two DID Document Metadata objects. Exact details of how this is generated will be added later.
+As of now, only one state and one governance controller can be set for an Alias Output. Multiple controller support can be done by future updates.
```
cycraig, line 143:
```suggestion
As of now, only one state controller and one governor can be set for an Alias Output. Support for multiple controllers may be possible depending on future updates of the protocol.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-        }
-      ]
-    },
-    "meta": {
-      "updated": "2022-01-21T09:50:28Z"
-    }
-  }
-}
-```
+A state controller can directly update the DID Document and the amount of IOTA coins held by the Output, but it can't destroy the output. A governor controller, one the other hand, can indirectly update the DID Document by updating the state controller. It also can destroy the output by having a governance transition to the empty state.
```
cycraig, line 145: Place this above the previous paragraph. Avoid saying "empty state" for destruction, it can cause confusion because that is how we deactivate DID Documents now, by publishing an empty `stateMetadata`.

```suggestion
A state controller can directly update the DID Document and the amount of coins held by the Alias Output, but it cannot destroy the output. A governor controller, one the other hand, can indirectly update the DID Document by updating the state controller. It can also destroy the output by performing a governance transition to the empty state.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-
-### Update
-
-In order to update a DID Document, either an Integration or a Diff DID message needs to be generated. It is RECOMMENDED to use only Integration DID messages if the DID Document is updated very infrequently and it is expected to never go beyond 100 updates in the lifetime of the DID. If that is not the case, it is RECOMMENDED to use as many Diff DID messages instead with a maximum of around 100 updates per Diff chain. 
+In order to read the latest DID document associated with a DID, the Alias Output of containing the DID Document
+must be queried using the `Alias ID`. This can be done using an [inx indexer](https://github.com/iotaledger/inx-indexer)
+which normal nodes usually include by default.
 
-#### Creating an Integration DID message
+### Updating DID Document
```
cycraig, line 163:
```suggestion
### Update
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-    },
-    "meta": {
-      "updated": "2022-01-21T09:50:28Z"
-    }
-  }
-}
-```
+A state controller can directly update the DID Document and the amount of IOTA coins held by the Output, but it can't destroy the output. A governor controller, one the other hand, can indirectly update the DID Document by updating the state controller. It also can destroy the output by having a governance transition to the empty state.
 
 ## CRUD Operations
```
cycraig, line 147: All these operations really need numbered steps, an implementer should be able to replicate the DID Method exactly by following them.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
+Create, Read, Update and Delete (CRUD) operations that change the DID Documents are to be submitted to an IOTA Tangle in order to be publicly available.
+They have to update or destroy the Alias Output including the DID document.
 
 ### Create
 
-To generate a new DID, the method described in [generation](#generation) must be followed. A basic DID Document must be created that includes the public key used in the DID creation process as a `verificationMethod` with a capability invocation verification relationship. This DID Document must be formatted as an Integration DID message, signed using the same keypair used to generate the tag, and published to an IOTA Tangle on the index generated out of the public key used in the DID creation process. 
+Creating a new DID is done by creating a new Alias Output containing a serialized document as described in [Anatomy of serialized document](#anatomy-of-serialized-document). This requires the fulfillment of all the
+Stardust requirements including the byte cost.
 
 ### Read
```
cycraig, line 157: This needs to detail how to obtain the Alias Id from a DID, unpack the state metadata and, for instance, replace the placeholder `did:0:0` with the actual DID.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +86,90 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-1. Create and serialize a `diff` JSON object that contains all the changes. 
-2. Set the `did` field to the DID value that this update applies to.
-3. Retrieve the IOTA MessageId from the previous Diff chain message, or Integration DID message if this message is the first in the Diff chain. In addition, retrieve a keypair for signing the DID Document.
-4. Set the `previousMessageId` field to the IOTA MessageId value.
-5. Create a `proof` object referencing the public key and signature suite from a verification method in the DID Document able to sign updates as per the [Signing Keys](#signing-keys) section.
-6. Create a cryptographic signature using the same keypair and add the result to the appropriate field(s) inside the `proof` object.
+Updating can be achieved by a state transition of the Alias Output. A state controller can update the
+`State Metadata` in the Alias Output which reflects in updating the DID Document.
 
 ### Delete
```
cycraig, line 168: Have two subsections under this section: `Deactivate` and `Destroy`.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +184,24 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
-
 ### Private Key Management
 
-All private keys or seeds used for the `did:iota` method should be equally well protected by the users. The signing key is especially important as it controls how keys are added or removed, providing full control over the identity. The IOTA Identity framework utilizes the [Stronghold project](https://github.com/iotaledger/stronghold.rs), a secure software implementation isolating digital secrets from exposure to hacks or leaks. Developers may choose to add other ways to manage the private keys in a different manner.  
+All private keys or seeds used for the `did:iota` method should be equally well protected by the users. The signing key is especially important as it controls how keys are added or removed, providing full control over the identity. The IOTA Identity framework utilizes the [Stronghold project](https://github.com/iotaledger/stronghold.rs), a secure software implementation isolating digital secrets from exposure to hacks or leaks. Developers may choose to add other ways to manage the private keys in a different manner.
```
cycraig, line 201: Remove all references to "signing key". We should refer to the keys for the state controller and governor now.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +184,24 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
-
-### Correlation Risks
-
-As with any DID method, identities can be linked if they are used too often and their usage somehow becomes public. IOTA provides a simple solution for this problem: as creating identities is completely feeless, it is recommended to make new identities on a per interaction basis (Pairwise DIDs). The IOTA Identity framework will provide support for this in the future. 
+The public IOTA Tangle networks are immutable networks. This means that once something is uploaded, it can never be completely removed. That directly conflicts with certain privacy laws such as GDPR, which have a 'right-to-be-forgotten' for Personal Identifiable Information (PII). As such, users should NEVER upload any PII to the Tangle, including inside DID Documents. The IOTA Identity framework allows Verifiable Credentials to be published to the Tangle directly, however this feature should only be utilized by Identity for Organisation and Identity for Things.
```
cycraig, line 207: This needs to be expanded a bit to mention how destroying an Alias Output does remove it from the current state of the ledger, but it may still be stored by permanodes or other parties recording historical ledger states.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +184,24 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
 
 ## Privacy Considerations
 
 ### Personal Identifiable Information
 
-The public IOTA Tangle networks are immutable networks. This means that once something is uploaded, it can never be completely removed. That directly conflicts with certain privacy laws such as GDPR, which have a 'right-to-be-forgotten' for Personal Identifiable Information (PII). As such, users should NEVER upload any PII to the Tangle, including inside DID Documents. The IOTA Identity framework allows Verifiable Credentials to be published to the Tangle directly, however this feature should only be utilized by Identity for Organisation and Identity for Things. 
-
-### Correlation Risks
-
-As with any DID method, identities can be linked if they are used too often and their usage somehow becomes public. IOTA provides a simple solution for this problem: as creating identities is completely feeless, it is recommended to make new identities on a per interaction basis (Pairwise DIDs). The IOTA Identity framework will provide support for this in the future.
```
cycraig, line 360: We should still mention such risks, they are even greater now since it can correlate your identity with your wallet.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +184,24 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
 
 The IOTA Identity framework also standardized certain `services` that are embedded in the DID Document. It is RECOMMENDED to implement these when implementing the `did:iota` method.
 
 Currently standardized `services`:
+
 * Nothing yet.
 
 ## Security Considerations
 
-The `did:iota` method is implemented on the [IOTA Tangle](https://iota.org), a public permissionless and feeless Distributed Ledger Technology (DLT), making it resistant against almost all censorship attack vectors. Up until the `Coordicide` update for the IOTA network, a reliability on the coordinator exists for resolving ordering conflicts. This has a minor censorship possibility, that, in the worst case, can prevent ordering conflicts from resolving to a DID. However, these can only be published by the owner of the private key and therefore does not constitute a usable attack vector. Lastly, a node may decide to censor DID messages locally, however a user can easily use another node.
```
cycraig, line 326: We can still mention the public, permissionless and censorship aspects here. And that individual nodes could potentially censor DIDs but again, any other node could be used.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +184,24 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
 
 Currently standardized `services`:
+
 * Nothing yet.
 
 ## Security Considerations
 
-The `did:iota` method is implemented on the [IOTA Tangle](https://iota.org), a public permissionless and feeless Distributed Ledger Technology (DLT), making it resistant against almost all censorship attack vectors. Up until the `Coordicide` update for the IOTA network, a reliability on the coordinator exists for resolving ordering conflicts. This has a minor censorship possibility, that, in the worst case, can prevent ordering conflicts from resolving to a DID. However, these can only be published by the owner of the private key and therefore does not constitute a usable attack vector. Lastly, a node may decide to censor DID messages locally, however a user can easily use another node. 
-
-Since DID messages are always to be signed and the 'chain of custody' of the signing key can be traced throughout the identity lifespan, replay, message insertion, deletion, modification and man-in-the-middle attacks are prevented.
```
cycraig, line 328: We really need a section on what aspects we are protected against by using the ledger. Notably how it deals with conflicts, replays, man-in-the-middle-attacks etc. The state index also provides a good linear history, if one has it.

We are not transparently cryptographically verifiable to clients anymore though (we rely on the protocol), so we might want to consider whether it's worth including some sort of signature on the DID Document payload as well.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +85,137 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
 When no IOTA network is specified, it is assumed that the DID is located on the `main` network. This means that the following DIDs will resolve to the same DID Document:
+
 ```
-did:iota:main:H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV
-did:iota:H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV
+did:iota:main:0xe4edef97da1257e83cbeb49159cfdd2da6ac971ac447f233f8439cf29376ebfe
+did:iota:0xe4edef97da1257e83cbeb49159cfdd2da6ac971ac447f233f8439cf29376ebfe
 ```
 
 ### IOTA-Tag
```
eike-hass, line 104: We need to come up with a different term. The did spec calls it "method specific-identifier". Other suggestions:
`identifier`
`iota-id`

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +85,137 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-### Anatomy of the Proof object
-
-Following the proof format in the [Verifiable Credential standard](https://www.w3.org/TR/vc-data-model/#proofs-signatures), at least one proof mechanism, and the details necessary to evaluate that, MUST be expressed for a DID Document uploaded to the Tangle. 
+Note that `did:0:0` is a placeholder of the actual DID.
 
-The proof object is an embedded proof that contains all information to be verifiable. It contains one or more cryptographic proofs that can be used to detect tampering and verify the authorship of a DID creation or update. It mostly follows LD-Proofs standard.
+## Controllers
 
-**Type**
+A state controller can directly update the DID Document and the amount of coins held by the Alias Output, but it cannot destroy the output. A governor, on the other hand, can indirectly update the DID Document by updating the state controller. It can also destroy the output by performing a governance transition without producing an Alias Output with the same `Alias ID`.
```
eike-hass, line 151:
```suggestion
A state controller can directly update the DID Document and the amount of coins held by the Alias Output, but it cannot destroy the output. A governor, on the other hand, can indirectly update the DID Document by updating the state controller. They can also destroy the output by performing a governance transition without producing an Alias Output with the same `Alias ID`.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -58,246 +85,137 @@ base-char = 1-9 A-H J-N P-Z a-k m-z
...
-```
+As of now, only one state controller and one governor can be set for an Alias Output. Support for multiple controllers may be possible depending on future updates of the protocol.
 
 ## CRUD Operations
 
-Create, Read, Update and Delete (CRUD) operations that change the DID Documents are to be submitted to an IOTA Tangle in order to be publicly available. They will either have to be a valid Int DID message or Diff DID message, submitted on the correct index of the identity. 
+Create, Read, Update and Delete (CRUD) operations that change the DID Documents are to be submitted to an IOTA Tangle in order to be publicly available. They have to update or destroy the Alias Output including the DID document.
 
-### Create
+**These operation require fund transfer to cover storage deposit. Transactions must be carefully done in order to avoid fund loss**
```
eike-hass, line 159: Elaborate on "carefully done"

documentation/docs/specs/did/iota_did_method_spec.md
cycraig: We use `tag` in the framework though. The `method-specific-identifier` includes the `network` (e.g. `main:0x123`, `dev:0x123`, `rms:0x123`, etc.), so tag distinguishes "just the end bit"?

eike-hass: Good point, since there is no obvious replacement, let's leave it as is.

eike-hass:
@abdulmth What about the remaining TODOs? Do you need help tackling them?

documentation/docs/specs/did/iota_did_method_spec.md
abdulmth: The state controller can't destroy the Alias Output AFAIK. Only the governor can.

eike-hass: You are absolutely right, this was meant as an attempt at more inclusive speech. Maybe we just go for `The governor can ...`.

PhilippGackstatter (commented):
Some comments below. Since this is a specification I'm erring on the side of being precise, so sorry for the nitpicks.

Haven't reviewed everything yet, will do the rest shortly.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,29 +14,50 @@ keywords:
...
+The unspent transaction output ([UTXO](https://wiki.iota.org/IOTA-2.0-Research-Specifications/5.1UTXO)) model defines a ledger state where outputs are created by a transaction consuming outputs of previous transactions as inputs. IOTA and Shimmer have several output types, the relevant ones for the IOTA DID Method are: Basic Outputs for value transactions, and Alias Outputs.
 
-Blockchain | Tangle
-:---------:|:---------:
+All outputs must hold a minimum amount of coins to be stored on the ledger. For output types that can hold arbitrary data, for instance the Alias Output, the amount of coins held by the output must cover the byte cost of the data stored. This helps control the ledger size from growing uncontrollably while guaranteeing the data is stored on the ledger indefinitely, which is important for resolving DID Documents. This deposit is fully refundable and can be reclaimed when the output is destroyed.
```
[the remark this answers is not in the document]
PhilippGackstatter: Wow, that's really weird. But sure, then let's match TIP-18.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -47,257 +68,146 @@ A DID that uses this method MUST begin with the following prefix: `did:iota`. Fo
...
 
-In order to create a valid Integration DID message, the following steps are to be performed:
-1. Create and serialize a `diff` JSON object that contains all the changes. 
-2. Set the `did` field to the DID value that this update applies to.
-3. Retrieve the IOTA MessageId from the previous Diff chain message, or Integration DID message if this message is the first in the Diff chain. In addition, retrieve a keypair for signing the DID Document.
-4. Set the `previousMessageId` field to the IOTA MessageId value.
-5. Create a `proof` object referencing the public key and signature suite from a verification method in the DID Document able to sign updates as per the [Signing Keys](#signing-keys) section.
-6. Create a cryptographic signature using the same keypair and add the result to the appropriate field(s) inside the `proof` object.
+#### Deactivate
+Temporarily deactivating a DID can be done by deleting the content of the `State Meadata` in the Alias Output. This will render the DID unresolvable.
```
PhilippGackstatter, line 201: It can still be resolved, but it will yield a deactivated document.

```suggestion
Temporarily deactivating a DID can be done by deleting the content of the `State Meadata` in the Alias Output.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -47,257 +68,146 @@ A DID that uses this method MUST begin with the following prefix: `did:iota`. Fo
...
+1. Create and new Alias Output with the same `Alias ID` as the current output with empty `State Metadata`.
```
PhilippGackstatter, line 202: "Create new" could be read as an output different from the current one, which is not accurate.

```suggestion
1. Set the Alias Output's `State Metadata` field to an empty byte array.
```

PhilippGackstatter (changes requested):
I think the CRUD section could be simplified a lot if we differentiated clearly between:

- the concerns of TIP-18,
- and the DID concerns: Transforming a DID Document into a State Metadata Document, the State Metadata headers, how to create a deactivated DID document when resolving an output and finding an empty state metadata.

Then we would have to be much less redundant with what TIP-18 already specifies in detail, e.g. how to create a valid Alias Output, what addresses can be set as the state controller, governor, how to create a valid state or governance transition, storage deposit, etc. Someone that wants to implement this, needs to understand the Alias Output part of TIP-18 anyway. Even if we duplicate some of the spec, we will never be as accurate as TIP-18, so the implementors will have to consult that TIP regardless. It also makes our life maintaining this spec correctly easier, because we can focus on how exactly we configure the alias output for DIDs.

For example, for the "Create" operation it seems sufficient to mention:
1. Create the content of the DID Document like verification methods, services, etc.
2. Create the `state metadata byte array` from the DID document as described in the [Anatomy of the State Metadata](#anatomy-of-the-state-metadata).
3. Create a new Alias Output and store the `state metadata byte array` in the output's `State Metadata` field.
4. Publish the output.

(where anatomy of the state metadata clearly mentions how the DID document to state metadata document transformation works, and vice versa).

All the concerns around storage deposit, what unlock conditions to set, etc. is not really our concern, so the above seems sufficient.

My review comments address the current state if we decide to keep the current approach, but I would prefer rewriting it to not redundantly mention what TIP-18 already specifies.

abdulmth:
@PhilippGackstatter I agree that we should abstract the Stardust details away from the specs. Let's keep it in this PR and I will add that to the specs improvements issue #974.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,290 +14,210 @@ keywords:
...
+   },
+   "metadata":{
+      "created":"2022-08-02T21:39:48Z",
+      "updated":"2022-08-02T21:39:48Z"
+   }
 }
 ```
 
-### Anatomy of the Diff object
+In a state metadata document, every occurrence of the DID must be replaced by `did:0:0`.
```
PhilippGackstatter, line 156: Actually, let's state this in the table, as suggested in another comment, and remove this so it's all in one place.

```suggestion
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,290 +14,210 @@ keywords:
...
-**Type**
+In the `State Metadata` of the Alias Output must be a byte packed payload with header fields as follows:
 
-The proof object MUST include a `type` property. This property references a verification method type's signature algorithm, as standardized in the [DID spec registries](https://www.w3.org/TR/did-spec-registries/#verification-method-types) or standardized via the method specification. 
+| Name          | Type         | Description                                                        |
+|---------------|--------------|--------------------------------------------------------------------|
+| Document Type | ByteArray[3] | Set to value **DID** to denote a DID Document.                     |
+| Version       | uint8        | Set value **1** to denote the version number of this method        |
+| Encoding      | uint8        | Set to value to **0** to denote JSON encoding without compression. |
+| Payload       | ByteArray    | Payload encoded according to `Encoding`.                           |
```
PhilippGackstatter, line 119:
```suggestion
| Payload       | ByteArray    | A DID document where every occurrence of the DID from the document's `id` field is replaced by `did:0:0`. It must be encoded according to `Encoding`.                           |
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,290 +14,210 @@ keywords:
...
-To determine the order of any DID messages, the following steps are to be performed:
-1. Order is initially established by recreating the chain based on the `previousMessageId` linkages. 
-   1. For Int DID messages, the one with no `previousMessageId` is first.
-   2. For Diff DID messages, the one with a `previousMessageId` referencing the IOTA MessageId of the corresponding Integration DID message is first.
-2. When two or more Messages compete, an order must be established between them.
-3. To determine the order, check which milestone confirmed the messages.
-4. If multiple messages are confirmed by the same milestone, we order based on the IOTA MessageId with alphabetical ordering.
+The following steps can be used to read the latest DID document associated with a DID.
+1. Obtain the `Alias ID` from the DID by extracting the `iota-tag` from the DID, see [DID Format](#did-format).
+2. Query the Alias Output corresponding to the `Alias ID`, using a node running the [inx indexer](https://github.com/iotaledger/inx-indexer). Nodes usually include this indexer by default.
```
PhilippGackstatter, line 189: Step 5 might seem odd, but since I would suggest to no longer derive the final DID from the alias id of the output, as stated elsewhere, we should at least do this basic check.

```suggestion
1. Obtain the `Alias ID` from the DID by extracting the `iota-tag` from the DID, see [DID Format](#did-format).
2. Obtain the network of the DID by extracting the `iota-network` from the DID, see [DID Format](#did-format).
3. Query the Alias Output corresponding to the `Alias ID` using a node running the [inx indexer](https://github.com/iotaledger/inx-indexer). Nodes usually include this indexer by default.
4. Assert that the extracted network matches the one returned from the node. Return an error otherwise.
5. Assert that the `Alias ID` of the returned output matches the `Alias ID` extracted from the DID. Return an error otherwise.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,290 +14,210 @@ keywords:
...
+3. Retrieve the value of the `State Metadata` field from the returned output.
+4. Check if its content matches the [Anatomy of the State Metadata](#anatomy-of-the-state-metadata). Return an error otherwise.
+5. Decode the DID document from the state metadata.
+6. Replace the placeholder `did:0:0` with the actual DID which can be derived from the `Alias ID` of the output.
```
PhilippGackstatter, line 193: We don't actually have to derive it from the Alias ID, but can just use the one given as input. That avoids us having to explain how to get the BECH32 HRP/network name. Since we just resolved the DID on the network extracted from the initial DID and ensured it matches in one of the earlier steps, it's fine.

```suggestion
6. Replace the placeholder `did:0:0` with the DID given as input.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,290 +14,210 @@ keywords:
...
-1. Create a new DID Document that contains all the new target state. This SHOULD include the new desired changes and all the changes inside the previous Diff Chain, otherwise these changes are lost!
-2. Retrieve the IOTA MessageId from the previous Integration DID message and a keypair for signing the DID Document.
-3. Set the `previousMessageId` field to the IOTA MessageId value.
-4. Create a `proof` object referencing the public key and signature suite from a verification method in the DID Document able to sign updates as per the [Signing Keys](#signing-keys) section.
-5. Create a cryptographic signature using the same keypair and add the result to the appropriate field(s) inside the `proof` object.
+1. Create the content of the updated DID Document.
+2. Create the payload and the headers as described in the [Anatomy of the State Metadata](#anatomy-of-the-state-metadata).
+3. Create a new Alias Output with the same `Alias ID` as the current output.
+4. Increment the `State Index`.
+5. Leave the unlock conditions unchanged.
```
PhilippGackstatter, line 204: That was my suggestion, but it's actually incorrect. It's perfectly fine to update the unlock conditions during a DID update. Again, I would love to just leave this point out because this is not our concern, it's the concern of output validation, i.e. TIP-18. No matter if we write this, or "Set state controller and governor" like in _Deactivation_, we're stating incorrectly what is actually allowed. Technically, the user can do whatever is valid according to TIP-18.

The correct way would be to say something like "Set valid unlock conditions." at which point we're not saying anything helpful and might as well leave it out.

As discussed, it's fine to leave it as-is for now and address it later, but we should address it at some point.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +232,32 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
 
 ## Privacy Considerations
 
 ### Personal Identifiable Information
 
-The public IOTA Tangle networks are immutable networks. This means that once something is uploaded, it can never be completely removed. That directly conflicts with certain privacy laws such as GDPR, which have a 'right-to-be-forgotten' for Personal Identifiable Information (PII). As such, users should NEVER upload any PII to the Tangle, including inside DID Documents. The IOTA Identity framework allows Verifiable Credentials to be published to the Tangle directly, however this feature should only be utilized by Identity for Organisation and Identity for Things. 
+The public IOTA Tangle networks are immutable networks. This means that once something is uploaded, it can never be completely removed. For example, destroying an Alias Output will remove it from the ledger state, but it can still be stored in permanodes or by any party that records historical ledger states.
```
PhilippGackstatter, line 257: Nit: Wording,

```suggestion
The public IOTA Tangle networks are immutable networks. This means that once something is included, it can never be completely removed. For example, destroying an Alias Output will remove it from the ledger state, but it can still be stored in permanodes or by any party that records historical ledger states.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +232,32 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
+
+That directly conflicts with certain privacy laws such as GDPR, which have a 'right-to-be-forgotten' for Personal Identifiable Information (PII). As such, users should NEVER upload any PII to the Tangle, including inside DID Documents. The IOTA Identity framework allows Verifiable Credentials to be published to the Tangle directly, however this feature should only be utilized by Identity for Organisation and Identity for Things.
```
PhilippGackstatter, line 259: > The IOTA Identity framework allows Verifiable Credentials to be published to the Tangle directly

Not yet? Does that refer to Public VCs?

```suggestion
That directly conflicts with certain privacy laws such as GDPR, which have a 'right-to-be-forgotten' for Personal Identifiable Information (PII). As such, users should NEVER upload any PII to the Tangle, including inside DID Documents. The IOTA Identity framework allows Verifiable Credentials to be published to the Tangle directly, however this feature should only be utilized by Identity for Organisations and Identity for Things.
```

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -312,49 +232,32 @@ The IOTA Identity framework currently supports two Verification Method Types:
...
 
 ### Correlation Risks
 
-As with any DID method, identities can be linked if they are used too often and their usage somehow becomes public. IOTA provides a simple solution for this problem: as creating identities is completely feeless, it is recommended to make new identities on a per interaction basis (Pairwise DIDs). The IOTA Identity framework will provide support for this in the future. 
+As with any DID method, identities can be linked if they are used too often and their usage somehow becomes public. Additionally, a DID can be correlated with funds if the Alias Output used to store the DID Document or any of its controllers is used for holding, transferring or controlling coins or NFTs.
```
PhilippGackstatter, line 263: > As with any DID method, identities can be linked if they are used too often and their usage somehow becomes public. 

I know you didn't write this, but this should be clarified. Currently, it's barely understandable what it's supposed to mean, and how one is supposed to use a DID to avoid or mitigate this problem.

PhilippGackstatter (approved):
A couple more details need to be ironed out, some of which I suggested wrongly in the last review, sorry.

For the follow-up PR we can additionally consider to use the keywords from https://www.rfc-editor.org/rfc/rfc2119.

Other than that, looks good to me if we address the duplication of TIP-18 concerns somewhat in a future PR.

documentation/docs/specs/did/iota_did_method_spec.md
abdulmth: Why is this incorrect? state transition doesn't allow changing controllers.

documentation/docs/specs/did/iota_did_method_spec.md
abdulmth: the payload is explained under the table, it has the document and the metadata too.

documentation/docs/specs/did/iota_did_method_spec.md
PhilippGackstatter: True that. Sorry I got confused on that part :man_shrugging:.

documentation/docs/specs/did/iota_did_method_spec.md
PhilippGackstatter: Yeah, I'm aware, but I primarily wanted everything in one place, so the important information of having replace the `did:0:0` isn't almost hidden under the table in a different place.

eike-hass (approved):

documentation/docs/specs/did/iota_did_method_spec.md
cycraig: It should link out to the W3C specification section, which has some advice on how to mitigate it: https://www.w3.org/TR/did-core/#did-correlation-risks

abdulmth: Thanks Criag, that was my idea too.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -14,290 +14,210 @@ keywords:
...
+| Document Type | ByteArray[3] | Set to value **DID** to denote a DID Document.                     |
+| Version       | uint8        | Set value **1** to denote the version number of this method        |
+| Encoding      | uint8        | Set to value to **0** to denote JSON encoding without compression. |
+| Payload       | ByteArray    | Payload encoded according to `Encoding`.                           |
 
-The IOTA Identity implementation currently supports:
-- `JcsEd25519Signature2020` 
+| Name      | Description                                     |
+|-----------|-------------------------------------------------|
+| ByteArray | A dynamically sized, but unprefixed byte array. |
```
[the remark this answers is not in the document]
abdulmth: I think you mean TIP 21.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -186,11 +187,14 @@ Once the transaction is confirmed, the DID is published and can be formatted by
...
-2. Query the Alias Output corresponding to the `Alias ID`, using a node running the [inx indexer](https://github.com/iotaledger/inx-indexer). Nodes usually include this indexer by default.
-3. Retrieve the value of the `State Metadata` field from the returned output.
-4. Check if its content matches the [Anatomy of the State Metadata](#anatomy-of-the-state-metadata). Return an error otherwise.
-5. Decode the DID document from the state metadata.
-6. Replace the placeholder `did:0:0` with the actual DID which can be derived from the `Alias ID` of the output.
+2. Obtain the network of the DID by extracting the `iota-network` from the DID, see [DID Format](#did-format).
+3. Query the Alias Output corresponding to the `Alias ID` using a node running the [inx indexer](https://github.com/iotaledger/inx-indexer). Nodes usually include this indexer by default.
+4. Assert that the extracted network matches the one returned from the node. Return an error otherwise.
+5. Assert that the `Alias ID` of the returned output matches the `Alias ID` extracted from the DID. Return an error otherwise.
+7. Retrieve the value of the `State Metadata` field from the returned output.
```
PhilippGackstatter, line 194: Numbering.

documentation/docs/specs/did/iota_did_method_spec.md
```diff
@@ -111,12 +111,14 @@ This tag identifies the Alias Output where the DID Document is stored, and it wi
...
+|---------------|--------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Document Type | ByteArray[3] | Set to value **DID** to denote a DID Document.                                                                                                           |
+| Version       | uint8        | Set value **1** to denote the version number of this method                                                                                              |
+| Encoding      | uint8        | Set to value to **0** to denote JSON encoding without compression.                                                                                       |
+| Payload       | ByteArray    | A DID document and its metadata, where every occurrence of the DID in the document is replaced by `did:0:0`. It must be encoded according to `Encoding`. |
+
+Next to [TIP-21](https://github.com/Wollac/protocol-rfcs/blob/serialization/tips/TIP-0021/tip-0021.md), we use the following type definitions:
```
PhilippGackstatter, line 121:
```suggestion
Next to [TIP-21](#data-types--subschema-notation), we use the following type definitions:
```

documentation/docs/specs/did/iota_did_method_spec.md
cycraig: > Query the Alias Output corresponding to the `Alias ID` using a node running the [inx indexer]

> Assert that the extracted network matches the one returned from the node. Return an error otherwise.

We should probably state which REST APIs to call?

PR #976, opened 2022-08-19
Improve Wasm TypeScript build

eike-hass:
# Description of change
Use tsconfig `paths` to resolve modules from wasm folder. Added `tsc-alias` to then fix the paths for the actual output.

Moved and renamed the `tsconfig.node.json` so editors can resolve the `path` configuration.

Targeting `feat/wasm-shimmer-client` to reason about the change.
Can only be merged after the `DID` type in the bindings is fixed, since the problem leads to build errors.
Then should probably be retargeted to `dev`.

## Type of change

- [ ] Documentation Fix

## Change checklist


bindings/wasm/lib/stardust_identity_client.ts
```diff
@@ -1,8 +1,7 @@
 // Copyright 2020-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-// @ts-ignore: path is set to match runtime transpiled js path when bundled.
-import {IStardustIdentityClient, StardustDID, StardustDocument, StardustIdentityClientExt} from './identity_wasm';
+import {IStardustIdentityClient, StardustDID, StardustDocument, StardustIdentityClientExt} from 'identity_wasm';
 
 // NOTE: this import path is replaced with `/web` in the `build/web.js` script.
 import type {Client, INodeInfoWrapper, SecretManager} from '@cycraig/iota-client-wasm/node';
```
cycraig, line 7: Can we use the same path alias for the client import too?

bindings/wasm/package.json
```diff
@@ -11,8 +11,8 @@
   },
   "scripts": {
     "build:src": "cargo build --lib --release --target wasm32-unknown-unknown",
-    "bundle:nodejs": "wasm-bindgen target/wasm32-unknown-unknown/release/identity_wasm.wasm --typescript --weak-refs --target nodejs --out-dir node && node ./build/node && tsc --project ./tsconfig.node.json --outDir node",
-    "bundle:web": "wasm-bindgen target/wasm32-unknown-unknown/release/identity_wasm.wasm --typescript --weak-refs --target web --out-dir web && node ./build/web && tsc --project ./tsconfig.web.json --outDir web",
+    "bundle:nodejs": "wasm-bindgen target/wasm32-unknown-unknown/release/identity_wasm.wasm --typescript --weak-refs --target nodejs --out-dir node && node ./build/node && tsc --project ./lib/tsconfig.json && tsc-alias -p ./lib/tsconfig.json",
+    "bundle:web": "wasm-bindgen target/wasm32-unknown-unknown/release/identity_wasm.wasm --typescript --weak-refs --target web --out-dir web && node ./build/web && tsc --project ./lib/tsconfig.web.json && tsc-alias -p ./lib/tsconfig.web.json",
```
cycraig, line 15: This PR moved and renamed the `tsconfig.node.json` file, the other scripts using `ts-mocha` in `package.json` need to be updated with the new path: `test:account:node`, `test:stardust`, `test:unit:node`.

bindings/wasm/lib/stardust_identity_client.ts
eike-hass: Will absolutely try!

cycraig: I tried `tsc-alias` with `paths` myself yesterday and didn't make much progress, but hopefully I was just missing something obvious.

bindings/wasm/package.json
cycraig: This actually breaks all the tests even after changing the path to `tsconfig`, I don't know how to run `tsc-alias` with `ts-mocha`.

```
  1) Uncaught error outside test suite:
     Uncaught Cannot find module 'identity_wasm'
```

bindings/wasm/lib/stardust_identity_client.ts
cycraig: The repo for `ts-mocha` says it uses [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) and to pass `--paths` for it to work? 

https://github.com/piotrwitek/ts-mocha

Should we use `tsconfig-paths` instead? https://npmtrends.com/tsc-alias-vs-tsconfig-paths-vs-tscpaths-vs-tspath

cycraig:
> Can only be merged after the DID type in the bindings is fixed, since the problem leads to build errors.

Fixed on `dev` branch by #977 now.

bindings/wasm/lib/stardust_identity_client.ts
eike-hass: My understanding is that `tsconfig-paths` works at runtime, while `ts-alias` fixes the paths in the artifacts. So we might actually need to use both 🙈

cycraig (approved):
Looks good to me, thanks for replacing my hacks!

Only nit is having four `tsconfig` files.

PR #1039, opened 2022-09-21
Update RevocationBitmap spec

### LINKED ISSUE #911: [Task] Clarify `id` field restriction in `StatusList2021`
## Description

`StatusList2021` and `RevocationList2020` mandate for the `id` field of `credentialStatus`:

> It MUST NOT be the URL for the status list.

It's unclear why this is the case and we did not adopt this restriction for `RevocationBitmap2022`, so we should clarify why this is the case in those specs.

An [issue](https://github.com/w3c-ccg/vc-status-list-2021/issues/31) was opened to ask for clarification.

## Resources

https://github.com/iotaledger/identity.rs/pull/879#discussion_r895764087
https://github.com/iotaledger/identity.rs/pull/879#discussion_r900129500

PhilippGackstatter:
# Description of change

Changes the spec and implementation of `RevocationBitmap2022` to recommend making the `id` of `credentialStatus` unique. This is recommended by the VC spec as mentioned in #911.

We decided to go with a backwards-compatible change that only RECOMMENDS setting the `index`, and omitting it is allowed spec-wise. The implementation will always set the index to match the recommendation.

fixes #911

## Type of change

- [ ] Documentation Fix

Extended and modified existing tests to test the new behavior.

## Change checklist

### DIFF
--- a/documentation/docs/specs/revocation_bitmap_2022.md
+++ b/documentation/docs/specs/revocation_bitmap_2022.md
@@ -33,7 +33,7 @@ For an issuer to enable verifiers to check the status of a verifiable credential
 
 | Property | Description |
 | :--- | :--- |
-| `id` | The constraints on the `id` property are listed in the [Verifiable Credentials Data Model specification](https://www.w3.org/TR/vc-data-model/). The `id` MUST be a [DID URL](https://www.w3.org/TR/did-core/#did-url-syntax) that is the URL to a [Revocation Bitmap Service](#revocation-bitmap-service) in the DID Document of the issuer, with an `index` query set to the same value as `revocationBitmapIndex`. |
+| `id` | The constraints on the `id` property are listed in the [Verifiable Credentials Data Model specification](https://www.w3.org/TR/vc-data-model/). The `id` MUST be a [DID URL](https://www.w3.org/TR/did-core/#did-url-syntax) that is the URL to a [Revocation Bitmap Service](#revocation-bitmap-service) in the DID Document of the issuer. It SHOULD include an `index` query set to the same value as `revocationBitmapIndex`, to uniquely identify the `credentialStatus`. If the `index` query is present, implementations SHOULD reject statuses where the `index` query's value does not match `revocationBitmapIndex`. |
 | `type` | The `type` property MUST be `"RevocationBitmap2022"`. |
 | `revocationBitmapIndex` | The `revocationBitmapIndex` property MUST be an unsigned, 32-bit integer expressed as a string. This is the index of the credential in the issuer's revocation bitmap. Each index SHOULD be unique among all credentials linking to the same [Revocation Bitmap Service](#revocation-bitmap-service). |
 
--- a/documentation/docs/specs/revocation_bitmap_2022.md
+++ b/documentation/docs/specs/revocation_bitmap_2022.md
@@ -15,7 +15,7 @@ keywords:
 
 ## Abstract
 
-This specification describes an on-Tangle mechanism for publishing the revocation status of [Verifiable Credentials](../concepts/verifiable_credentials/overview) embedded in an issuer's DID document.
+This specification describes an on-ledger mechanism for publishing the revocation status of [Verifiable Credentials](../concepts/verifiable_credentials/overview) embedded in an issuer's DID document.
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


documentation/docs/specs/revocation_bitmap_2022.md
```diff
@@ -15,7 +15,7 @@ keywords:
 
 ## Abstract
 
-This specification describes an on-Tangle mechanism for publishing the revocation status of [Verifiable Credentials](../concepts/verifiable_credentials/overview) embedded in an issuer's DID document.
+This specification describes an on-ledger mechanism for publishing the revocation status of [Verifiable Credentials](../concepts/verifiable_credentials/overview) embedded in an issuer's DID document.
```
cycraig, line 18: Nit: both "on-ledger" and "on-Tangle" are technically inaccurate since it depends on the DID Method implementation, which is not the concern of the revocation specification. I would remove it since the sentence goes on to mention that it is "embedded" in the DID Document, but it's fine either way.

documentation/docs/specs/revocation_bitmap_2022.md
```diff
@@ -33,7 +33,7 @@ For an issuer to enable verifiers to check the status of a verifiable credential
 
 | Property | Description |
 | :--- | :--- |
-| `id` | The constraints on the `id` property are listed in the [Verifiable Credentials Data Model specification](https://www.w3.org/TR/vc-data-model/). The `id` MUST be a [DID URL](https://www.w3.org/TR/did-core/#did-url-syntax) that resolves to a [Revocation Bitmap Service](#revocation-bitmap-service) in the DID Document of the issuer. |
+| `id` | The constraints on the `id` property are listed in the [Verifiable Credentials Data Model specification](https://www.w3.org/TR/vc-data-model/). The `id` MUST be a [DID URL](https://www.w3.org/TR/did-core/#did-url-syntax) that is the URL to a [Revocation Bitmap Service](#revocation-bitmap-service) in the DID Document of the issuer. It SHOULD include an `index` query set to the same value as `revocationBitmapIndex`, to uniquely identify the `credentialStatus`. If the `index` query is present, implementations SHOULD reject statuses where the `index` query's value does not match `revocationBitmapIndex`. |
```
cycraig, line 36: Dereferencing a service endpoint according to the DID Resolution specification uses a different approach.

E.g.
https://w3c-ccg.github.io/did-resolution/#example-9

I find this inconsistent and annoying, and I think the way it is now (i.e. `did:example:123?index=5#service`) would still be compatible since [dereferencing a primary DID Url resource](https://w3c-ccg.github.io/did-resolution/#dereferencing-algorithm-primary) only applies if we use the `?service` query parameter? Is there any reason to switch to their approach, which would be a breaking change?

identity_credential/src/credential/revocation_bitmap_status.rs
```diff
@@ -19,14 +19,37 @@ use crate::error::Result;
 pub struct RevocationBitmapStatus(Status);
 
 impl RevocationBitmapStatus {
-  const INDEX_PROPERTY_NAME: &'static str = "revocationBitmapIndex";
+  const REVOCATION_BITMAP_INDEX_PROPERTY: &'static str = "revocationBitmapIndex";
```
cycraig, line 22: Nit: bit redundant since it's accessed as `RevocationBitmapStatus::REVOCATION_BITMAP_INDEX_PROPERTY`, whereas before it was `RevocationBitmapStatus::INDEX_PROPERTY_NAME`, but explicitness is probably preferred.

identity_credential/src/credential/revocation_bitmap_status.rs
```diff
@@ -65,19 +88,57 @@ impl TryFrom<Status> for RevocationBitmapStatus {
...
+        return Err(Error::InvalidStatus(format!(
+          "missing required property '{}'",
+          Self::REVOCATION_BITMAP_INDEX_PROPERTY
+        )));
+      };
+
+    let revocation_bitmap_index: u32 = if let Value::String(index) = revocation_bitmap_index {
+      index.parse::<u32>().map_err(|err| {
+        Error::InvalidStatus(format!(
+          "property '{}' cannot be converted to an unsigned, 32-bit integer: {err}",
```
cycraig, line 111: Nit: the error messages for `u32` conversions are inconsistent. See line 72 and 79 above for the index property conversion.

This index property validation code is also duplicated, can it be extracted into a function instead?

cycraig (approved):
Minor comments.

cycraig:
Add the `Wasm` label too, this affects all bindings and should reflect in their changelogs.

documentation/docs/specs/revocation_bitmap_2022.md
PhilippGackstatter: > Is there any reason to switch to their approach, which would be a breaking change?

I completely agree that this feels inconsistent. I would be fine with keeping it as-is now, too, but I'm concerned a conformant resolver would run into branch `3)` in dereferencing the primariy DID Url resource, rather than `2)`, because the DID Url now contains a query:

> 3) Otherwise, if the input [DID URL](https://w3c-ccg.github.io/did-resolution/#dfn-did-url) contains a [DID path](https://w3c-ccg.github.io/did-resolution/#dfn-did-path) and/or [DID query](https://w3c-ccg.github.io/did-resolution/#dfn-did-query):

https://w3c-ccg.github.io/did-resolution/#example-3

Or is that not applicable for some reason?

cycraig: It will take branch `3`, so it's up to the universal resolver implementation what to do at that point.

I think it's fine? The entire service should be validated, so getting only the service endpoint seems inapplicable. I don't know why they introduced a special case to ignore the service type etc. and only extract the endpoint.

PhilippGackstatter: Yeah probably. Will leave it as-is.

HenriqueNogara (approved):

Issue #1060, opened 2022-10-21
[Task] Add credential and presentation examples


PhilippGackstatter:
## Description

Add examples for creating, signing, verifying and revoking credentials and presentations. These should match the examples from 0.6 in functionality as much as possible.

## Motivation

Example parity with the 0.6 version.

## Resources

- https://github.com/iotaledger/identity.rs/blob/support/v0.6/examples/account/create_vc.rs
- https://github.com/iotaledger/identity.rs/blob/support/v0.6/examples/account/create_vp.rs
- https://github.com/iotaledger/identity.rs/blob/support/v0.6/examples/account/revoke_vc.rs

## To-do list
Create a task-specific to-do list. Please link PRs that match the TODO list item behind the item after it has been submitted.

- [ ] Port the examples listed in _Resources_ for Rust and Wasm.

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

abdulmth:
Wasm examples still missing

PR #1070, opened 2022-11-07
Add v. credentials and presentations examples

abdulmth:
# Description of change
Add missing verifiable credentials and verifiable presentations examples

#1060 

## Type of change
- [ ] Documentation Fix

### DIFF
--- a/examples/0_basic/6_create_vp.rs
+++ b/examples/0_basic/6_create_vp.rs
@@ -23,7 +23,6 @@ use identity_iota::credential::FailFast;
 use identity_iota::credential::Presentation;
 use identity_iota::credential::PresentationBuilder;
 use identity_iota::credential::PresentationValidationOptions;
-use identity_iota::credential::PresentationValidator;
 use identity_iota::credential::Subject;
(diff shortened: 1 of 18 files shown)


examples/0_basic/6_create_vp.rs
```diff
@@ -0,0 +1,198 @@
...
+    .shared_validation_options(credential_validation_options)
+    .subject_holder_relationship(SubjectHolderRelationship::AlwaysSubject);
+
+  PresentationValidator::validate(
+    &presentation,
+    &alice_document,
+    &[issuer_document],
+    &presentation_validation_options,
+    FailFast::FirstError,
+  )?;
```
olivereanderson, line 189: I think it might be better to use the `Resolver::verify_presentation` API here and demonstrate that it will resolve any DID documents it needs that are not passed in. Perhaps the example can assume that the verifier already has Alice's document, but not the issuers?

olivereanderson (approved):
LGTM,

Thanks for working on this!

olivereanderson (approved):
Looks good to me, 

Thanks for working on this!

PR #1076, opened 2022-11-14
Add revocation examples

abdulmth:
# Description of change
Add example for revoking verifiable credentials in Rust and Wasm.

#1060 

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/bindings/wasm/examples/src/0_basic/7_revoke_vc.ts
+++ b/bindings/wasm/examples/src/0_basic/7_revoke_vc.ts
@@ -1,0 +1,157 @@
+// Copyright 2020-2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { Bip39 } from "@iota/crypto.js";
+import { Client, MnemonicSecretManager } from "@iota/iota-client-wasm/node";
+import { IAliasOutput, IRent, TransactionHelper } from "@iota/iota.js";
+import {
+    Credential,
+    CredentialValidationOptions,
+    CredentialValidator,
+    FailFast,
+    IotaDocument,
+    IotaIdentityClient,
+    IotaService,
+    IotaVerificationMethod,
+    ProofOptions,
+    Resolver,
+    RevocationBitmap,
+} from "../../../node";
+import { API_ENDPOINT, createDid } from "../util";
+
+/**
+ * This example shows how to revoke a verifiable credential.
+ * It demonstrates two methods for revocation. The first uses a revocation bitmap of type `RevocationBitmap2022`,
+ * while the second method simply removes the verification method (public key) that signed the credential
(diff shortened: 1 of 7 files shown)


eike-hass:
I assume for Wasm the example needs to be added to the tests as well. For Rust that is automatic, right?

bindings/wasm/examples/src/0_basic/7_revoke_vc.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    // Sign Credential.
+    let signedVc = issuerDocument.signCredential(unsignedVc, keypairIssuer.private(), "#key-1", ProofOptions.default());
+    console.log(`Credential JSON > ${JSON.stringify(signedVc, null, 2)}`);
+
+    // ===========================================================================
+    // Revocation of the Verifiable Credential.
+    // ===========================================================================
+
```
olivereanderson, line 105: Perhaps we should also demonstrate that the credential can be verified successfully prior to revocation?

examples/0_basic/7_revoke_vc.rs
```diff
@@ -0,0 +1,192 @@
...
+    .build()?;
+
+  // Sign the Credential with the issuer's verification method.
+  issuer_document.sign_data(&mut credential, key_pair.private(), "#key-1", ProofOptions::default())?;
+  println!("Credential JSON > {:#}", credential);
+
+  // ===========================================================================
+  // Revocation of the Verifiable Credential.
+  // ===========================================================================
+
```
olivereanderson, line 133: Again maybe we should demonstrate that the credential can be successfully verified at this point before it gets revoked?

examples/0_basic/7_revoke_vc.rs
```diff
@@ -0,0 +1,192 @@
...
+  let original_method: DIDUrl<IotaDID> = issuer_document.resolve_method("#key-1", None).unwrap().id().clone();
+  issuer_document.remove_method(&original_method).unwrap();
+
+  // Publish the changes.
+  let alias_output: AliasOutput = client.update_did_output(issuer_document.clone()).await?;
+  let alias_output: AliasOutput = AliasOutputBuilder::from(&alias_output).finish(client.get_token_supply()?)?;
+  client.publish_did_output(&secret_manager_issuer, alias_output).await?;
+
+  // We expect the verifiable credential to be revoked.
+  let mut resolver: Resolver = Resolver::new();
```
olivereanderson, line 174: nit: Possibly better to use `Resolver<IotaDocument>` here instead. In which case line 176 also needs to be changed to use an `IotaDocument` instead. 

You can decide what you want to do here.

olivereanderson (approved):
LGTM!

Wrote a few comments.

PR #1080, opened 2022-11-16
Remove legacy crates

PhilippGackstatter:
# Description of change

## Added

- Add `IotaDID::is_placeholder` which was required for the updated `identity_agent` tests and seemed generally useful to have.

## Changed

- Remove `identity_account`.
- Remove `examples_legacy`.
- Remove most of `identity_account_storage` except
  - src/stronghold/* (Stronghold wrapper types)
  - src/types/encryption/* (encryption types)
  - src/utils/crypto.rs (PBKDF2 key derivation)
- Remove dependency of `identity_agent` on `identity_account` and re-add it to the root `Cargo.toml`.
- Remove most of `bindings/wasm/src/account/`.
- Comment out CI checks for `stronghold-nodejs` bindings because it depended on `identity_account_storage`.

## Links to any relevant issues

n/a

## Type of change
Add an `x` to the boxes that are relevant to your changes.

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

## How the change has been tested

CI runs.

## Change checklist
Add an `x` to the boxes that are relevant to your changes.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1081, opened 2022-11-17
Use alias instead of relative paths in examples

abdulmth:
# Description of change
Fix examples' links for VC wiki pages.

This PR also replaces the examples' relative paths in imports eg: `import { IotaDocument } from "../../../node";` with an alias, eg: 
`import { IotaDocument } from "@iota/identity-wasm/node"`;
This makes it match the import from a normal node module.
Adding paths to `.tsconfig` is not enough since `ts-node` and `ts-mocha` can't resolve these aliases. [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) solves this problem nicely. 

## Type of change
- [x] Documentation Fix


eike-hass (approved):

PhilippGackstatter:
FYI: I also fixed the example links in #1080. Depending what we merge first, there might be merge conflicts.

Issue #154, opened 2021-03-05
Domain Name Verification


JelleMillenaar:
Implement and standardize a service endpoint called DomainNameVerification. It links to a domain that is owned by the identity, such as the website of the organization. This domain can be queried, which should resolve in a standardized JSON response that lists the DID(s) used to represent the domain owner. If the DID that lists the service is also listed in the returning query, we have proof that the DID is in control over the domain. This URL may be displayed to a Verifier to add additional trust to an identity. 

Resources:
[Presentation (2019)](https://youtu.be/xQSIvt1aq8c?t=7822) that explains the concept on a high-level.
TODO: Find other resources of other projects

- [ ] Write a specification for the DomainNameVerification service endpoint
- [ ] Implement service
- [ ] Create a test / example

nothingismagick:
While domain querying with http can be useful, I would recommend considering something like proof at the DNS layer. Something like how a DMARC TXT record can be used. 

<img width="705" alt="image" src="https://user-images.githubusercontent.com/35242872/110120711-b42bea80-7dbd-11eb-9f06-d91071e6409c.png">

JelleMillenaar:
Interesting @nothingismagick, that would remove some weaknesses in the design. Some quick research from my end shows we would have to use [TXT records](https://en.wikipedia.org/wiki/TXT_record#cite_note-rfc1035-3) which allows a maximum of 255 characters and is used for Verification of Domain Ownership by [Google](https://support.google.com/webmasters/answer/9008080?hl=en&visit_id=637505475005258768-2598258310&rd=1#zippy=%2Cdomain-name-provider%2Ctxt-record-instructions) and [Facebook](https://developers.facebook.com/docs/sharing/domain-verification/verifying-your-domain). It seems you can even bypass the 255 character limit by [splitting records](https://support.pagely.com/hc/en-us/articles/360038558812-Resolving-TXT-Record-Size-Limits), which would allow an array of multiple DIDs associated with an organization. This does of course create linkability between those DIDs but would allow an organization to be represented with multiple DID methods.

sdellava:
I would suggest to generalize the service endpoint: If the resolved query cointain a list of DID, the service will return it, otherwise it should return the content with the appropriate content type (JSON, HTML, ... ). 

The service point can be used for many different use cases.

Issue #1017, opened 2022-09-14
[Task] Refactor errors to avoid delegating `Display` to the source


olivereanderson:
## Description
Currently the errors of the IOTA Identity framework do not follow the following recommendation from the Rust error handling working group :
> An error type with a source error should either return that error via source or include that source's error message in its own Display output, but never both.

Not following this recommendation leads to duplicate error messages when printing the error along with its causes (using for instance the popular error handling crate `anyhow` or the upcoming [`Report` type](https://doc.rust-lang.org/std/error/struct.Report.html)). See the [What the error handling project group is working towards blog post](https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html) for more details. 

Example of somewhere where the recommendation is not followed: 
```
#[derive(Debug, thiserror::Error, strum::IntoStaticStr)]
pub enum identity_did::Error {
// .. 
  #[error("{0}")]
  InvalidDID(#[from] crate::did::DIDError), 
// .. 
}
```

Note that the `#[from]` annotation also means that `Error::source` is delegated to the  wrapped error `crate::did::DIDError` whenever the variant `InvalidDID` is matched.  

Now this code: 
```
use identity_did::{Error, did::DIDError}; 

fn main() -> anyhow::Result<()> {
    let err = Error::InvalidDID(DIDError::Other("DID length exceeds the current maximum"));
    Err(err)?;
    Ok(())
}
``` 

produces the following output:
```
Error: DID length exceeds the current maximum

Caused by:
    DID length exceeds the current maximum
```

In other words the error message is duplicated. 

If we instead had defined the variant as follows:
```
  #[error("invalid DID handling")]
  InvalidDID(#[from] crate::did::DIDError),
```
the output would then be 
```
Error: invalid DID handling

Caused by:
    DID length exceeds the current maximum
```
which does not duplicate the message in "Caused by".  

## Motivation
This is a Subtask of https://github.com/iotaledger/identity.rs/issues/534. Fixing this issue would unlock the possibility to explore using `anyhow` in the Wasm bindings to conveniently produce better error messages there.

PR #1094, opened 2022-11-24
Add Support for Domain Linkage

abdulmth:
# Description of change
- Add [DIDConfigurationResource](https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-resource) and a convenient builder for [DomainLinkageCredential](https://identity.foundation/.well-known/resources/did-configuration/#domain-linkage-credential).
- Add domain linkage example.
- Domain Linkage Verifier.

#154

## Type of change

- [ ] Documentation Fix


identity_resolver/src/resolution/domain_linkage_verifier.rs
```diff
@@ -0,0 +1,278 @@
...
+    let configuration_string: &str = include_str!("./tests/fixtures/did_configuration/config1.json");
+    let mut configuration: DIDConfigurationResource =
+      DIDConfigurationResource::from_json(configuration_string).unwrap();
+    let credential: &mut Credential = configuration.linked_dids_mut().get_mut(0).unwrap();
+    let (doc, _) = compose_did_document(credential.issuer.url().to_string());
+    let res = verify_config(doc, &configuration).await;
+    assert!(res.is_err());
+  }
+
+  // Todo: Add further tests for error cases.
```
PhilippGackstatter, line 248: Indeed it would be great to verify that the cases that `verify_domain_linkage_credential` seems to catch are indeed caught. You can setup a general function that does the generic setup (load credential, create issuer document, create configuration resource) and then take a closure that modifies a small part of the test and verify the result with another closure.

identity_resolver/src/resolution/domain_linkage_verifier.rs
```diff
@@ -0,0 +1,278 @@
...
+    let resolved_doc: AbstractThreadSafeValidatorDocument = self.resolver.resolve(issuer_did).await.map_err(|err| {
+      Error::new(ErrorCause::DNSVerificationError {
+        reason: format!("Issuer resolution failed! {}", err),
+      })
+    })?;
+    Ok(resolved_doc)
+  }
+}
+
+mod tests {
```
PhilippGackstatter, line 204:
```suggestion
#[cfg(test)]
mod tests {
```

identity_credential/src/credential/did_configuration_resource.rs
```diff
@@ -0,0 +1,302 @@
...
+  fn builder() {
+    let issuer = Issuer::Url(Url::parse("did:example:issuer").unwrap());
+    let credential: Credential = DomainLinkageCredentialBuilder::new()
+      .issuance_date(Timestamp::now_utc())
+      .expiration_date(Timestamp::now_utc())
+      .issuer(issuer)
+      .origin(Url::parse("http://www.example.com").unwrap())
+      .build()
+      .unwrap();
+    let _config: DIDConfigurationResource = DIDConfigurationResource::new(vec![credential]);
```
PhilippGackstatter, line 216: This line can be removed, I guess?

identity_credential/src/credential/did_configuration_resource.rs
```diff
@@ -0,0 +1,302 @@
...
+      .origin(Url::parse("http://www.example.com").unwrap())
+      .build();
+
+    assert!(matches!(credential, Err(Error::MissingExpirationDate)));
+  }
+
+  #[test]
+  fn from_json() {
+    const JSON1: &str = include_str!("../../tests/fixtures/well-known-configuration.json");
+    let _well_known: DIDConfigurationResource = DIDConfigurationResource::from_json(JSON1).unwrap();
```
PhilippGackstatter, line 257: Nit: In these cases it's fine to:
```suggestion
    DIDConfigurationResource::from_json(JSON1).unwrap();
```

identity_credential/src/credential/did_configuration_resource.rs
```diff
@@ -0,0 +1,302 @@
...
+    let mut domains = OrderedSet::new();
+    domains.append(Url::parse("https://foo.example-1.com").unwrap());
+    domains.append(Url::parse("https://bar.example-2.com").unwrap());
+
+    let service: Service<CoreDID, Object> = DIDConfigurationResource::create_linked_domain_service(
+      CoreDIDUrl::parse("did:example:123#foo").unwrap(),
+      domains,
+    ).unwrap();
+
+    // Add a new Service.
```
PhilippGackstatter, line 271:
```suggestion
```

identity_credential/src/credential/did_configuration_resource.rs
```diff
@@ -0,0 +1,302 @@
...
+    let service_from_json: Service<CoreDID, Object> = Service::from_json_value(json!({
+        "id":"did:example:123#foo",
+        "type": "LinkedDomains",
+        "serviceEndpoint": {
+          "origins": ["https://foo.example-1.com", "https://bar.example-2.com"]
```
PhilippGackstatter, line 276: Nit: Would be slightly better to define these urls once and then include them here to prevent copy-paste mistakes.

examples/0_basic/8_domain_linkage.rs
```diff
@@ -0,0 +1,135 @@
...
+  // The DID Configuration resource can be fetched using
+  // let configuration_resource: DIDConfigurationResource = DomainLinkageVerifier::fetch_configuration(Url::parse("https://foo.example.com/".to_owned())?).await?;
+  // But since we're working locally, we will simply deserialize the JSON.
+  let configuration_resource: DIDConfigurationResource =
+    DIDConfigurationResource::from_json(&configuration_resource_json)?;
+
+  // Verify the "fetched" DID Configuration resource.
+  let verification_result = verifier.verify_configuration(&configuration_resource).await;
+
+  // Verification returns ok
```
PhilippGackstatter, line 113:
```suggestion
  // Verification succeeds.
```

PhilippGackstatter (changes requested):
Thanks for working on this! Really looking forward to this feature.

Currently, `DomainLinkageVerifier` lives in `identity_resolver`. However, if we can, I think that type should mirror the `PresentationValidator` and `CredentialValidator` types more. Those live in `identity_credential` and are injected with the necessary issuers and holder from the `Resolver` which is a nice separation of concerns. `identity_resolver` should resolve things, `identity_credential` should verify things, roughly speaking. (I'm going to slightly break that myself in the following suggestions).

I would suggest to

- move `DomainLinkageVerifier` to `identity_credential`. 
- change signature of `DomainLinkageVerifier::verify_domain_linkage_credential` to `DomainLinkageVerifier::verify_domain_linkage_credential<DOC: ValidatorDocument>(issuers: &[DOC], credential: &Credential, verifier_options: &VerifierOptions)` and make that method public. It does what `DomainLinkageVerifier::verify_domain_linkage_credential` currently does minus the resolution. The issuer documents are passed in, similar to `PresentationValidator` and `CredentialValidator`. That is handled either by the user themselves by calling `Resolver::resolve_issuers` or by a convenience method we can add: `Resolver::verify_domain_linkage_credential(credential: &Credential, verifier_options: &VerifierOptions)`.
  - I think it's fine to provide a method to verify all linked credentials in a configuration, but it should also be possible to verify an individual one as mentioned in the spec. Hence my suggestion to make `verify_domain_linkage_credential` public.
  - Since we're verifying credentials I think we should also take `VerifierOptions` here and pass that down.
- move `DomainLinkageVerifier::fetch_configuration` in `DIDConfigurationResource::fetch` but feature-gated so `identity_credential` does not depend on `reqwest`  by default.
- rename `DIDConfigurationResource` to `DomainLinkageConfiguration`. The spec name is pretty nondescript.

identity_resolver/src/resolution/domain_linkage_verifier.rs
```diff
@@ -0,0 +1,278 @@
...
+
+  /// Verifies a [DID Configuration Resource](https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-resource) for the following criteria:
+  /// - [Domain Linkage Credential](https://identity.foundation/.well-known/resources/did-configuration/#domain-linkage-credential)s
+  ///   are spec compliant and not expired.
+  /// - [DID Configuration Resource Verification](https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-resource-verification)
+  ///
+  /// Note:
+  /// - Only [Linked Data Proof Format](https://identity.foundation/.well-known/resources/did-configuration/#linked-data-proof-format)
+  ///   is supported.
+  /// - The controller of the DID is not verified.
```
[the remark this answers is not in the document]
abdulmth: In the [specs](https://identity.foundation/.well-known/resources/did-configuration/#did-configuration-resource-verification) there's a point about he controller: 

> a Verifier SHOULD consider the entity controlling the origin and the Controller of the DID to be the same entity.

This is not verified, maybe it should be more precise here.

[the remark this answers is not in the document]
PhilippGackstatter: Thanks for the pointer. The spec just notes what assumptions we can make after successful verification, so I think we can remove that bullet point here.

identity_credential/src/credential/did_configuration_resource.rs
```diff
@@ -0,0 +1,302 @@
...
+
+  /// List of extra properties (should be empty to be spec compliant).
+  pub fn properties_mut(&mut self) -> &mut Object {
+    &mut self.properties
+  }
+
+  /// Convenient function to create a spec compliant [Linked Domain Service Endpoint](https://identity.foundation/.well-known/resources/did-configuration/#linked-domain-service-endpoint)
+  pub fn create_linked_domain_service<D: DID>(
+    did_url: DIDUrl<D>,
+    domains: impl Into<OrderedSet<Url>>,
```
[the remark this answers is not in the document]
olivereanderson: Although the Well Known DID Configuration spec allows for an array, the DID Core spec specifies that a [ service endpoint must be a string, a map or a set] hence I think it is correct to keep `OrderedSet<Url>` here.

olivereanderson (changes requested):
Thanks for tackling this, it is indeed a very useful feature! 

In addition to @PhilippGackstatter's feedback I would suggest adding the following Error types to the `identity_credential` crate in the validator module .

```rust 
struct DomainLinkageCredentialVerificationError {
    // could also consider a vec of these if you want to aggregate all errors in a 
    // given credential, but it is probably fine to throw an error as soon as one is encountered. 
    cause: DomainLinkageCredentialVerificationErrorCause, 
    source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>
}
// implement a "getter" for the cause field. 

#[non_exhaustive]
enum DomainLinkageCredentialErrorCause {
    MissingExpirationDate,
    ExceededExpirationDate,
    MissingOrigin,
    OriginMismatch, 
    ImpermissibleIdProperty, 
    IssuerSubjectMismatch,
    InvalidIssuanceDate,
    SignatureVerificationFailure,
}

struct DIDConfigurationResourceVerificationError {
   credential_idx_to_error: BtreeMap<usize, DomainLinkageCredentialVerificationError >
}
```
and these should be thrown  by the corresponding verification methods on the `DomainLinkageVerifier` .  The errors should implement the `Error` trait (at least those types whose name end in `Error`. For the `ErrorCause` it is really enough to just implement `Display`, but you might want to use `thiserror` for that in which case `Error` is also implemented). Note that the names I suggested are very long so please feel free to come up with better names!


In the `identity_resolver` crate you can extend the `ErrorCause` enum with the following  variants:
```
.. 
DomainLinkageVerification(DomainLinkageVerificationError), 
DIDResourceConfigurationVerification(DIDResourceConfigurationError),
ConfigurationResourceFetching(Box<dyn std::error::Error + Send + Sync + 'static>), 
}

```
and the `ResolutionAction` enum in that crate can also be extended with a variant giving a hint about the index of the domain linkage credential whose issuer could not be resolved similar to the existing `PresentationIssuersResolution(usize)` variant.

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,265 @@
...
+                return Err(DomainLinkageVerificationError {
+                  cause: DomainLinkageVerificationErrorCause::IssuerSubjectMismatch,
+                  source: None,
+                });
+              }
+            }
+            Err(_) => {
+              return Err(DomainLinkageVerificationError {
+                cause: DomainLinkageVerificationErrorCause::InvalidSubjectId,
+                source: None,
```
PhilippGackstatter, line 199:
```suggestion
            Err(err) => {
              return Err(DomainLinkageVerificationError {
                cause: DomainLinkageVerificationErrorCause::InvalidSubjectId,
                source: Some(Box::new(err)),
```

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,265 @@
...
+          .ok_or(DomainLinkageVerificationError {
+            cause: DomainLinkageVerificationErrorCause::InvalidSubjectOrigin,
+            source: None,
+          })?;
+        let origin_url = Url::parse(origin)
+          .map_err(|_err| Url::parse("https://".to_owned() + origin))
+          .map_err(|_err| DomainLinkageVerificationError {
+            cause: DomainLinkageVerificationErrorCause::InvalidSubjectOrigin,
+            source: None,
+          })?;
```
PhilippGackstatter, line 222: This maps the error of the first `Url::parse` to the result of the second `Url::parse` which is not what we want. It also always retries to parse the Url no matter what error is returned.
Suggestion:

```suggestion
     let origin_url: Url = match Url::parse(origin) {
          Ok(url) => url,
          Err(identity_core::Error::InvalidUrl(url::ParseError::RelativeUrlWithoutBase)) => {
            match Url::parse("https://".to_owned() + origin) {
              Ok(url) => url,
              Err(err) => {
                return Err(DomainLinkageVerificationError {
                  cause: DomainLinkageVerificationErrorCause::InvalidSubjectOrigin,
                  source: Some(Box::new(err)),
                });
              }
            }
          }
          Err(err) => {
            return Err(DomainLinkageVerificationError {
              cause: DomainLinkageVerificationErrorCause::InvalidSubjectOrigin,
              source: Some(Box::new(err)),
            });
          }
        };
```

Requires `url = { version = "2.2", default-features = false }` in the `Cargo.toml`.

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,265 @@
...
+  use crate::credential::Credential;
+  use crate::credential::DomainLinkageConfiguration;
+  use crate::validator::domain_linkage_validator::DomainLinkageVerifier;
+  use crate::validator::CredentialValidationOptions;
+  use identity_core::common::Url;
+  use identity_core::convert::FromJson;
+  use identity_did::document::CoreDocument;
+
+  #[test]
+  pub(crate) fn test() {
```
PhilippGackstatter, line 253: Include more tests, use better test names :wink:.

identity_credential/src/validator/errors.rs
```diff
@@ -139,4 +140,65 @@ impl Display for CompoundPresentationValidationError {
...
+}
+
+impl Display for DomainLinkageVerificationError {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    write!(f, "{:?}; ", self.cause)?;
+    match &self.source {
+      Some(err) => write!(f, "{}", err)?,
+      None => {}
+    }
+    Ok(())
```
PhilippGackstatter, line 167:
```suggestion
    write!(f, "{}", self.cause)
```

Print the `Display` impl of the `cause`, not the `Debug` one.
Don't include the source error in the `Display` impl. Higher-level functionality/crates will do the formatting of these error chains. See also #1017.

identity_credential/src/validator/errors.rs
```diff
@@ -139,4 +140,65 @@ impl Display for CompoundPresentationValidationError {
...
+  pub cause: DomainLinkageVerificationErrorCause,
+  /// Source of the error.
+  pub source: Option<Box<dyn Error + Send + Sync + 'static>>,
+}
+impl DomainLinkageVerificationError {
+  /// Provides the cause of the error.
+  pub fn get_cause(&self) -> &DomainLinkageVerificationErrorCause {
+    &self.cause
+  }
+}
```
PhilippGackstatter, line 158: Is this really needed? The field is already public.

identity_credential/src/validator/errors.rs
```diff
@@ -139,4 +140,65 @@ impl Display for CompoundPresentationValidationError {
   }
 }
 
-impl std::error::Error for CompoundPresentationValidationError {}
+impl Error for CompoundPresentationValidationError {}
+
+#[derive(Debug, thiserror::Error)]
+/// An error caused by a failure to verify a Domain Linkage configuration or credential.
+pub struct DomainLinkageVerificationError {
```
PhilippGackstatter, line 147: Should we mark this `[non_exhaustive]`? Probably not, but just want to make sure. @olivereanderson any opinions?

identity_credential/tests/fixtures/did_configuration/issuer-did-document.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "did:iota:rms:0xe5d5a7c02874b0284ccc2f28bada68780cad26702e7e8186d521e9615a73bc45",
+  "verificationMethod": [
+    {
+      "id": "did:iota:rms:0xe5d5a7c02874b0284ccc2f28bada68780cad26702e7e8186d521e9615a73bc45#key-1",
+      "controller": "did:iota:rms:0xe5d5a7c02874b0284ccc2f28bada68780cad26702e7e8186d521e9615a73bc45",
+      "type": "Ed25519VerificationKey2018",
```
PhilippGackstatter, line 7: Is this file still needed when you use in-memory `CoreDocument`s for tests?

identity_credential/tests/fixtures/did_configuration/config1.json
```diff
@@ -0,0 +1,27 @@
+{
```
PhilippGackstatter, line 1: Consider giving the file a more descriptive name.

PhilippGackstatter (changes requested):
Thanks for all the improvements, much appreciated, it's looking good overall!

A few comments on style, formatting, being good citizens in the Rust API world, errors, docs and other things.

I'm not too keen on including code in comments in the examples. They might go stale very fast, and then we have incorrect examples. My weak opinion is that we should rather describe what is commented.

identity_credential/src/credential/domain_linkage_configuration.rs
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  /// List of the issuers of the Domain Linkage Credentials.
+  pub fn issuers(&self) -> Vec<Url> {
+    self
+      .linked_dids
+      .iter()
+      .map(|linked_did| linked_did.issuer.url().clone())
+      .collect()
+  }
```
olivereanderson, line 125: I think Rust users would typically prefer an iterator to be returned here: 
```suggestion
  pub fn issuers(&self) -> impl Iterator<Item = &Url> {
    self
      .linked_dids
      .iter()
      .map(|linked_did| linked_did.issuer.url())
  }
```

identity_credential/src/validator/errors.rs
olivereanderson: Maybe it is best to instead make the fields private and introduce a getter for `cause` (the `source` is retrievable through the error trait).

identity_credential/src/validator/errors.rs
```diff
@@ -139,4 +140,65 @@ impl Display for CompoundPresentationValidationError {
...
+    }
+    Ok(())
+  }
+}
+
+#[non_exhaustive]
+#[derive(Debug, thiserror::Error, strum::IntoStaticStr)]
+pub enum DomainLinkageVerificationErrorCause {
+  #[error("invalid credential")]
+  CredentialValidationError,
```
olivereanderson, line 175: This should probably wrap `CompoundCredentialValidationError` or something like that to give more information about what went wrong when validating the attached credential.

identity_credential/src/credential/domain_linkage_configuration.rs
```diff
@@ -0,0 +1,160 @@
...
+
+  /// Fetches the the DID Configuration resource via a GET request at the
+  /// well-known location: "`origin`/.well-known/did-configuration.json".
+  #[cfg(feature = "domain-linkage-fetch")]
+  pub async fn fetch_configuration(mut origin: Url) -> Result<DomainLinkageConfiguration> {
+    if origin.scheme() != "https" {
+      return Err(DomainLinkageError("origin` does not use `https` protocol".into()));
+    }
+    // todo: set file size limit to 1MB.
```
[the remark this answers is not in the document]
abdulmth: I couldn't find a standard way to limit the size here with `reqwest` I'm not sure how to do this here, maybe use a stream if `reqwest` supports this?

identity_credential/src/credential/linked_domain_service.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    let service_from_json: Service<CoreDID, Object> = Service::from_json_value(json!({
+        "id":"did:example:123#foo",
+        "type": "LinkedDomains",
+        "serviceEndpoint": {
+          "origins": [domain_1, domain_2]
+        }
+    }))
+    .unwrap();
+    assert_eq!(service.into_service(), service_from_json);
```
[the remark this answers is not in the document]
PhilippGackstatter: If a conversion API is implemented, then implementing `From` and `Into` traits should always be preferred over an inherent function.

identity_credential/src/validator/errors.rs
PhilippGackstatter: You can try `cargo clean`, that sometimes helps with IDE problems.

identity_credential/src/credential/domain_linkage_configuration.rs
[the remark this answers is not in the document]
PhilippGackstatter: I think we can do it this way:

```rust
    // We use a stream so we can limit the size of the response to 1 MiB.
    let mut stream = client
      .get(origin.to_string())
      .send()
      .await
      .map_err(|err| DomainLinkageError(Box::new(err)))?
      .bytes_stream()
      .take(1_048_576);

    let mut json: Vec<u8> = Vec::new();
    while let Some(item) = stream.next().await {
      match item {
        Ok(bytes) => json.extend(bytes),
        Err(err) => return Err(DomainLinkageError(Box::new(err))),
      }
    }

    let domain_linkage_configuration: DomainLinkageConfiguration =
      DomainLinkageConfiguration::from_json_slice(&json).map_err(|err| DomainLinkageError(Box::new(err)))?;
```

identity_credential/src/validator/errors.rs
olivereanderson: From a logging perspective you are totally correct. 

The problem with the vague `CredentialValidation` variant is that one doesn't get an idea of which validations may fail by looking at the error. I am wondering whether it would make sense to either wrap `ValidationError` in the `CredentialValidation` variant (and call `CredentialValidator` with `FailFast::Yes` since domain linkage validation fails on the first error otherwise), or perhaps we include the relevant variants from  `ValidationError` as additional variants here. The latter is a bit more work to program, but would probably be nicer for TS users because more context becomes available with `Error.name`.  

Let me know what you prefer/think and also please free to chime in @PhilippGackstatter.

identity_credential/src/credential/domain_linkage_configuration.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Note:
+/// - Only [Linked Data Proof Format](https://identity.foundation/.well-known/resources/did-configuration/#linked-data-proof-format)
+///   is supported.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(try_from = "__DomainLinkageConfiguration")]
+pub struct DomainLinkageConfiguration(__DomainLinkageConfiguration);
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct __DomainLinkageConfiguration {
```
olivereanderson, line 38: Does this type need to be public?

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,530 @@
...
+    credential: &Credential<T>,
+    domain: Url,
+    validation_options: &CredentialValidationOptions,
+  ) -> DomainLinkageValidationResult {
+    credential
+      .check_structure()
+      .map_err(|err| DomainLinkageValidationError {
+        cause: DomainLinkageValidationErrorCause::CredentialValidationError,
+        source: Some(Box::new(err)),
+      })?;
```
PhilippGackstatter, line 87: This is done in `CredentialValidator::validate` already.

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,530 @@
...
+      }
+    }
+  }
+
+  /// Validates a [Domain Linkage Credential](https://identity.foundation/.well-known/resources/did-configuration/#domain-linkage-credential).
+  ///
+  /// *`issuer`: issuer of the credential.
+  /// *`credential`: domain linkage Credential to be verified.
+  /// *`domain`: the domain hosting the credential.
+  pub fn validate_credential<T: Serialize, DOC: ValidatorDocument>(
```
PhilippGackstatter, line 76: Would be great if you could add some comments in this functions to each section of the validation. Because the function is so long it would be helpful to have a rough overview of what each section validates.
Alternatively, split the function into smaller functions with descriptive names.
(Or whatever you discuss with Oliver).

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,530 @@
...
+        let type_1 = types.get(0).ok_or(DomainLinkageValidationError {
+          cause: DomainLinkageValidationErrorCause::InvalidTypeProperty,
+          source: None,
+        })?;
+        let type_2: &String = types.get(1).ok_or(DomainLinkageValidationError {
+          cause: DomainLinkageValidationErrorCause::InvalidTypeProperty,
+          source: None,
+        })?;
+        let expected_types: HashSet<&str> = HashSet::from([
+          Credential::<T>::base_type(),
```
PhilippGackstatter, line 126: Checking whether this is in the credential is part of `CredentialValidator::check_structure` which is part of `CredentialValidator::validate` already.

PhilippGackstatter (approved):
Looks good to me, thanks a lot for your work on this!

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,530 @@
...
+          });
+        }
+      }
+      OneOrMany::Many(_) => {
+        return Err(DomainLinkageValidationError {
+          cause: DomainLinkageValidationErrorCause::MultipleCredentialSubjects,
+          source: None,
+        });
+      }
+    }
```
olivereanderson, line 219: I find it easier to read this if it is split up a bit more with some comments explaining the intention of each block. Here is a suggestion:

```suggestion
 // Extract credential subject
    let OneOrMany::One(ref credential_subject) = credential.credential_subject else {
      return Err(DomainLinkageValidationError {
        cause: DomainLinkageValidationErrorCause::MultipleCredentialSubjects,
        source: None,
      });
    };

    // Validate credential subject
    {
      let subject_id = credential_subject.id.as_deref().ok_or(DomainLinkageValidationError {
        cause: DomainLinkageValidationErrorCause::MissingSubjectId,
        source: None,
      })?;

      let subject_did = CoreDID::parse(subject_id.as_str()).map_err(|_| DomainLinkageValidationError {
        cause: DomainLinkageValidationErrorCause::InvalidSubjectId,
        source: None,
      })?;

      if issuer_did != subject_did {
        return Err(DomainLinkageValidationError {
          cause: DomainLinkageValidationErrorCause::IssuerSubjectMismatch,
          source: None,
        });
      }
    }

    // Extract and validate origin
    {
      let origin: &str = credential_subject
        .properties
        .get("origin")
        .and_then(|value| value.as_str())
        .ok_or(DomainLinkageValidationError {
          cause: DomainLinkageValidationErrorCause::InvalidSubjectOrigin,
          source: None,
        })?;

      let origin_url: Url = match Url::parse(origin) {
        Ok(url) => Ok(url),
        Err(identity_core::Error::InvalidUrl(url::ParseError::RelativeUrlWithoutBase)) => {
          Url::parse("https://".to_owned() + origin).map_err(|err| DomainLinkageValidationError {
            cause: DomainLinkageValidationErrorCause::InvalidSubjectOrigin,
            source: Some(Box::new(err)),
          })
        }
        Err(other_error) => Err(DomainLinkageValidationError {
          cause: DomainLinkageValidationErrorCause::InvalidSubjectOrigin,
          source: Some(Box::new(other_error)),
        }),
      }?;

      if origin_url.origin() != domain.origin() {
        return Err(DomainLinkageValidationError {
          cause: DomainLinkageValidationErrorCause::OriginMismatch,
          source: None,
        });
      }
    }
```

PhilippGackstatter: I think that's great, thanks for the comments, too!

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,485 @@
...
+          });
+        }
+      }
+      OneOrMany::One(_) => {
+        Err(DomainLinkageValidationError {
+          cause: DomainLinkageValidationErrorCause::InvalidTypeProperty,
+          source: None,
+        })?;
+      }
+    };
```
olivereanderson, line 120: Since `CredentialValidator::validate` already checks for the existence of the `VerifiableCredential` type, this can be simplified to the following:
```suggestion
   // Validate type.
    if !credential.types.iter().any(|type_| type_ == DomainLinkageConfiguration::domain_linkage_type()) {
      return Err(DomainLinkageValidationError {
        cause: DomainLinkageValidationErrorCause::InvalidTypeProperty,
        source: None,
      });
    }
```

identity_credential/src/validator/domain_linkage_validator.rs
```diff
@@ -0,0 +1,485 @@
...
+  /// *`credential`: domain linkage Credential to be verified.
+  /// *`domain`: the domain hosting the credential.
+  pub fn validate_credential<T: Serialize, DOC: ValidatorDocument>(
+    issuer: &DOC,
+    credential: &Credential<T>,
+    domain: Url,
+    validation_options: &CredentialValidationOptions,
+  ) -> DomainLinkageValidationResult {
+    let issuer_did: CoreDID =
+      CoreDID::parse(credential.issuer.url().to_string()).map_err(|err| DomainLinkageValidationError {
```
olivereanderson, line 82: Nit: This will save an allocation (because `CoreDID` is not yet smart enough to re-use the owned string it is given): 
```
CoreDID::parse(credential.issuer.url().as_str()).map_err(|err| DomainLinkageValidationError {
```

olivereanderson (approved):
This looks great! 
Thank you for working on this cool feature :+1: 

I added two more comments that are mostly nits.

Issue #1103, opened 2023-01-13
[Task] `0.7` release tasks. 


olivereanderson:
## Description
At a high level these are the remaining tasks before `0.7` can be released.:

- Replace `JcsEd25519Signature2020` with [VC-JWT](https://w3c.github.io/vc-jwt/).  
Implement an API for securing verifiable data (such as Credentials, Presentations and arbitrary JSON) with JWTs in accordance with the [VC-JWT spec](https://w3c.github.io/vc-jwt/). The aforementioned API must use a key storage implementation and can thus be seen as an Account replacement.   Functionality for more advanced proofs may be added as an addition in the future, but JWT with JWS digital signatures will be the focus for now. 

- Refactor the `CredentialValidator` and `PresentationValidator` to operate with JWTs (and the validators introduced in #1094 after it has been merged).  

- Refactor the library to work better with the new JWT-based APIs and make it simpler to obtain feature parity between the Rust library and the bindings.  

## Motivation
Motivation for replacing `JcsEd25519Signature2020` proofs with JWTs:  Securing verifiable data with JWTs is desirable in order to align with established projects like [EBSI ](https://ec.europa.eu/digital-building-blocks/wikis/display/EBSIDOC/E-signing+and+e-sealing+Verifiable+Credentials+and+Verifiable+Presentations) . It also provides some amount of cryptographic agility to implementers in a relatively simple manner.  Furthermore there are known issues with the `JcsEd25519Signature2020` spec we are currently using (see [#22](https://github.com/decentralized-identity/JcsEd25519Signature2020/issues/22) and [#26](https://github.com/decentralized-identity/JcsEd25519Signature2020/issues/26)). 

## Resources
Link to any resources relevant for the task such as issues, PRs, reference implementations, or specifications.

## To-do list
Create a task-specific to-do list. Please link PRs that match the TODO list item behind the item after it has been submitted.

- [x] Split identity_did crate.
- [x] Create identity_jose from libjose parts. [Philipp]
- [x] Remove DID generic on CoreDocument.
- [x] Prepare `WasmCoreDocument` to be cheaply cloneable (for use with async methods that need `self` with a static lifetime)
- [x] Implement `KeyStorage`
- [x] Implement `KeyIdStorage`
- [x] Implement `publicKeyJwk`
- [x] Implement CoreDocumentExt::{create_vc_jwt,create_vp_jwt} [started in #1148]
- [x] Remove (the now redundant) Document trait. 
[...]

PR #1105, opened 2023-01-23
Add JSON Object Signing and Encryption capabilities

PhilippGackstatter:
# Description of change

This PR introduces JOSE (JSON Object Signing and Encryption) capabilities in the `identity_jose` crate. This is in preparation to adding support for signing credentials with `VC-JWT`. See also #1103 for more details.

The primary source of the implementation is our (unused) `libjose` crate, from which the JWK, JWS and JWT implementations were extracted. JWE is not required for now and so was not added to the new crate. 

The biggest change to the `libjose` crate is how signatures are crated and how they are verified. In particular, `libjose` included signing capabilities for RSA, P256, HMAC and Ed25519 using in-memory keys. Because we want to use our more secure architecture that delegates signing to some KMS (like Stronghold, a cloud KMS or potentially a hardware-secure module), the API was changed accordingly.
The included capabilities were removed to remove the dependency on various cryptographic crates. Instead the `Encoder` and `Decoder` API essentially take closures that are expected to sign and verify, respectively.

A full roundtrip example:

```rust
#[tokio::test]
  async fn test_encoder_decoder_roundtrip() {
    let secret_key = Arc::new(SecretKey::generate().unwrap());
    let public_key = secret_key.public_key();

    let sign_fn = move |protected, unprotected, msg: Vec<u8>| {
      let sk = secret_key.clone();
      async move {
        let header_set: JwtHeaderSet<JwsHeader> = JwtHeaderSet::new().protected(&protected).unprotected(&unprotected);
        if header_set.try_alg().map_err(|_| "missing `alg` parameter")? != JwsAlgorithm::EdDSA {
          return Err("incompatible `alg` parameter");
        }
        let sig: _ = sk.sign(msg.as_slice()).to_bytes();
        Ok(jwu::encode_b64(sig))
      }
    };

    let verify_fn = |protected: Option<&JwsHeader>, unprotected: Option<&JwsHeader>, msg: &[u8], sig: &[u8]| {
      let header_set: JwtHeaderSet<JwsHeader> = JwtHeaderSet::new().protected(protected).unprotected(unprotected);
      if header_set.try_alg().map_err(|_| "missing `alg` parameter")? != JwsAlgorithm::EdDSA {
        return Err("incompatible `alg` parameter");
      }

      let signature_arr = <[u8; crypto::signatures::ed25519::SIGNATURE_LENGTH]>::try_from(sig)
        .map_err(|err| err.to_string())
        .unwrap();

      let signature = crypto::signatures::ed25519::Signature::from_bytes(signature_arr);
      if public_key.verify(&signature, msg) {
        Ok(())
      } else {
        Err("invalid signature")
      }
    };

    let mut header: JwsHeader = JwsHeader::new(JwsAlgorithm::EdDSA);
    header.set_kid("did:iota:0x123#signing-key");

    let mut claims: JwtClaims<serde_json::Value> = JwtClaims::new();
    claims.set_iss("issuer");
    claims.set_iat(
      SystemTime::now()
        .duration_since(SystemTime::UNIX_EPOCH)
        .unwrap()
        .as_secs() as i64,
    );
    claims.set_custom(serde_json::json!({"num": 42u64}));

    let token: String = Encoder::new(sign_fn)
      .recipient(Recipient::new().protected(&header))
      .encode_serde(&claims)
      .await
      .unwrap();

    let token: _ = Decoder::new(verify_fn).decode(token.as_bytes()).unwrap();

    let recovered_claims: JwtClaims<serde_json::Value> = serde_json::from_slice(&token.claims).unwrap();

    assert_eq!(claims, recovered_claims);
  }
```

part of #1103.

## Type of change

- [ ] Documentation Fix

Tests with test vectors from various RFCs are included.

## Change checklist


olivereanderson (changes requested):
Thanks for working on this! 

I understand that it is hard to know exactly what to bring in here as this crate is intended to be a building block for higher level functionality which we have yet to flesh out. Taking that into account I think this is already pretty good! 

What I am missing is some more documentation for public items, perhaps especially in the `jws` module. In fact it would be really nice with some module level documentation for `encoder` and `decoder`. 

I also wish it was a bit easier to get an overview over where the tests for a given function can be found since the tests are not in the same file (in this case understandably so). Maybe the [cov_mark crate](https://docs.rs/cov-mark/1.1.0/cov_mark/) can help with that? (Matklad has written some pretty good blog posts about that crate [here](https://ferrous-systems.com/blog/coverage-marks/) and [here](https://matklad.github.io/2018/06/18/a-trick-for-test-maintenance.html)).

On the architecture side I am not sure whether we want to include the possibility to attach private key parameters in a JWK at all as we want to discourage in-memory keys, but that is perhaps something we can revisit in a later PR. 

This was my first read through. I will go through the tests more carefully tomorrow.

identity_jose/src/jwt/claims.rs
```diff
@@ -0,0 +1,205 @@
...
+  /// Provides a unique identifier for the JWT.
+  ///
+  /// [More Info](https://tools.ietf.org/html/rfc7519#section-4.1.7)
+  #[serde(skip_serializing_if = "Option::is_none")]
+  jti: Option<String>, // JWT ID
+  /// Identifies a decentralized digital identity.
+  ///
+  /// [More Info](https://www.w3.org/TR/did-core/)
+  #[serde(skip_serializing_if = "Option::is_none")]
+  did: Option<String>, // Decentralized Identifier
```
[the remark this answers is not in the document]
PhilippGackstatter: Good catch, probably not worth having if we don't see an immediate use case. Will remove.

identity_jose/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+license = "Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/iotaledger/identity.rs"
+description = "A library for JOSE (JSON Object Signing and Encryption)"
+
+[dependencies]
+base64 = { version = "0.21.0", default-features = false, features = ["std"] }
+iota-crypto = { version = "0.15.3", default-features = false, features = ["std", "sha"] }
+serde = { version = "1.0", default-features = false, features = ["std", "derive"] }
+serde_json = { version = "1.0", default-features = false, features = ["std"] }
```
olivereanderson, line 17: Nit: Perhaps these crates should also go in `workspace.dependencies` since they seem to be fairly common (either as essential dependencies or at least as dev-dependencies).

identity_jose/src/jws/decoder.rs
```diff
@@ -0,0 +1,251 @@
...
+  pub fn payload(mut self, value: &'b [u8]) -> Self {
+    self.payload = Some(value);
+    self
+  }
+
+  /// Decode the given `data` which is a base64url-encoded JWS.
+  ///
+  /// The `verify_fn` closure must be provided to verify signatures on the JWS.
+  /// Only one signature must match in order for the JWS to be considered valid,
+  /// hence `verify_fn` can error on signatures it cannot verify.
```
olivereanderson, line 127: Consider adding a sentence explaining when multiple signatures can and cannot be expected. Also it would be nice to describe in some detail when `verify_fn` must return an error. Something like "verify_fn must return an error if the `alg` parameter in the header describes a cryptographic algorithm `verify_fn` is not designed to handle,  or if signature verification fails".

identity_jose/src/jws/encoder.rs
```diff
@@ -0,0 +1,300 @@
...
+use crate::jws::JwsHeader;
+use crate::jws::Recipient;
+use crate::jwu;
+
+/// The protected JWS header.
+pub type EncoderProtectedHeader = JwsHeader;
+/// The unprotected JWS header.
+pub type EncoderUnprotectedHeader = JwsHeader;
+/// The message to sign as a byte vector.
+pub type EncoderMessage = Vec<u8>;
```
olivereanderson, line 21: Is `Message`  standard JOSE parlance? If not it might be an idea to write a sentence about what to expect here.

olivereanderson (approved):
Thanks for the improved documentation!

I approve this with the understanding that we will iterate heavily upon this crate before our next release as there are still a lot of missing docs and there should be more tests (also for the unhappy path).

identity_jose/src/jws/encoder.rs
PhilippGackstatter: > JSON Web Signature (JWS)
>    A data structure representing a digitally signed or MACed message.
> https://www.rfc-editor.org/rfc/rfc7515#section-2

I think it's common terminology.

identity_jose/Cargo.toml
PhilippGackstatter: Yeah, that makes sense. Once I've merged this PR into main I will make a small PR that hoists `serde_json` into the top-level `Cargo.toml`.

PR #1108, opened 2023-01-26
Replace deprecated workflow functions

eike-hass:
# Description of change
Central workflow actions we use setup our rust tooling in CI have been unmaintained for a while. We need to move away from them since some of their dependencies and assumptions have bee deprecated by GitHub and might cease to work in the future. 
The two noteworthy deprecations are: 

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

Some alternatives (and general dissatisfaction) have been discussed in issues in the actions repositories:

https://github.com/actions-rs/toolchain/issues/221
https://github.com/actions-rs/toolchain/issues/219

Note:
We still need to find a replacement for  [actions-rs/audit-check](https://github.com/actions-rs/audit-check) which runs `cargo-audit` in the background, but also handles GitHub specific reporting.

The changes in this PR address the deprecations in our code and replace the `actions-rs` actions. The toolchain setup is now done through a custom action in all places.

## Type of change

- [X] Bug fix (a non-breaking change which fixes an issue)
- [ ] Documentation Fix

CI passes, although some workflows will only be triggered during a release, so there is some potential for things breaking still.

## Change checklist

### DIFF
--- a/.github/actions/rust/rust-setup/action.yml
+++ b/.github/actions/rust/rust-setup/action.yml
@@ -4,10 +4,9 @@ inputs:
   target:
     description: 'Additionally install specified target for this toolchain, ex. x86_64-apple-darwin'
     required: false
-  profile:
-    description: 'Execute rustup set profile {value} before installing the toolchain, ex. minimal'
+  components:
+    description: 'Comma-separated string of additional components to install e.g. `clippy`, `rustfmt`'
     required: false
-    default: 'minimal'
   os:
     description: 'OS of the runner, used for cache key construction.'
     required: true
@@ -44,12 +43,10 @@ runs:
       run: echo "CURRENT_DATE=$(Get-Date -Format "yyyy-MM-dd")" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
     - name: Install stable toolchain
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@ce8f65846d7180d2ce63b1e74483d981800b9e22
       with:
         toolchain: stable
-        override: true
-        profile: ${{ inputs.profile }}
-        target: ${{ inputs.target }}
+        targets: ${{ inputs.target }}
 
     - name: Cache cargo
       uses: actions/cache@v2.1.7
@@ -72,6 +69,8 @@ runs:
     # Allows dependencies updated on crates.io between runs to trigger storing an updated cache,
     # which hashing Cargo.toml files alone does not.
     - name: Cargo update
+      run: cargo update 
+
       uses: actions-rs/cargo@v1
       with:
         command: update
--- a/.github/actions/rust/rust-setup/action.yml
+++ b/.github/actions/rust/rust-setup/action.yml
@@ -45,11 +45,16 @@ runs:
       run: echo "CURRENT_DATE=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
 
     - name: Install stable toolchain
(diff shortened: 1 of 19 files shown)


.github/actions/rust/rust-setup/action.yml
```diff
@@ -33,26 +40,25 @@ runs:
...
-        override: true
-        profile: ${{ inputs.profile }}
-        target: ${{ inputs.target }}
+        toolchain: ${{ inputs.toolchain }}
+        targets: ${{ inputs.target }}
+        components: ${{ inputs.components }}
+
+    - name: Override toolchain
+      shell: bash
+      run: rustup override set ${{steps.toolchain.outputs.name}}
```
PhilippGackstatter, line 57: Supernit:
```suggestion
      run: rustup override set ${{ steps.toolchain.outputs.name }}
```

.github/workflows/build-and-test.yml
```diff
@@ -117,22 +118,16 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: cargo clean
 
+        # Build the library, tests, and examples without running them to avoid recompilation in the run tests step
       - name: Build all features
```
PhilippGackstatter, line 122:
```suggestion
      - name: Build with all features
```

.github/workflows/clippy.yml
```diff
@@ -20,45 +20,43 @@ jobs:
...
-          profile: minimal
-          toolchain: stable
-          target: wasm32-unknown-unknown
-          override: true
+          os: ${{ runner.os }}
+          job: ${{ github.job }}
+          target: x86_64-unknown-linux-gnu, wasm32-unknown-unknown
           components: clippy
-
+
```
PhilippGackstatter, line 32: Nit: Remove whitespace.

.github/workflows/test-docs-build.yml
```diff
@@ -27,17 +27,15 @@ jobs:
...
       - name: Test Rust Documentation
-        uses: actions-rs/cargo@v1
         env:
           RUSTDOCFLAGS: "-D warnings --cfg docsrs"
-        with:
-          command: doc
-          toolchain: nightly
-          args: --all-features --no-deps --workspace
+          RUSTUP_TOOLCHAIN: "nightly"
+        run: cargo doc --all-features --no-deps --workspace
```
PhilippGackstatter, line 41: Does this definitely run with nightly? Maybe we can use `cargo +nightly doc ...` instead to make sure.

eike-hass: According to the docs it [def](https://rust-lang.github.io/rustup/overrides.html) should work, but will change to override shorthand if that is more idiomatic.

PhilippGackstatter: Alright, then it's fine either way. Thanks!

.github/workflows/format.yml
```diff
@@ -58,4 +58,4 @@ jobs:
           dprint check
 
       - name: cargo-license-template check
-        run: cargo license-template --template .license_template --ignore .license_template_ignore --verbose
+        run: cargo +nightly license-template --template .license_template --ignore .license_template_ignore --verbose
```
PhilippGackstatter, line 61: Probably doesn't matter if run with `nightly` because license-template should be agnostic over the environment, but also shouldn't be a problem.

PhilippGackstatter (approved):
Looks good to me, thanks!

olivereanderson (approved):
Thanks for fixing this!

PR #1110, opened 2023-01-30
Remove generics in `CoreDocument`, `VerificationMethod`, `Service`, `DIDUrl` and `LinkedDomainService`. 

olivereanderson:
# Description of change
This PR simplifies the library by removing the Document trait as well as all generics on `CoreDocument`, `Service`,`VerificationMethod`, `DomainLinkageService` and `DIDUrl`.  As this is a somewhat radical change the reviewers are encouraged to carefully read the justification below and judge if it is sufficient. 

This PR additionally upgrades the `wasm-bindgen` dependency to version `0. 2.84` as we need [the following change](https://github.com/rustwasm/wasm-bindgen/pull/3117).  As a consequence of that I think we can remove `bindings/build/lint::lintPtrNullWithoutFree`, but I will leave that for another PR. 

## Why?
At the time the generic DID parameter was introduced it brought a lot of value as we could remove several casts that required opting out of the safe subset of Rust. The problem with 
generics in the DID document however is that one cannot directly put documents not parameterized with the same types in a container (such as a `Vec` or `slice`). We tried to work around this problem with a `Document` trait and a closely related object safe `ValidatorDocument` trait, but when designing the new `Resolver` it was hard to work with those traits. More specifically I found it almost impossible to implement the Document trait for something like 

```enum DocumentType {SomeDIDType(CoreDocument<SomeDIDType>), OtherDIDType(CoreDocument<OtherDIDType>)}```

 or more realistically `enum DocumentType {Iota(IotaDocument), Core(CoreDocument)}`. This led to the introduction of the `AbstractThreadSafeValidatorDocument` that both has a terrible name and is a very leaky abstraction. This is very much the sort of thing we do not want users to have to deal with. 

One possible way of discarding the `AbstractThreadSafeValidatorDocument` would be to make the `Resolver` only resolve `CoreDocument` (in which case all document types would have to implement `Into<CoreDocument>`). With that approach one would have to use a specific client directly if one wishes to preserve method specific information when resolving. After discussing with the team we arrived at the conclusion that we would prefer not to do that. 

Another possibility would be to focus more on promoting the object safe `ValidatorDocument` trait, but that trait is fairly abstract as is now and such an approach would need more careful consideration. Furthermore it is likely that we want verification functionality to not be a direct concern of the document, but rather another structure such as a `JwtVerifier` where custom cryptographic algorithms can be registered. In other words this seems to be a case where it is a good idea to separate data from functionality. 

The third approach which is what this PR does is to just remove the generics from `CoreDocument` altogether. I have yet to see the rewards of having type enforced method specific DID Urls for a document's verification methods or services. In fact most of the time you would want them to have the same DID component (not just method) as that of the document which is something we cannot enforce in the type system, and there could also be niche cases where you would even want the URLs to refer to other DID methods which the spec allows (this is perhaps more so the case for controllers). If I am overlooking something I would of course be happy to strongly consider closing this PR. 

## What about the non-DID specific parameters in VerificationMethod, Service, etc. 
In practice we have so far found that they are either not used, or a (JSON) `Object` representation is enough. If we should later become aware of (or get requests for supporting) use-cases requiring the ability to inject more specific types into these structures, we can consider re-adding those (non-DID specific) parameters in an almost non-breaking manner (in the case when the same parameters can be used across different document types only though). 

Also note that the generic parameter `T` in the `properties` field in the current `CoreDocument<D,T,U,V>` can be avoided by instead introducing a wrapper 
```
#[Derive(Serialize, Deserialize)]
struct MyDocument {
    #[serde(flatten)]
    doc: CoreDocument,
    #[serde(flatten)]
    my_additional_properties: MyProperties 
} 
```

## What about identifiers one would expect to be method specific? 
A good example of such an identifier is the identifier of a DID document published with the `IOTA` UTXO method specification. Even though represented internally using a `CoreDID`, one will still receive a `&IotaDID` when calling `IotaDocument::id`. Like before this is not expressible in the safe subset of Rust (yet), but this time we wrap this functionality in a safe interface generated by the `ref-cast` crate. A crate created by `dtolnay` with more than five million downloads. Unfortunately the `#![forbid(unsafe_code)]` lint has to be removed from the `identity_iota_core` crate to accommodate for this as the macro from `ref-cast` generates an `unsafe` code block (while checking its soundness). 

Similarly one would expect the controller(s) of an `IotaDocument` to be IOTA DIDs. To accommodate for this the `IotaDocument::controller` method now returns an iterator over the controller(s) and this uses the same trick as mentioned in the previous paragraph to obtain `IotaDID` references (in the bindings there is no change as we need to clone anyway). 

Note that there is (at least to the best of my knowledge) no unchecked way to mutate the identifier of an `IotaDocument` (or its controllers) after construction in the public API hence 
handing out `IotaDID` references to said identifiers should be sound, but I am very much aware that this approach creates more opportunities for bugs to be introduced. 

## How do these changes affect the PresentationValidator? 
The API stays pretty much the same as before except all the `ValidatorDocument` trait bounds are now replaced with `AsRef<CoreDocument>` which should be simpler for Rust developers to understand. 

## Short summary 
The advantages are: 
- Simpler structs (less generics).  
- Simpler trait bounds (`AsRef<CoreDocument>` instead of `ValidatorDocument`).
- Less manual monomorphization in the bindings. 
- The Resolver can now easily be configured to return an enum that both wraps `IotaDocument` and satisfies `AsRef<CoreDocument>`, hence `AbstractThreadSafeValidatorDocument` can be removed. I have some ideas for how to generalize the `Resolver` in the bindings, but that is for another PR. 
- With the ideas we have for JWT verification we probably want to keep reading key material from DID documents, but it is likely better for the verification functionality/logic to reside elsewhere. 
- More methods available directly on the `CoreDocument` / `IotaDocument` without needing to import the `Document` trait (more suggestions from IDEs).
- Less work for users to integrate their custom DID document types with the functionality of our library (provided that the types we provide are sufficient for their use-cases). All that is needed is to implement `AsRef<CoreDocument>`.  

The disadvantages are:
- Increased chances of bugs in/with `IotaDID`.  
- Potentially leads to more runtime checks (But I struggle to come up with a use-case where you extract a verification method or service from an already resolved document and need the DIDUrl of said service or method to be of a certain specialized type for further processing).   
- The `AsRef<CoreDocument>` trait bound forces implementers to build their types using 
`CoreDocument`. This is something a trait abstraction might be able to avoid, but as it is now the `Document` trait still pretty much enforces the use of some types from our library. 

## Open Questions
- Do we want to merge this?
- Should we also remove the `DID` trait and instead bake its methods directly in to `CoreDID` and `IotaDID` so users don't need to import it (the trait is no longer really necessary for polymorphism).
- Is there enough value in returning `&IotaDID` from `IotaDocument::id` (and an iterator of these in the case of `controller`), or would it also be OK to just return `&CoreDID` and we can worry less about accidental bugs breaking user's expectations?

Sub-task of #1103 . 

## Type of change

- [ ] Documentation Fix

Runs under `cargo test`. 

## Change checklist

### CODE REVIEW

File: identity_document/src/document/core_document.rs [hunk 1]
```diff
@@ -148,29 +144,106 @@ impl<D: DID + KeyComparable, T, U, V> CoreDocumentData<D, T, U, V> {
 
     Ok(())
   }
+
+  // Apply the provided fallible functions to the DID components of `id`, `controller`, methods and services
+  // respectively.
+  fn try_update_identifiers<F, G, H, L, E>(
```
PhilippGackstatter, line 150:
```suggestion
  fn try_map<F, G, H, L, E>(
```

- Can you elaborate why `controller_map`, `method_map` and `services_map` are needed and cannot be the same? Is applying different mappings to methods and services a common enough use case to warrant this method allowing for it or should this method rather do just the simplest case of mapping all contained DIDs in the same way?
- Just to make sure: Can this produce invalid results like the former `try_map` could, e.g. produce methods referencing embedded method identifiers, or services with identifiers matching method identifiers? Not really, right? Because we're mapping only the DID not the DIDUrl?

identity_document/src/document/core_document.rs
```diff
@@ -865,12 +807,67 @@ where
...
+    G: FnMut(CoreDID) -> std::result::Result<CoreDID, E>,
+    H: FnMut(CoreDID) -> std::result::Result<CoreDID, E>,
+    L: FnMut(CoreDID) -> std::result::Result<CoreDID, E>,
+    M: FnOnce(crate::Error) -> E,
+  {
+    let data = self
+      .data
+      .try_update_identifiers(id_update, controller_update, methods_update, service_update)?;
+    CoreDocument::try_from(data).map_err(error_cast)
+  }
```
PhilippGackstatter, line 840: Do we absolutely need this method?

identity_document/src/document/core_document.rs
```diff
@@ -865,12 +807,67 @@ where
...
+
+  /// Unchecked version of [Self::try_update_identifiers](Self::try_update_identifiers()).
+  pub fn update_identifiers_unchecked<F, G, H, L>(
+    self,
+    id_update: F,
+    mut controller_update: G,
+    mut methods_update: H,
+    mut service_update: L,
+  ) -> Self
```
PhilippGackstatter, line 849: Same for this method, do we really need it? The API surface of `CoreDocument` is quite large as it is, I'm just wary of adding more convenience methods.

identity_iota_core/src/did/iota_did.rs
```diff
@@ -50,6 +46,18 @@ impl IotaDID {
...
+  /// Convert a `CoreDID` reference to an `IotaDID` reference without checking the referenced value.
+  ///  
+  /// # Warning
+  /// This method should only be called on [`CoreDIDs`](CoreDID) that
+  /// are known to satisfy the requirements of the IOTA UTXO specification.  
+  ///
+  /// # Memory safety
+  ///
+  /// The `ref-cast` crate ensures a memory safe implementation.  
+  pub(crate) const fn from_inner_ref_unchecked(did: &CoreDID) -> &Self;
```
PhilippGackstatter, line 59:
```suggestion
  /// Convert a `CoreDID` reference to an `IotaDID` reference without checking the referenced value.
  ///  
  /// # Warning
  /// This method should only be called on [`CoreDIDs`](CoreDID) that
  /// are known to satisfy the requirements of the IOTA UTXO specification.  
  ///
  /// # Memory safety
  ///
  /// The `ref-cast` crate ensures a memory safe implementation.  
  #[ref_cast_custom]
  pub(crate) const fn from_inner_ref_unchecked(did: &CoreDID) -> &Self;
```

We typically put comments above macros.

PhilippGackstatter (commented):
Preliminary comments, still have to review about half the PR.

identity_document/src/document/core_document.rs
olivereanderson: > Can you elaborate why controller_map, method_map and services_map are needed and cannot be the same? 

In `identity_iota_core::StateMetadataDocument::into_iota_document` this method is called with different maps for `controller_map` (because it does extra checks) than `method_map`. In all of our use-cases `method_map` is always the same as `services_map` however, but there might be methods where this is not the case.  We cannot accommodate for every imaginable use case with a method like this, but I don't think it hurts to include this level of flexibility. 

> Just to make sure: Can this produce invalid results like the former try_map could,

Yes it can (although that would be very much an edge case): If you have a document with a service identified by `did:foo:123#example` and a method identified by `did:bar:123#example` and then you map `did:foo -> did:baz` and also `did:bar -> did:baz`.

identity_document/src/document/core_document.rs
olivereanderson: I tried to remove all of these maps, but since we need to mutate identifiers when resolving an `IotaDocument` we need some way of doing that. I would rather  do it to this than add methods to `OrderedSet` that make it possible to break its invariants.

identity_document/src/document/core_document.rs
olivereanderson: We do use it in `StateMetadataDocument` thus I would prefer to keep it. Would you be open for creating a new module in `identity_document` named (something like) `advanced_api` where we place all of these mapping methods, and maybe also things like `resolve_method_mut`, `id_mut_unchecked` etc. and then we can feature gate that module under an `advanced-api` feature?

identity_document/src/document/core_document.rs
PhilippGackstatter: Fair enough, I'm fine with leaving it as-is (except the name, if you agree :smile:).

identity_document/src/document/core_document.rs
PhilippGackstatter: Ah, sorry the changed files made it look like there were just two versions of this function, but one is actually a private fn on `CoreDocumentData`.
So I'm fine with exposing this in general, but the main point here was whether the error_cast is really necessary? I'd be fine with just requiring `Result<CoreDID, identity_document::Error>` to not add even more generics on that method.

identity_document/src/document/core_document.rs
PhilippGackstatter: The main point is that users can build it themselves by calling `try_update_identifiers`, hence why I call it a convenience method. It doesn't access any private fields that only an inherent method could. 
Would prefer to remove it to keep the API surface somewhat in check, but will leave the decision to you.

identity_iota_core/src/document/iota_document.rs
```diff
@@ -819,4 +863,149 @@ mod tests {
       format!("{{\"doc\":{},\"meta\":{}}}", document.document, document.metadata)
     );
   }
+
+  #[test]
+  fn deserializing_id_from_other_method_fails() {
```
PhilippGackstatter, line 868: Prefix the newly added tests with `test_`.

bindings/wasm/src/iota/iota_document.rs
```diff
@@ -591,20 +584,20 @@ impl From<WasmIotaDocument> for IotaDocument {
...
 
-  #[wasm_bindgen(typescript_type = "IotaService[]")]
-  pub type ArrayIotaService;
+  #[wasm_bindgen(typescript_type = "Service[]")]
+  pub type ArrayService;
 
-  #[wasm_bindgen(typescript_type = "IotaVerificationMethod[]")]
-  pub type ArrayIotaVerificationMethods;
+  #[wasm_bindgen(typescript_type = "VerificationMethod[]")]
+  pub type ArrayVerificationMethods;
```
PhilippGackstatter, line 600: Some of these types (like these two right here) are now duplicated across `iota_document.rs` and `wasm_core_document.rs`, perhaps we can move them into `common/types.rs`?

PhilippGackstatter (approved):
Love this PR! We should have done that a long time ago. More flexibility and simplicity at the cost of a bit less safety. You've nicely pointed out that even the safety isn't really a concern.

identity_iota_core/src/document/iota_document.rs
olivereanderson: Why? I never understood why we do that. From a code point of view it is annotated with "#[test]" so I know it is a test and `cargo test` doesn't exercise anything else (apart from doc tests which are also highlighted as such in the terminal).

PhilippGackstatter: Fair enough, I'm mainly interested in consistency / following conventions here, but I guess it's fine either way. Disregard my comment :)

identity_document/src/document/core_document.rs
olivereanderson: > It doesn't access any private fields that only an inherent method could

It actually does (on the last line) as it returns `CoreDocument` directly and does not go through `CoreDocumentData::try_into::<CoreDocument>`. Users don't have access to this optimization.  I think I would prefer to keep this as is, but I see your point.

bindings/wasm/src/iota/iota_document.rs
olivereanderson: Nice catch!

identity_iota_core/src/document/iota_document.rs
```diff
@@ -456,23 +503,20 @@ mod iota_document_revocation {
...
 }
 
+// TODO: Do we still want something like the commented From implementation below?
+/*
 impl From<(IotaCoreDocument, IotaDocumentMetadata)> for IotaDocument {
   fn from((document, metadata): (IotaCoreDocument, IotaDocumentMetadata)) -> Self {
     Self { document, metadata }
   }
 }
+*/
```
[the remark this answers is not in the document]
olivereanderson: I added a `TryFrom` implementation instead. Should I write in the documentation for `try_from`  something like " this does not check the relationship between the `CoreDocument` and the `IotaDocumentMetadata` " ?

bindings/wasm/examples/src/0_basic/0_create_did.ts
```diff
@@ -53,7 +53,7 @@ export async function createIdentity(): Promise<{
 
     // Insert a new Ed25519 verification method in the DID document.
     let keypair = new KeyPair(KeyType.Ed25519);
-    let method = new IotaVerificationMethod(document.id(), keypair.type(), keypair.public(), "#key-1");
+    let method = new VerificationMethod(document.id().toCoreDid(), keypair.type(), keypair.public(), "#key-1");
```
abdulmth, line 56: can't we add a method directly on `IotaDocument` to get the `CoreDID`? something like `id_core()` to avoid chaining here?

abdulmth (approved):
Looks good to me, I agree with Philipp that this PR is an improvement and I like how simpler everything became. I added minor comments, mostly nits.

bindings/wasm/examples/src/0_basic/0_create_did.ts
olivereanderson: That is a great idea! (also better for performance).

PR #1115, opened 2023-02-08
Add Wasm Bindings for Domain Linkage

abdulmth:
# Description of change
Wasm bindings for the Domain Linkage feature introduced in #1094

## Type of change

- [ ] Documentation Fix


eike-hass:
The new TS example should be added to the tests.

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    // =====================================================
+    // Create Linked Domain service
+    // ====================================================
+
+    let domainFoo = "https://foo.example.com";
+    let domainBar = "https://bar.example.com";
+
+    // Create a Linked Domain Service to enable the discovery of the linked domains through the DID Document.
+    // This is optional
```
eike-hass, line 50: Why is it optional? I'd say either remove or explain.

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
```diff
@@ -0,0 +1,176 @@
...
+    let linkedDomainServices: LinkedDomainService[] = didDocument
+        .service()
+        .filter(service => LinkedDomainService.isValid(service))
+        .map(service => LinkedDomainService.fromService(service));
+
+    // Get the domains included in the Linked Domain Service.
+    let domains: string[] = linkedDomainServices[0].domains();
+
+    // Fetch the DID Configuration resource (For example using the Fetch API).
+    // Note that redirection must be disabled when fetching.
```
eike-hass, line 143: Should mention why and that it is an implementers decision.

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
abdulmth: It means that the specs does not require the DID Document to have a service for correct domain linkage.
I edited the comment.

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
abdulmth: The main issue here was that Fetch API has different implementations for Node.js and the browser. But maybe we can still add this feature in the future. I might create an issue for that.

eike-hass: We should explain why it should be disabled.

abdulmth: I understand you now, It's in the specs, I'll add that.

bindings/wasm/src/credential/domain_linkage_validator.rs
```diff
@@ -0,0 +1,128 @@
...
+    let IValidateCredentialHelper {
+      issuer,
+      credential,
+      domain,
+      validation_options,
+    } = options.into_serde::<IValidateCredentialHelper>().wasm_result()?;
+
+    match issuer {
+      RustSupportedDocument::Core(doc) => {
+        DomainLinkageValidator::validate_credential(&doc, &credential, domain, &validation_options).wasm_result()?;
```
PhilippGackstatter, line 68: There's an inconsistency here between `validate_credential` taking `Url` and `validate_linkage` taking `&Url`. As far as I can tell both can take `&Url` which saves you some clones.

bindings/wasm/src/credential/domain_linkage_validator.rs
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "IValidateLinkage")]
+  pub type IValidateLinkage;
+}
+
+/// Fields for constructing a new {@link LinkedDomainService}.
```
PhilippGackstatter, line 84: Use correct docs.

bindings/wasm/src/credential/domain_linkage_validator.rs
```diff
@@ -0,0 +1,128 @@
...
+  validation_options: CredentialValidationOptions,
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "IValidateCredential")]
+  pub type IValidateCredential;
+}
+
+/// Fields for constructing a new {@link LinkedDomainService}.
```
PhilippGackstatter, line 110: Use correct docs.

bindings/wasm/src/credential/linked_domain_service.rs
```diff
@@ -0,0 +1,95 @@
...
+  #[wasm_bindgen(js_name = fromService)]
+  pub fn from_service(service: &WasmService) -> Result<WasmLinkedDomainService> {
+    Ok(Self(LinkedDomainService::try_from(service.0.clone()).wasm_result()?))
+  }
+
+  /// Returns `true` if a @link{Service} is a valid Linked Domain Service.
+  #[wasm_bindgen(js_name = isValid)]
+  pub fn is_valid(service: &WasmService) -> bool {
+    LinkedDomainService::try_from(service.0.clone()).is_ok()
+  }
```
PhilippGackstatter, line 69: Would remove this due to its inefficiency as mentioned in the example comment.

bindings/wasm/src/credential/linked_domain_service.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+#[wasm_bindgen]
+extern "C" {
+  #[wasm_bindgen(typescript_type = "ILinkedDomainService")]
+  pub type ILinkedDomainService;
+}
+
+/// Fields for constructing a new {@link LinkedDomainService}.
```
PhilippGackstatter, line 78: Use correct docs.

PhilippGackstatter (approved):
Looks good to me, thank you!

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
```diff
@@ -0,0 +1,178 @@
...
+    // But since the DID Configuration
+    // resource isn't available online in this example, we will simply use the JSON.
+    let fetchedConfigurationResource = DomainLinkageConfiguration.fromJSON(configurationResource);
+
+    // Retrieve the issuers of the Domain Linkage Credentials which correspond to the possibly linked DIDs.
+    let issuers: Array<string> = fetchedConfigurationResource.issuers();
+    const issuerDocument: IotaDocument = await didClient.resolveDid(IotaDID.parse(issuers[0]));
+
+    // Validate the linkage between the Domain Linkage Credential in the configuration and the provided issuer DID.
+    // Validation is succeeds when no error is thrown.
```
eike-hass, line 120:
```suggestion
    // Validation succeeds when no error is thrown.
```

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
```diff
@@ -0,0 +1,178 @@
...
+    // Fetch the DID Configuration resource (For example using the Fetch API).
+    // Note that according to the specs, the DID Configuration resource must exist
+    // at the origin's root, Well-Known Resource directory.
+    const _configurationUrl = `${domainFoo}/.well-known/did-configuration.json")`;
+
+    // But since the DID Configuration
+    // resource isn't available online in this example, we will simply use the JSON.
+    let fetchedConfigurationResource = DomainLinkageConfiguration.fromJSON(configurationResource);
+
+    // Retrieve the issuers of the Domain Linkage Credentials which correspond to the possibly linked DIDs.
```
eike-hass, line 115:
```suggestion
    // Retrieve the issuers of the Domain Linkage Credentials which correspond to the possibly linked DIDs.
    // Note that in this example we only validate the first entry in the Credential.  
```

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
```diff
@@ -0,0 +1,178 @@
...
+    // → Case 2: starting from a DID
+    // =====================================================
+
+    const didDocument: IotaDocument = await didClient.resolveDid(did);
+
+    // Get the Linked Domain Services from the DID Document.
+    let linkedDomainServices: LinkedDomainService[] = didDocument
+        .service()
+        .filter(service => LinkedDomainService.isValid(service))
+        .map(service => LinkedDomainService.fromService(service));
```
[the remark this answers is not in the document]
eike-hass: ``type.prototype` doesn't look right

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    // Get the domains included in the Linked Domain Service.
```
eike-hass, line 140:
```suggestion
    // Get the domains included in the Linked Domain Service.
    // Note that in this example we only validate the first entry in the Service.
```

bindings/wasm/examples/src/1_advanced/6_domain_linkage.ts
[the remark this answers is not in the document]
PhilippGackstatter: `service.type` returns a `string[]` of the `type` field in the service, it's not the language-level type concept.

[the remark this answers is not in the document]
PhilippGackstatter: `service.type` could be a getter, no? I mean whatever is more idiomatic.

`fromService` throws an exception on an invalid service that has `"LinkedDomains"` in its `type` field and that's totally fine, because that would be an actual unexcepted error. One should be able to easily implement the `isValid` behavior by try catching around `fromService`.

I would prefer the simple type string check over the `isValid` approach because `fromService` already does that exact structure check so it wastes time by doing it effectively twice.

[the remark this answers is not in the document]
abdulmth: the problem with your approach is that if a service has "linkedDomains" in it's `type` but it's an invalid linked domain service, it will **not** be filtered out and your code will try to convert it to a LinkedDomainService wich will throw an exception and exit, while `isValid` returns a boolean and allows the filtering earlier, preventing any issues when trying to construct a `LinkedDomainService`.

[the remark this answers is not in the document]
PhilippGackstatter: Yes, and I think that's okay. You can still use that to have the isValid behavior by wrapping it in try catch, no?

```typescript
let linkedDomainServices: LinkedDomainService[] = didDocument
        .service()
        .filter(service => service.type.includes("LinkedDomains"))
        .map(service => {
            try {
                LinkedDomainService.fromService(service)
            } catch (ex) {
                // ignore if invalid
            }
        });
```

That doesn't duplicate the structure checking for all services of the document.

In the end I'll leave the decision to you, whatever you think is more idiomatic in TS is fine with me :slightly_smiling_face:.

eike-hass (approved):

PR #1120, opened 2023-02-14
Introduce `IToCoreDocument` and document locks in the bindings. 

olivereanderson:
# Description of change
This PR generalizes the `CredentialValidator`, `PresentationValidator` and `Resolver` in the bindings to support the use of custom DID document types (which is already possible in Rust). 

This PR also introduces locks used internally in `WasmCoreDocument` and `WasmIotaDocument` which we will need when we later introduce the new KMS API. Since the generalization and use of locks affect each other I decided to do both in the same PR in order to get an idea of how that combination might work. 

## The new `IToCoreDocument` interface
This is the typescript analogue of `AsRef<CoreDocument>` that we use in Rust. It just requires implementers to add a method `toCoreDocument` on their document type which gives a representation of the underlying DID document as the `CoreDocument` type we ship.

 If one builds a type by inheriting from `CoreDocument` this interface is automatically implemented (and will actually not even be called by the Identity framework in that case).  Implementers may however in some cases prefer to restrict access to their document type in which case extending `CoreDocument` is less desirable. In that case one may choose to instead implement an  `toCoreDocument` that clones the underlying `CoreDocument` (or otherwise produces it somehow). This is essentially the case for `IotaDocument`.  See the mocks in `tests/credential.ts` to get a more concrete idea of how implementers can implement `IToCoreDocument`. 

## Custom TS functions to avoid expensive clones 
Some custom TS functions for internal use are added in the `lib` folder which are called internally by Rust. This allows us to both get a cheap copy of a document type and avoid nulling out pointers passed from TS. 

## Locks 
For now `WasmCoreDocument` and `WasmIotaDocument` wrap `Rc<tokio::sync::RwLock>`. I used `tokio::sync::RwLock` because it was the only lock I  could find with both an `async` and blocking API. The locks are currently used in such a way that it should be convenient to swap out the lock with another one in the future if desired.  

Care needs to be taken however once we introduce async methods on `WasmCoreDocument` and `WasmIotaDocument` in order to provide the thread from being blocked forever. This can happen with `Promise.all` if a user introduces an async function that blockingly reads (resp. writes) to a document while a write guard (resp. read guard) is being held over an await point.  For this reason it might be a good idea to use `try_write` instead of `blocking_write` and make the non-asynchronous functions that need `&mut` fallible. 

If it turns out we need to hold write guards across await points as well, then we might consider making every method fallible and use `RefCell` instead of `tokio::sync::RwLock` for better performance.  

## ImportedDocumentLock 
This is a new type used internally that can be thought of as an imported document from TS that "implements" `AsRef<CoreDocument>` , but with a few caveats. All calls to the custom TS functions mentioned above are made by this type.    

The `ImportedDocumentLock` type is used by the `CredentialValidator` and `PresentationValidator` and allows us to work with any document type passed in from TS as long as it implements `IToCoreDocument`. It uses reference counting for cheap clones, but that also means we need to remember to drop it in order to prevent memory leaks. When working with arrays it might require extra allocation for the read guards, but that should not be expensive in comparison to credential/presentation validation. In theory those allocations could maybe be avoided by switching to `Arc<tokio::sync::RwLock>` (because we can then call [this method](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#method.try_read_owned)), or using something like the [self_cell crate](https://crates.io/crates/self_cell), but I doubt it is worth it. 

Finally the usage of `ImportedDocumentLock` could be made much nicer (and perhaps even more secure from memory leaks) if we wrote the Resolver entirely in typescript. That would also provide a nicer API for TS users and it avoids the necessity of removing some of the `AsRef<CoreDocument>` bounds one might expect in the Rust resolver that had to be done in this PR in order to accommodate for the bindings. Furthermore we could then also remove one of the generics on the Rust Resolver by requiring all handlers to provide `Send` futures thus simplifying that API. For these reasons I would recommend porting the Resolver to TS (while keeping the Rust Resolver in the Rust library of course). 

## Open questions: 
- What should we do about the possibility of TS users accidentally blocking the thread once we introduce async methods on the document(s)? (This can also be answered when that time comes). 
- Should we proceed with writing the resolver for the bindings entirely in TS? (this can also be done in a later PR and we can then also clean up other things introduced here).

Sub-task of #1103 .

## Type of change

- [ ] Documentation Fix

Tests succeed locally. 

## Change checklist


olivereanderson:
Need to fix the examples.

bindings/wasm/src/did/wasm_core_document.rs
```diff
@@ -452,12 +508,26 @@ impl WasmCoreDocument {
...
+    let guard = self.0.blocking_read();
+    let signer = guard.signer(&private_key);
     let signer = signer.options(options);
     let signer = signer.method(&method_query);
     signer.sign(&mut data).wasm_result()?;
     JsValue::from_serde(&data).wasm_result()
   }
+
+  #[wasm_bindgen(js_name = clone)]
+  /// Deep clones the `CoreDocument`.
```
PhilippGackstatter, line 520:
```suggestion
  /// Deep clones the `CoreDocument`.
  #[wasm_bindgen(js_name = clone)]
```

bindings/wasm/src/did/wasm_core_document.rs
```diff
@@ -452,12 +508,26 @@ impl WasmCoreDocument {
...
+  pub fn deep_clone(&self) -> WasmCoreDocument {
+    WasmCoreDocument(Rc::new(CoreDocumentLock::new(self.0.blocking_read().clone())))
+  }
+
+  /// ### Warning
+  /// This is for internal use only do not call this method.
```
PhilippGackstatter, line 526:
```suggestion
  /// This is for internal use only. Do not rely on or call this method.
```

bindings/wasm/src/did/wasm_core_document.rs
```diff
@@ -513,11 +583,45 @@ struct ICoreDocumentHelper {
...
+  /// Deserializes an instance from a JSON object.
+  #[wasm_bindgen(js_name = fromJSON)]
+  pub fn from_json(json: &JsValue) -> crate::error::Result<WasmCoreDocument> {
+    use crate::error::WasmResult;
+    json
+      .into_serde()
+      .map(|value| Self(Rc::new(CoreDocumentLock::new(value))))
+      .wasm_result()
+  }
+}
```
PhilippGackstatter, line 604: Can you add this to the impl block above and use the conventional `WasmResult` import?

bindings/wasm/src/did/wasm_core_document.rs
```diff
@@ -513,11 +583,45 @@ struct ICoreDocumentHelper {
...
+  #[wasm_bindgen(typescript_type = "CoreDocument | IAsCoreDocument")]
+  pub type IAsCoreDocument;
+
+  #[wasm_bindgen(typescript_type = "Array<CoreDocument | IAsCoreDocument>")]
+  pub type ArrayIAsCoreDocument;
+}
+
+#[wasm_bindgen(typescript_custom_section)]
+pub const TS_AS_REF_CORE_Document: &'static str = r#"
+interface IAsCoreDocument {
```
PhilippGackstatter, line 623: I think the more conventional way to call this would be `IToCoreDocument` in the JS/TS world, as everything is `to` there. This feels a bit like leaking Rust terminology to TS. But not strongly opposed to it and if this results in a ton of renaming then let's leave it as-is.

bindings/wasm/src/iota/iota_document.rs
```diff
@@ -566,22 +614,66 @@ impl WasmIotaDocument {
...
-  fn from(document: IotaDocument) -> Self {
-    Self(document)
+  #[wasm_bindgen(js_name = clone)]
+  /// Returns a deep clone of the `IotaDocument`.
+  pub fn deep_clone(&self) -> WasmIotaDocument {
+    WasmIotaDocument(Rc::new(IotaDocumentLock::new(self.0.blocking_read().clone())))
+  }
+
+  /// ### Warning
+  /// This is for internal use only do not call this method.
```
PhilippGackstatter, line 635:
```suggestion
  /// This is for internal use only. Do not rely on or call this method.
```

bindings/wasm/src/resolver/mod.rs
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 mod constructor_input;
-mod supported_document_types;
+mod specialized_ts_type_definitions;
```
PhilippGackstatter, line 5: Suggestion.
```suggestion
mod type_definitions;
```

bindings/wasm/src/resolver/wasm_resolver.rs
```diff
@@ -237,26 +230,53 @@ impl WasmResolver {
...
+      };
+
+      let (holder_result, issuers_result) = futures::join!(holder_handler(), issuers_handler());
+
+      let holder = holder_result.wasm_result()?;
+      let issuers = issuers_result.wasm_result()?;
+
+      let holder_guard = holder.blocking_read();
+      let issuers_guard: Vec<ImportedDocumentReadGuard<'_>> =
+        issuers.iter().map(ImportedDocumentLock::blocking_read).collect();
```
olivereanderson, line 277: Perhaps these should be changed to the async read versions since we are in an async context. Only worth it if we decide not to port the resolver to TS though.

PhilippGackstatter: > Only worth it if we decide not to port the resolver to TS though.

What do you mean here? We are already porting the resolver to TS, aren't we?

PhilippGackstatter (approved):
Looks good to me. It's a nice hack you came up with!

bindings/wasm/src/did/wasm_core_document.rs
```diff
@@ -40,23 +45,40 @@ use proc_typescript::typescript;
...
+    Self(tokio::sync::RwLock::new(input))
+  }
+
+  pub(crate) fn blocking_read(&self) -> impl Deref<Target = CoreDocument> + '_ + Into<ImportedDocumentReadGuard<'_>> {
+    self.0.blocking_read()
+  }
+
+  pub(crate) fn blocking_write(&self) -> impl DerefMut<Target = CoreDocument> + '_ {
+    self.0.blocking_write()
+  }
```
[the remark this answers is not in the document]
olivereanderson: So that it will be (relatively) easier to swap out the lock later if we want to. I have tried to hide the lock implementation as much as possible (also internally).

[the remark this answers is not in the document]
PhilippGackstatter: Using the concrete type makes reading the other lock stuff easier so I would prefer that. I don't think it'd be very hard to swap the lock.

bindings/wasm/src/resolver/wasm_resolver.rs
olivereanderson: I think it would be valuable to write a resolver in TS instead of using `WasmResolver`.  Note that almost all the methods on `WasmResolver` have quite a bit of code and it isn't as simple as just forwarding to the Rust implementation as is the case with most other types in the bindings. Since there are several other advantages with that approach on both the Rust and TS side I would recommend it.

olivereanderson: Aside/Regardless: I've changed the locking to use the async versions here now.

bindings/wasm/src/did/wasm_core_document.rs
```diff
@@ -452,12 +508,49 @@ impl WasmCoreDocument {
...
+  #[wasm_bindgen(js_name = _shallowCloneInternal)]
+  pub fn shallow_clone(&self) -> WasmCoreDocument {
+    WasmCoreDocument(self.0.clone())
+  }
+
+  // ===========================================================================
+  // Serialization
+  // ===========================================================================
+
+  /// Serializes this to a JSON object.
```
eike-hass, line 540: technically it a JS object, but the `toJSON()` is a bingen IIRC

bindings/wasm/tests/credentials.ts
```diff
@@ -169,7 +218,7 @@ describe("CredentialValidator, PresentationValidator", function() {
             const issuerDoc = new IotaDocument("iota");
             const issuerKeys = new KeyPair(KeyType.Ed25519);
             issuerDoc.insertMethod(
-                new VerificationMethod(issuerDoc.id().toCoreDid(), KeyType.Ed25519, issuerKeys.public(), "#iss-0"),
+                new VerificationMethod(issuerDoc.id().asCoreDid(), KeyType.Ed25519, issuerKeys.public(), "#iss-0"),
```
eike-hass, line 221: Could we avoid the need to cast here completely?

olivereanderson: Yes. I will remove the cast.

bindings/wasm/src/did/wasm_core_document.rs
olivereanderson: Would it be better with "Serializes this to JSON"?  (or perhaps JSON string?)

bindings/wasm/src/did/wasm_core_document.rs
olivereanderson: @eike-hass Would like your input on this one.

bindings/wasm/src/did/wasm_core_document.rs
eike-hass: It does neither of those, that is the problem. JSON is the stringified format, but in this case it's plain JS objects.

eike-hass: So we could keep the method name (to align with bindgen), but document it as `Serializes to plain JS representation` or similar.

bindings/wasm/src/did/wasm_core_document.rs
eike-hass: The `to` notation is mostly introduced by bindgen, but I think we have precedence for `to` then `as` so let'S go with what is more common in our codebase.

eike-hass (approved):

PR #1133, opened 2023-02-28
Split (JWS) `Decoder` functionality.  

olivereanderson:
# Description of change

## Main change 
This PR introduces a stronger separation between JWS decoding and signature verification. In other words decoding and verifying a JWS (signature) is now done in two steps rather than one. 

This is necessary for the Resolver as it needs to partially decode the JWS to find the DIDs of issuers and resolve their documents before signature verification can take place. This separation of concerns also enables use cases where the JWK intended for verification must be extracted from the JWS header. 

 The new verification API also provides a simpler way to pass a `Jwk` to the verification logic than previously in order to provide a better user experience when working with the higher level Credential and Presentation validator APIs. 
  
## Further details
With regards to custom signature verification we now use a trait `JwsSignatureVerifier` instead of a simple closure. The main reason for this is that although it is probably fine to pass a verifier when calling the `Decoder` , the `validate` method on the `PresentationValidator` already takes many arguments and it is likely that one will always want to use the same `JwsSignatureVerifier` there. Hence it is likely we will want to make the new  `PresentationValidator` stateful and introduce a generic parameter for the verifier.

Since closures cannot be named the default for the generic parameter would have to be a function pointer or a boxed `dyn Fn`. The former of these will not work with the bindings and they both introduce a potential drop in performance.  Hence it is better to use a trait which also interacts better with documentation.  Having said that I have still added a struct that wraps a closure to somewhat preserve the ergonomics of the previous closure-based API.

## Changes to `Decoder` configuration 
This PR also shifts more towards static configuration by splitting the `decode` method into `decode_compact_serialization`, `decode_general_serialization` and `decode_flattened_serialization` instead of figuring out which format to expect based on a config. 

The reason for this change was mainly to accommodate for separating the decoding and verification process without affecting performance too much.   

## Alternative handler based design 
With regards to custom signature verification I also considered a design more similar to the `Resolver` where users register algorithm specific handlers that the decoder dispatches verification calls to depending on the algorithm in the JOSE header.  

The advantage of that approach is that it is easier to build on the defaults (for instance a user can add their own handler for `HS256` while keeping the default handler for `EdDSA` in the "mapping"). This alternative approach is not zero cost, but the overhead of the hashmap lookups and pointer indirection is probably fairly low. On the other hand I would rather spend time on strengthening our default implementation by adding support for more algorithms that are both popular and still considered safe to use  so there is less of a need for custom implementations in the first place. 

## Open questions 
- Does the `Decoder` need to be ported to the bindings, or is it enough to just port the `JwsSignatureVerifier` trait as an interface? (I would prefer to keep the `Decoder` as a low level construct in Rust). 
- Is the handler based design outlined above preferable? 

Sub-task of #1103. 

## Type of change

- [ ] Documentation Fix

All tests run locally. 

## Change checklist


olivereanderson:
I am starting to think that the `jwk_provider` argument in `Decoder::decode` should be adapted to decide itself whether to extract the Jwk from the header or provide one of "its own".

identity_jose/src/error.rs
```diff
@@ -27,10 +27,16 @@ pub enum Error {
   InvalidContent(&'static str),
   #[error("invalid key format for type `{0}`")]
   KeyError(&'static str),
-  #[error("unsupported jws algorithm")]
-  UnsupportedAlgorithm,
+  #[error("mismatch between header and jwk alg values")]
+  AlgorithmMismatch,
+  #[error("attempt to parse an unregistered jws algorithms")]
```
olivereanderson, line 32:
```suggestion
  #[error("attempt to parse an unregistered jws algorithm")]
```

identity_jose/src/jws/custom_verification/jws_verifier.rs
```diff
@@ -0,0 +1,176 @@
...
+  fn verify(&self, input: VerificationInput<'_>, public_key: &Jwk) -> Result<(), SignatureVerificationError>;
+}
+
+// =================================================================================================================
+// Implementation
+// ================================================================================================================
+/// Simple wrapper around a closure capable of verifying a JWS signature. This wrapper implements
+/// [`JwsSignatureVerifier`].
+///
+/// Note: One can convert a closure to this wrapper using the [`From`].
```
olivereanderson, line 64:
```suggestion
/// Note: One can convert a closure to this wrapper using the [`From`] trait.
```

identity_jose/src/jws/decoder.rs
```diff
@@ -65,189 +188,277 @@ struct Flatten<'a> {
...
+  /// Append values to the list of permitted extension parameters.
   pub fn critical(mut self, value: impl Into<String>) -> Self {
     self.crits.get_or_insert_with(Vec::new).push(value.into());
     self
   }
 
-  pub fn payload(mut self, value: &'b [u8]) -> Self {
-    self.payload = Some(value);
-    self
+  /// Decode a JWS encoded with the [JWS compact serialization](https://www.rfc-editor.org/rfc/rfc7515#section-3.1)
```
olivereanderson, line 212:
```suggestion
  /// Decode a JWS encoded with the [JWS compact serialization format](https://www.rfc-editor.org/rfc/rfc7515#section-3.1)
```

identity_jose/src/jws/decoder.rs
```diff
@@ -65,189 +188,277 @@ struct Flatten<'a> {
...
+      signature: parse_utf8(signature)?,
+    };
+
+    let payload = Self::expand_payload(detached_payload, Some(payload))?;
+
+    self.decode_signature(payload, signature)
   }
 
-  /// Decode the given `data` which is a base64url-encoded JWS.
+  /// Decode a JWS encoded with the [flattened JWS JSON serialization](https://www.rfc-editor.org/rfc/rfc7515#section-7.2.2)
```
olivereanderson, line 245:
```suggestion
  /// Decode a JWS encoded with the [flattened JWS JSON serialization format](https://www.rfc-editor.org/rfc/rfc7515#section-7.2.2)
```

identity_jose/src/jws/decoder.rs
```diff
@@ -65,189 +188,277 @@ struct Flatten<'a> {
...
+      .signatures
+      .next()
+      .map(|signature| self.decoder.decode_signature(self.payload, signature))
+  }
+}
 
-  fn check_alg(&self, value: JwsAlgorithm) -> Result<()> {
-    check_slice_param("alg", self.algs.as_deref(), &value)
+impl Decoder {
+  /// Decode a JWS encoded with the [general JWS JSON serialization](https://www.rfc-editor.org/rfc/rfc7515#section-7.2.1)
```
olivereanderson, line 335:
```suggestion
   /// Decode a JWS encoded with the [general JWS JSON serialization](https://www.rfc-editor.org/rfc/rfc7515#section-7.2.1)
```

identity_jose/src/jwt/header_set.rs
```diff
@@ -75,16 +75,29 @@ impl<'a, T> JwtHeaderSet<'a, T> {
...
     self.unprotected = value.into();
     self
   }
 
+  /// Get the unprotected header if it is set.
+  pub fn unprotected(&self) -> Option<&'a T> {
+    self.unprotected
+  }
+
+  // TODO: When is this used?
```
PhilippGackstatter, line 100: Not sure, we can probably remove it.

identity_jose/Cargo.toml
```diff
@@ -21,9 +21,20 @@ thiserror.workspace = true
 url = { version = "2.3.1", default-features = false, features = ["serde"] }
 zeroize = { version = "1.5.7", default-features = false, features = ["std", "zeroize_derive"] }
 
+[features]
+# Enables static method `verify_eddsa_jws_prechecked_alg`
+eddsa = ["iota-crypto/ed25519"]
+# Enables construction of the default JWS `Decoder`.
+default-jws-signature-verifier = ["eddsa"]
```
PhilippGackstatter, line 28: Can we not make a dedicated `EdDSAJwsSignatureVerifier` and gate that behind the `eddsa` feature. Future default implementations could add their own feature, so we wouldn't need `default-jws-signature-verifier`.

PhilippGackstatter (approved):
Looks good to me, thank you for splitting the decoding and verification!

identity_jose/src/jws/decoder.rs
```diff
@@ -6,41 +6,164 @@ use std::borrow::Cow;
...
+      .verify(input, public_key)
+      .map_err(Error::SignatureVerificationError)?;
+
+    // Construct token
+    let token: Token = {
+      let JwsValidationItem { headers, claims, .. } = self;
+      let (protected, unprotected): (JwsHeader, Option<Box<JwsHeader>>) = match headers {
+        DecodedHeaders::Protected(protected) => (protected, None),
+        DecodedHeaders::Both { protected, unprotected } => (protected, Some(unprotected)),
+        DecodedHeaders::Unprotected(_) => unreachable!("already checked the existence of protected header"),
```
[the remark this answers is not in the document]
olivereanderson: One can have both a protected and unprotected header, the point is that the code checks for `alg` which has to be in the protected header hence in this case there can not only be an unprotected header. I will think of a better way to state that concisely.

abdulmth (approved):

Issue #1126, opened 2023-02-16
[Task] `KeyIdStorage` trait 


olivereanderson:
## Description
We need an `IdentityStorage` trait to enable custom plug-ins for storing `KeyId`s generated by a `KeyStorage` (ongoing work in [the storage branch](https://github.com/iotaledger/identity.rs/blob/feat/storage)).

The trait might look something like this: 

```
async trait IdentityStorage {

   /// Insert a [`KeyId`] into the [`IdentityStorage`] under the given `key`. If there already exists an 
   /// entry for `key` in the storage an error must be returned immediately without altering the state 
   /// of the storage. 
   async fn insert_key_id(&self, key: MethodDigest, value: KeyId) -> IdentityStorageResult<()>;

   /// Obtain the `KeyId` associated with the given `key`. 
   async fn get_key_id(&self, key: &MethodDigest) -> IdentityStorageResult<KeyId>;

   /// Delete the `KeyId` associated with the given `key` from the [`IdentityStorage`].   
   async fn delete_key_id(&self, key: &MethodDigest) -> IdentityStorageResult<()>;  
}
``` 

The `MethodDigest` will be a new type uniquely describing key material present in a verification method. There should be a method on `VerificationMethod` returning its `MethodDigest`. Internally `MethodDigest` might look something like this: 
```
pub struct MethodDigest {
    version: u8, 
    value: u64
}

impl AsRef<[u8]> for MethodDigest {
  ...
} 
```
For now the value may be `Hash(<VerificationMethod fragment> + <VerificationMethod.jwk_thumbprint()>)` where `Hash` could be [Seahash](https://docs.rs/seahash/4.1.0/seahash/) (since hash collisions are detected at insertion time a cryptographic hash function is not necessary provided that applications properly guard how methods are inserted into document instances).  If we want to change the properties of `MethodDigest` in the future the `version` field can be made public and one should be able to set which version one wants to work with in the `Storage`. 

 
## Motivation
Higher level methods on the document types will call into `KeyStorage` when creating new verification methods which will then produce a key which can be looked up with the corresponding `KeyId`. We also want to provide a method to sign with the key corresponding to a public key present in a verification method in the given document. For this we need a way to map a verification method to the `KeyId` used by the `KeyStorage`. The `IdentityStorage` trait enables us to persist this mapping.   

## Resources
Sub-task of #1103. 

## To-do list
[...]

PR #1134, opened 2023-03-01
implement `KeyIdStorage` in Rust

abdulmth:
# Description of change
Add the `KeyIdStorage` trait as a way for saving `KeyId`s of verification method materials by introducing `MethodDigest` that takes a `VerificationMethod`.  This is an implementation of the issue https://github.com/iotaledger/identity.rs/issues/1126 with the difference that `IdentityStorage` is named here as `KeyIdStorage` (to be discussed). 

`KeyStorageError` is now refactored as generic `StorageError` for both `KeyStorage` and `KeyIdStorage`. 

`fixes issue #1126`.

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/identity_storage/src/error.rs
+++ b/identity_storage/src/error.rs
@@ -1,0 +1,155 @@
+// Copyright 2020-2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::borrow::Cow;
+use std::error::Error;
+use std::fmt::Debug;
+use std::fmt::Display;
+
+/// The error type for key storage operations.
+///
+/// Instances always carry a corresponding [`StorageErrorKind`] and may be extended with custom error messages and
+/// source.
+#[derive(Debug)]
+pub struct StorageError<T: StorageErrorKind> {
+  repr: Repr<T>,
+}
+
+pub trait StorageErrorKind: Display + Debug {
+  fn description(&self) -> &str;
+}
+
+impl<T: StorageErrorKind> Display for StorageError<T> {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    match self.repr {
+      Repr::Simple(ref cause) => write!(f, "{}", cause.description()),
+      Repr::Extensive(ref extensive) => {
+        write!(f, "{}", extensive.cause.description())?;
+        let Some(ref message) = extensive.message else {return Ok(())};
+        write!(f, " message: {}", message.as_ref())
+      }
+    }
+  }
+}
+
+impl<T: StorageErrorKind> Error for StorageError<T> {
+  fn source(&self) -> Option<&(dyn Error + 'static)> {
+    self.extensive().and_then(|err| {
+      err
+        .source
+        .as_ref()
+        .map(|source| source.as_ref() as &(dyn Error + 'static))
+    })
+  }
+}
+
+#[derive(Debug)]
+struct Extensive<T: StorageErrorKind> {
+  cause: T,
+  source: Option<Box<dyn Error + Send + Sync + 'static>>,
+  message: Option<Cow<'static, str>>,
+}
+
+#[derive(Debug)]
+enum Repr<T: StorageErrorKind> {
+  Simple(T),
+  Extensive(Box<Extensive<T>>),
+}
+
+impl<T: StorageErrorKind> From<T> for StorageError<T> {
+  fn from(cause: T) -> Self {
+    Self::new(cause)
+  }
+}
+
+impl<T: StorageErrorKind> From<Box<Extensive<T>>> for StorageError<T> {
+  fn from(extensive: Box<Extensive<T>>) -> Self {
+    Self {
+      repr: Repr::Extensive(extensive),
+    }
+  }
+}
+
+impl<T: StorageErrorKind> StorageError<T> {
+  /// Constructs a new [`StorageError`].  
+  pub fn new(cause: T) -> Self {
+    Self {
+      repr: Repr::Simple(cause),
+    }
+  }
+
+  /// Returns a reference to corresponding [`StorageErrorKind`] of this error.
+  pub fn kind(&self) -> &T {
+    match self.repr {
+      Repr::Simple(ref cause) => cause,
+      Repr::Extensive(ref extensive) => &extensive.cause,
+    }
+  }
+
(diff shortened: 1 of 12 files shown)


identity_storage/src/error.rs
```diff
@@ -0,0 +1,155 @@
...
+use std::error::Error;
+use std::fmt::Debug;
+use std::fmt::Display;
+
+/// The error type for key storage operations.
+///
+/// Instances always carry a corresponding [`StorageErrorKind`] and may be extended with custom error messages and
+/// source.
+#[derive(Debug)]
+pub struct StorageError<T: StorageErrorKind> {
```
olivereanderson, line 14: Using generics here is a good idea! 

It might even make sense to rename this to `CustomError<T>` and move it to `identity_core` so it can be used for all errors users need to return in their code. Not a requirement for this PR.

olivereanderson (approved):
This is good! 

Wrote a few comments and suggestions:

olivereanderson:
With regards to the name of the trait I think we can decide on that later.

identity_storage/src/key_id_storage/method_digest.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2020-2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use identity_verification::VerificationMethod;
+use seahash::SeaHasher;
+use std::hash::Hasher;
+
+/// Unique description through hashing of key materials in `VerificationMethod`s.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct MethodDigest {
```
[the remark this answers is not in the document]
PhilippGackstatter: Yeah if we want to be opaque about the contents, then we can do something like in `StateMetadataDocument` and add a pack and unpack method: https://github.com/iotaledger/identity.rs/blob/cbc2309fe254b4d4f1aba78d3f19e0e1eed88771/identity_iota_core/src/state_metadata/document.rs#L77

So the byte layout would be `<version_byte><value_little_endian_bytes>`.

identity_storage/src/key_id_storage/method_digest.rs
```diff
@@ -0,0 +1,35 @@
...
+  /// Version of hashing.
+  pub version: u8,
+  /// Hash value.
+  pub value: u64,
```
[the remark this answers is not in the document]
PhilippGackstatter: Agreed. Let's make the fields private.

identity_storage/src/key_id_storage/method_digest.rs
```diff
@@ -0,0 +1,35 @@
...
+    let method_data: Vec<u8> = verification_method.data().try_decode()?;
+    hasher.write(fragment.as_bytes());
+    hasher.write(&method_data);
+    let key_hash: u64 = hasher.finish();
+    Ok(Self {
+      version: 0,
+      value: key_hash,
+    })
+  }
+}
```
PhilippGackstatter, line 35: Let's add a test case here in the style of this one: https://github.com/iotaledger/identity.rs/blob/support/v0.6/identity_account_storage/src/types/key_location.rs#L106

Except you can just create a hardcoded verification method, create its digest resulting in value `X`, then add a test ensuring that the hardcoded verification method always hashes to `X`. That same method should result in the same `MethodDigest` in a Wasm test. I think we got rid of `wasm.rs` in `bindings/wasm/tests`, but it would be great if we could test that.

identity_storage/src/key_id_storage/method_digest.rs
```diff
@@ -0,0 +1,35 @@
...
+  pub version: u8,
+  /// Hash value.
+  pub value: u64,
+}
+
+impl MethodDigest {
+  /// Creates a new [`MethodDigest`].
+  pub fn new(verification_method: &VerificationMethod) -> identity_verification::Result<Self> {
+    let mut hasher: SeaHasher = SeaHasher::new();
+    let fragment = verification_method
```
PhilippGackstatter, line 21:
```suggestion
    let fragment: &str = verification_method
```

identity_storage/src/key_id_storage/memstore.rs
```diff
@@ -0,0 +1,114 @@
...
+#[cfg_attr(not(feature = "send-sync-storage"), async_trait(? Send))]
+#[cfg_attr(feature = "send-sync-storage", async_trait)]
+impl KeyIdStorage for KeyIdMemstore {
+  async fn insert_key_id(&self, key: MethodDigest, value: KeyId) -> KeyIdStorageResult<()> {
+    let mut key_id_store: RwLockWriteGuard<'_, KeyIdStore> = self.key_id_store.write().await;
+    key_id_store.insert(key, value);
+    Ok(())
+  }
+
+  /// Obtain the `KeyId` associated with the given `key`.
```
PhilippGackstatter, line 49: You don't have to document trait methods in the impl block, since it's already documented on the trait definition itself.

identity_storage/src/key_id_storage/memstore.rs
```diff
@@ -0,0 +1,114 @@
...
+    Self::new()
+  }
+}
+
+#[cfg_attr(not(feature = "send-sync-storage"), async_trait(? Send))]
+#[cfg_attr(feature = "send-sync-storage", async_trait)]
+impl KeyIdStorage for KeyIdMemstore {
+  async fn insert_key_id(&self, key: MethodDigest, value: KeyId) -> KeyIdStorageResult<()> {
+    let mut key_id_store: RwLockWriteGuard<'_, KeyIdStore> = self.key_id_store.write().await;
+    key_id_store.insert(key, value);
```
PhilippGackstatter, line 45: This doesn't return an error if the key already exists, which is different from the requirements described on `KeyIdStorage::insert_key_id`. Adding a test case for this would help catch this error.

PhilippGackstatter (approved):
Looks good, thank you!

- As for the name, I think I agree that we can change it later if necessary.
- Another thing I remembered was that I believe we wanted to originally expose the storage error to the Wasm bindings, so we could convert the enum variants properly between Wasm and Rust. Doesn't have to be done in this PR and not sure if it's worthwhile and viable in the first place, just wanted to raise it.

olivereanderson (approved):
This is great!

Thanks for the added `pack` and `unpack` methods on `MethodDigest`.

PR #1147, opened 2023-03-19
Wasm bindings for `KeyIdStorage`

abdulmth:
# Description of change
Wasm bindings for https://github.com/iotaledger/identity.rs/pull/1134.

https://github.com/iotaledger/identity.rs/issues/1126

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/bindings/wasm/src/storage/key_id_storage.rs
+++ b/bindings/wasm/src/storage/key_id_storage.rs
@@ -17,13 +17,13 @@ extern "C" {
   #[wasm_bindgen(typescript_type = "KeyIdStorage")]
   pub type WasmKeyIdStorage;
 
-  #[wasm_bindgen(method)]
+  #[wasm_bindgen(method, js_name = insertKeyId)]
   pub fn insert_key_id(this: &WasmKeyIdStorage, method_digest: WasmMethodDigest, key_id: String) -> Promise;
 
-  #[wasm_bindgen(method)]
+  #[wasm_bindgen(method, js_name = getKeyId)]
   pub fn get_key_id(this: &WasmKeyIdStorage, method_digest: WasmMethodDigest) -> PromiseString;
 
-  #[wasm_bindgen(method)]
+  #[wasm_bindgen(method, js_name = deleteKeyId)]
   pub fn delete_key_id(this: &WasmKeyIdStorage, method_digest: WasmMethodDigest) -> Promise;
 }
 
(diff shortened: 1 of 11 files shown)


bindings/wasm/src/storage/key_id_storage.rs
```diff
@@ -0,0 +1,85 @@
...
+  pub type WasmKeyIdStorage;
+
+  #[wasm_bindgen(method, js_name = insertKeyId)]
+  pub fn insert_key_id(this: &WasmKeyIdStorage, method_digest: WasmMethodDigest, key_id: String) -> Promise;
+
+  #[wasm_bindgen(method, js_name = getKeyId)]
+  pub fn get_key_id(this: &WasmKeyIdStorage, method_digest: WasmMethodDigest) -> PromiseString;
+
+  #[wasm_bindgen(method, js_name = deleteKeyId)]
+  pub fn delete_key_id(this: &WasmKeyIdStorage, method_digest: WasmMethodDigest) -> Promise;
```
olivereanderson, line 27: Either add a comment explaining that we must ensure that the `method_digest` argument is never passed in from JS because then they will get nulled out by Rust, or make the methods in this block (but not in the `KeyIdStorage` trait) take `&WasmMethodDigest`.

bindings/wasm/tests/key_id_storage.ts
```diff
@@ -0,0 +1,113 @@
...
+            type: "Ed25519VerificationKey2018",
+            publicKeyMultibase: "zHHoh9NQC9AUsK15Jyyq53VTujxEUizKDXRXd7zbT1B5u",
+        };
+
+        let verificationMethod = VerificationMethod.fromJSON(verificationMethodJson);
+        let methodDigest = new MethodDigest(verificationMethod);
+
+        let packed = methodDigest.pack();
+        let packedExpected = new Uint8Array([0, 74, 60, 10, 199, 76, 205, 180, 133]);
+        assert.deepStrictEqual(packed, packedExpected);
```
olivereanderson, line 18: Could you write a comment explaining that the purpose of this test is to ensure that the computed method digest for a given method is the same for Rust and Wasm and also write where the corresponding Rust test can be found?

olivereanderson (approved):
LGTM!

Added a few minor comments:

bindings/wasm/src/storage/key_id_storage.rs
abdulmth: Taking a reference here isn't allowed by bindgen. For now, I would leave it since it's not used and only there for eventual future need, the same thing exists for `WasmJwkStorage`.

PR #1148, opened 2023-03-28
JwkStorageDocument & JwtCredential validation

olivereanderson:
# Description of change
This PR adds a high level API for working with the JWT/JWS format using key material backed by a storage. 
This includes new methods on the document types ( `CoreDocument` and `IotaDocument`) providing the following functionality: 
-  Generate a new verification method with `Jwk` material where the private key is backed by a `Storage`.
-  Create a JWS with arbitrary bytes as a `payload` using a verification method in the document whose key material is backed by a `Storage`. 
-  Decode and cryptographically verify a JWS (encoded with the compact serialization format). 
-  Issuing a credential as a JWT/JWS following the [`VC-JWT` specification](https://w3c.github.io/vc-jwt/) (version 1.1.).

There is also a new `CredentialValidator` type which basically adapts the old credential validator to work with the `VC-JWT` format. The old validator type either needs to be updated or removed, but this has not been addressed in this PR. 

Here is an end-to-end example in Typescript demonstrating most parts of the new APIs.  
```javascript
       // Setup storage
       const keystore = new JwkMemStore();
        const keyIdStore = new KeyIdMemStore();
        const storage = new Storage(keystore, keyIdStore);

       // Generate a new method with corresponding key material backed by storage
        const fragment = "#key-1";
        await doc.generateMethod(
            storage,
            JwkMemStore.ed25519KeyType(),
            JwsAlgorithm.EdDSA,
            fragment,
            MethodScope.VerificationMethod(),
        );
        // Check that we can resolve the generated method.
        let method = doc.resolveMethod(fragment);
        assert.deepStrictEqual(method instanceof VerificationMethod, true);

        // We can now use the method we just generated to sign an arbitrary string.
        let testString = "test";
        const jws = await doc.createJwt(storage, fragment, testString, new JwsSignatureOptions());

        // Verify the signature and obtain a decoded token.
        const token = doc.verifyJws(jws, new JwsVerificationOptions());
        assert.deepStrictEqual(testString, token.claims());

        // Check that issuing a credential as a JWT works
        const credentialFields = {
            context: "https://www.w3.org/2018/credentials/examples/v1",
            id: "https://example.edu/credentials/3732",
            type: "UniversityDegreeCredential",
            credentialSubject: {
                id: "did:example:ebfeb1f712ebc6f1c276e12ec21",
                degree: {
                    type: "BachelorDegree",
                    name: "Bachelor of Science and Arts",
                },
            },
            issuer: doc.id(),
            issuanceDate: "2010-01-01T00:00:00Z",
        };

        const credential = new Credential(credentialFields);
        // Create the JWT in accordance with the VC-JWT specification. 
        const credentialJwt = await doc.createCredentialJwt(storage, fragment, credential, new JwsSignatureOptions());

        // Check that the credentialJwt can be decoded and verified
        let credentialValidator = new JwtCredentialValidator();
        const credentialRetrieved = credentialValidator.validate(
            credentialJwt,
            doc,
            JwtCredentialValidationOptions.default(),
            FailFast.FirstError,
        ).credential();
        assert.deepStrictEqual(credentialRetrieved.toJSON(), credential.toJSON());

        // Delete the generated method from the document and all corresponding data in the storage.
        const methodId = (method as VerificationMethod).id();
        await doc.purgeMethod(storage, methodId);
        // Check that the method can no longer be resolved.
        assert.deepStrictEqual(doc.resolveMethod(fragment), undefined);
```
Note that in the example we used the built-in JWS signature verifier when verifying the JWS signatures . This default verifier only supports the `EdDSA` algorithm, but this PR also introduces a `JwsSignatureVerifier` interface so users can pass their custom signature verifiers if they are interested in supporting other/more JWS algorithms. 
  
This PR also refactors the  JWS `Encoder` as the previous API we had envisioned turned out to be hard to integrate with the `JwkKeyStorage::sign` method (mainly because futures need to be `Send` or `?Send` depending on whether we are building the bindings or not). There are other ways we could have worked around this such as duplicating the `Encoder::encode` method using different trait bounds, encoding the requirements of the closure at the type level (essentially what is done for the `Resolver`), or introducing a super trait of `JwkKeyStorage` called `AsyncSign : MaybeSendSync`. In the end I decided to go for a solution that requires a bit more code than any of the other suggested alternatives, but in return is arguably more flexible and makes all of `identity_jose` callable in non-async contexts.   

 
## What remains to be done 
1. [Write  tests exercising the error paths and tests checking the various configuration options more thoroughly](https://github.com/iotaledger/identity.rs/issues/1149).   
2. [Explore whether deadlocks and/or panics can occur in the bindings when `sync` and `async` methods reading and writing to a document are concurrently interleaved with `Promise.any` / `Promise.all`](https://github.com/iotaledger/identity.rs/issues/1150). 
3. [Create a high level method for signing and validating presentations issued as JWTs (according to the `VC-JWT` specification)](https://github.com/iotaledger/identity.rs/issues/1151). 
4. Adapt the `DomainLinkage` functionality to the JWT format. 
5. Adapt the `Resolver` to work with the `VC-JWT` format. 
6. Create examples for these new features. 
7. Make a decision on whether to upgrade the old `Proof` format based signature/verification functionality to make it more flexible and align with the upcoming [Data integrity](https://w3c.github.io/vc-data-integrity/) format or to remove it from the API (at least for the time being). 

## Type of change

- [ ] Documentation Fix

Runs under `cargo test --all-features` , `cargo test --no-default-features` and `npm test:unit:node` locally. 

## Change checklist


identity_credential/src/validator/credential_validator.rs
olivereanderson, line 1: Started porting the tests for a couple of the  validation units. Many still need to be ported to the new validator.

identity_credential/src/validator/vc_jwt_validation/credential_jwt_validation_options.rs
```diff
@@ -0,0 +1,68 @@
...
+
+  /// Validation behaviour for [`credentialStatus`](https://www.w3.org/TR/vc-data-model/#status).
+  ///
+  /// Default: [`StatusCheck::Strict`].
+  #[serde(default)]
+  pub status: crate::validator::StatusCheck,
+
+  /// Options which affect the verification of the signature on the credential.
+  #[serde(default)]
+  pub verification_options: JwsVerificationOptions,
```
olivereanderson, line 34: This is the only difference from the old `CredentialValidationOptions`.

identity_credential/src/validator/vc_jwt_validation/credential_jwt_validator.rs
```diff
@@ -0,0 +1,546 @@
...
+impl Default for CredentialValidator {
+  fn default() -> Self {
+    Self::new()
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use identity_core::common::Duration;
+  // All tests here are essentially adaptations of the old CredentialValidator tests.
```
olivereanderson, line 470: There are still many left to adapt though. It may make sense to spread the testing of this type more out. In particular if we take a stance that signing should only be done using a `Storage` there can be some tests here where we mock the signature (and possibly the verifier)  and then some proper integration tests in `identity_storage`.

identity_core/Cargo.toml
```diff
@@ -26,7 +26,7 @@ zeroize = { version = "1.4", default-features = false }
 [dependencies.iota-crypto]
 version = "0.12.1"
 default-features = false
-features = ["ed25519", "random", "sha", "x25519"]
+features = ["ed25519", "random", "sha", "x25519", "std"]
```
olivereanderson, line 29: Need this in order for `crypto::Error` to implement `std::error::Error` so we can use the `#[source]` annotation with the wrapped variant.

identity_document/src/error.rs
```diff
@@ -10,12 +10,12 @@ pub type Result<T, E = Error> = ::core::result::Result<T, E>;
 #[derive(Debug, thiserror::Error, strum::IntoStaticStr)]
 pub enum Error {
   /// Caused by errors from the [identity_core] crate.
-  #[error("{0}")]
+  #[error("unspecified failure")]
   CoreError(#[from] ::identity_core::Error),
-  #[error("{0}")]
+  #[error("unspecified failure")]
```
olivereanderson, line 15: This was done quickly in order to avoid duplicating messages when printing the whole error chain in the bindings.

identity_storage/src/key_id_storage/memstore.rs
```diff
@@ -29,6 +29,11 @@ impl KeyIdMemstore {
       key_id_store: Shared::new(HashMap::new()),
     }
   }
+
+  /// Returns the number of items contained in the [`KeyIdMemstore`].
+  pub async fn count(&self) -> usize {
+    self.key_id_store.read().await.keys().count()
+  }
```
olivereanderson, line 36: This is only used for testing purposes.

identity_storage/src/key_storage/memstore.rs
```diff
@@ -41,6 +41,11 @@ impl JwkMemStore {
       jwk_store: Shared::new(HashMap::new()),
     }
   }
+
+  /// Returns the number of items contained in the [`JwkMemStore`].
+  pub async fn count(&self) -> usize {
+    self.jwk_store.read().await.keys().count()
+  }
```
olivereanderson, line 48: Only used for testing purposes.

identity_storage/src/key_storage/memstore.rs
```diff
@@ -65,7 +70,8 @@ impl JwkStorage for JwkMemStore {
 
     let mut jwk: Jwk = ed25519::encode_jwk(&private_key, &public_key);
     jwk.set_alg(alg.name());
-    let public_jwk: Jwk = jwk.to_public();
+    let mut public_jwk: Jwk = jwk.to_public().unwrap();
```
olivereanderson, line 73: `Unwrap` is OK because the `None` variant only occurs for `kty = oct`.

olivereanderson:
> We're interested in using SecP-256, so according to what you wrote:
> 
> > built-in JWS signature verifier when verifying the JWS signatures . This default verifier only supports the EdDSA algorithm, but this PR also introduces a JwsSignatureVerifier interface so users can pass their custom signature verifiers if they are interested in supporting other/more JWS algorithms.
> 
> we could take your example as it is, build our own `JwsSignatureVerifier` and pass that to verification, and it should work, right?

If by SecP-256 you mean using ECDSA with the secp256k1 curve and the SHA-256 cryptographic hash function, then yes definitely. You could indeed pass your own `JwsSignatureVerifier` built to support [`JwsAlgorithm::ES256K`](https://github.com/iotaledger/identity.rs/blob/main/identity_jose/src/jws/algorithm.rs#L42) to verification. So in the case of plain data: instead of 
```
const token = doc.verifyJws(jws, new JwsVerificationOptions());
```
You would write 
```
const token = doc.verifyJws(jws, new JwsVerificationOptions(), mySignatureVerifier);
```
and similarly in the case of verifying a credential it would be:
```
let credentialValidator = new JwtCredentialValidator(mySignatureVerifier);
        const credentialRetrieved = credentialValidator.validate(
            credentialJwt,
            doc,
            JwtCredentialValidationOptions.default(),
            FailFast.AllErrors,
        ).credential();
 ```

identity_storage/src/key_storage/memstore.rs
PhilippGackstatter: Note to self: Comment this directly in code.

PR #1175, opened 2023-05-29
Support JWT Presentations

abdulmth:
# Description of change
Add support for issuing and validating JWT presentation.

## Links to any relevant issues
https://github.com/iotaledger/identity.rs/issues/1151
(implementation does not match the issue description)

## Type of change
- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

Issue #1151, opened 2023-03-28
[Task] Create functionality for issuing presentations as JWTs


olivereanderson:
# Description
PR #1148 introduces functionality for issuing credentials in the JWT format according to the [VC-JWT specification](https://w3c.github.io/vc-jwt/) (version 1.1). Now we need to provide similar APIs for presentations. 

## Approach 
I would suggest to take a similar approach as what was done in the (simpler) case of Credentials. Where we keep users as close to the standard [VC Data Model](https://www.w3.org/TR/vc-data-model/#presentations-0) as possible.  In other words one essentially passes a `Presentation` to a (storage) document which in turn carries out the required transformations and yields a JWS. 

In analogy with the case of Credentials the golden path for validating a presentation issued as a JWS would then be a  (JWT-based) `PresentationValidator` that decodes and validates the JWS and returns a `PresentationToken` wrapping a `Presentation`. 

The complication here is the question of what the value of the `verifiableCredential` property of a presentation should before and after signing it.  The natural thing to do when signing is to place the JWT representations of each credential into this array, but when verifying we would want them in the standard `Credential` format. To solve this problem I would suggest using generics in Rust (and then manual monomorphization in the Bindings). More precisely we could make the `verifiable_credential` field in `Presentation` generic of type `T` and we introduce two type aliases: 

```
/// A Presentation that has yet to be issued. 
/// This presentation type wraps JWT/JWS representations of the contained credentials. 
pub type PreSignedPresentation = Presentation<String, ..>;

/// A presentation that has been decoded from a JWS.
pub type DecodedPresentation = Presentation<Credential, ..>;
```

The high level APIs for working with these would then be: 

```
/// Issue a presentation as a JWS in accordance with the VC-JWT specification. 
JwkStorageDocumentExt::sign_presentation(presentation: &PreSignedPresentation, ...) -> Result<String>`;

PresentationValidator::validate(presentation_jws: &str, ...) -> Result<PresentationToken>;
```
where `PresentationToken` is analogous to `CredentialToken` and defined as: 
```
/// Decoded presentation from a cryptographically verified JWS. 
pub struct PresentationToken {
    /// The decoded presentation
   pub  presentation: DecodedPresentation,
   
   /// The parsed protected header
   pub header: Box<JwsHeader>, 
   
   /// The parsed protected headers of the credentials in the 
   /// order they appear in `presentation.verifiabe_credential`. 
   pub credential_headers: Vec<JwsHeader>,
}
```

[...]

PR #1179, opened 2023-06-06
Wasm Bindings for JWT Presentations

abdulmth:
# Description of change
Add wasm bindings for JWT presentation and its validation.

https://github.com/iotaledger/identity.rs/issues/1151
https://github.com/iotaledger/identity.rs/pull/1175

## Type of change

- [ ] Documentation Fix


bindings/wasm/tests/storage.ts
```diff
@@ -186,58 +186,132 @@ describe("#JwkStorageDocument", function() {
...
-            credentialJwt,
-            doc,
-            JwtCredentialValidationOptions.default(),
-            FailFast.AllErrors,
-        ).credential();
-        // Check that customVerifer.verify was indeed called
-        assert.deepStrictEqual(customVerifier.verifications(), 2);
-        assert.deepStrictEqual(credentialRetrievedCustom.toJSON(), credential.toJSON());
+        assert.deepStrictEqual(decoded.credentials()[0].credential().toJSON(), credential.toJSON());
+        assert.equal(decoded.expirationDate()?.toString(), expirationDate?.toString());
```
PhilippGackstatter, line 267:
```suggestion
        assert.equal(decoded.expirationDate()!.toString(), expirationDate!.toString());
```

We know those properties are set, so we shouldn't use `?`, right? If they don't end up being set, we would like the test to fail.

PhilippGackstatter (approved):
Looks good to me.

PR #1180, opened 2023-06-09
Use JWT credentials for Domain Linkage

PhilippGackstatter:
# Description of change

Updates the domain linkage types to use JWT-based credentials.

Noteworthy changes:

- `test_multiple_subjects` was removed as it is not possible to create a JWT with multiple subjects since the `"sub"` claim of a JWT is of type string rather than array.
- Replace `Issuer` type with `CoreDID` on the `issuer`, since the spec prescribes:
  - > credentialSubject.id MUST be present, and MUST be a DID. 
- Validate that origin is a domain.

part of #1103 

## Type of change

- [ ] Documentation Fix

Domain Linkage tests were updated.

## Change checklist


bindings/wasm/examples/src/util.ts
```diff
@@ -15,6 +18,53 @@ export const FAUCET_ENDPOINT = "http://localhost:8091/api/enqueue";
 
 /** Creates a DID Document and publishes it in a new Alias Output.
 
+Its functionality is equivalent to the "create DID" example
+and exists for convenient calling from the other examples. */
+export async function createDidStorage(client: Client, secretManager: SecretManager, storage: Storage): Promise<{
```
abdulmth, line 23: Should we maybe rename this function? Without a context, it means creating a storage rather than DID. Something like `createIdentity` could work?

identity_credential/src/credential/mod.rs
```diff
@@ -7,7 +7,10 @@
 
 mod builder;
 mod credential;
+// TODO: Feature-gate Domain Linkage Types behind its own flag and require flag = ["validator"].
```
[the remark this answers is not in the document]
PhilippGackstatter: I intended to do this in a small follow-up PR.

bindings/wasm/examples/src/util.ts
PhilippGackstatter: Yeah, eventually we can just rename it to `createDid` like the other convenience function, but while we transition the examples that name is already taken. So I will do this when we have reworked all examples.

PR #1186, opened 2023-06-15
Make JWT presentations generic

abdulmth:
# Description of change

This PR makes `JwtPresentation` (which will be renamed to `Presentation`) generic over credentials, which allows credential types other than JWT. This allows for a much greater degree of flexibility, in particular mixing credentials of different types in a single Presentation. The VC Data Model allows for this, but the data model shipped by this library was not able to accommodate this so far. In the interest of future proofing, this change was made.

As a consequence, the Presentation Validator can no longer validate the credentials contained in a Presentation, as it may not have an appropriate credential validator for the given credential type. Therefore, the presentation validator only validates presentations, while the appropriate credential validator must be used for the contained credentials (e.g. the `JwtCredentialValidator` for credentials of type `Jwt`). This reduces the responsibility of the presentation validator to just presentations, in line with the single responsibility principle, and makes the codebase easier to understand.

The trade-off is that the user must verify all the credentials themselves by using the appropriate credential validator shipped by this library, or one that is user-provided. This change is reflected in the create_vp example.

https://github.com/iotaledger/identity.rs/pull/1175


bindings/wasm/examples/src/0_basic/6_create_vp.ts
```diff
@@ -152,40 +153,46 @@ export async function createVP() {
...
+    const presentationHolderDID: CoreDID = JwtPresentationValidator.extractHolder(presentationJwt);
+    const resolvedHolder = await resolver.resolve(presentationHolderDID.toString());
 
-    const presentationDids = JwtPresentationValidator.extractDids(presentationJwt);
-    const resolvedHolder = await resolver.resolve(presentationDids.holder.toString());
-    const resolvedIssuer = await resolver.resolve(presentationDids.issuers[0].toString());
-
-    // Validate the presentation and all the credentials included in it according to the validation options.
-    new JwtPresentationValidator().validate(
+    // Validate presentation, Note that this doesn't validate the included credentials.
```
PhilippGackstatter, line 167:
```suggestion
    // Validate presentation. Note that this doesn't validate the included credentials.
```

PhilippGackstatter (changes requested):
Thanks for tackling this! The current `JwtPresentation` should be renamed to just `Presentation`. Then we have a generic `Presentation` after removing the old `Presentation` type (in a separate PR). We can remove it together with the old validators, to not make this PR excessively large.

Moreover we can introduce a `Proof` type and use it as the type of the `proof` field in the (new) `Presentation` and (old) `Credential` types. This type could be:

```rust
#[derive(Debug, Clone, Deserialize, Serialize)]
pub struct Proof {
  #[serde(rename = "type")]
  type_: String,
  #[serde(rename = "flatten")]
  properties: Object,
}
```

To model a type that needs to have a `type` field and can hold any additional data. This way it will be easier to extract the proof type of a credential or presentation.

If it's too annoying to replace the current `Proof` in `Credential` then we can also deal with that in the future validator removal PR.

Once we have that, I think we can merge this PR and come up with a better approach for Wasm. The current one means Wasm can only hold JWT credentials in a JWT presentation. To allow for mixed types, we could introduce an enum that allows for either `Jwt` or `Credential` or we make it `Object` and let the user figure it out. The former is probably better.

examples/0_basic/6_create_vp.rs
```diff
@@ -174,46 +177,58 @@ async fn main() -> anyhow::Result<()> {
...
+    .iter()
+    .map(CredentialValidator::extract_issuer_from_jwt)
+    .collect::<Result<Vec<CoreDID>, _>>()?;
+  let issuers_documents: HashMap<CoreDID, IotaDocument> = resolver.resolve_multiple(&issuers).await?;
+
+  // Validate the credentials in the presentation.
+  let credential_validator: CredentialValidator = CredentialValidator::new();
+  let validation_options = CredentialValidationOptions::default();
+  for (index, jwt_vc) in jwt_credentials.iter().enumerate() {
+    //SAFETY: Indexing should be fine since we extracted the DID from each credential.
```
PhilippGackstatter, line 213:
```suggestion
    // SAFETY: Indexing should be fine since we extracted the DID from each credential.
```

examples/0_basic/6_create_vp.rs
```diff
@@ -174,46 +177,58 @@ async fn main() -> anyhow::Result<()> {
...
+  let jwt_credentials = &presentation.presentation.verifiable_credential;
+  let issuers: Vec<CoreDID> = jwt_credentials
+    .iter()
+    .map(CredentialValidator::extract_issuer_from_jwt)
+    .collect::<Result<Vec<CoreDID>, _>>()?;
+  let issuers_documents: HashMap<CoreDID, IotaDocument> = resolver.resolve_multiple(&issuers).await?;
+
+  // Validate the credentials in the presentation.
+  let credential_validator: CredentialValidator = CredentialValidator::new();
+  let validation_options = CredentialValidationOptions::default();
```
PhilippGackstatter, line 211:
```suggestion
  let validation_options: CredentialValidationOptions = CredentialValidationOptions::default();
```

examples/0_basic/6_create_vp.rs
```diff
@@ -174,46 +177,58 @@ async fn main() -> anyhow::Result<()> {
...
-  )?;
 
+  // Validate presentation. Note that this doesn't validate the included credentials.
+  let presentation_validation_options =
+    JwtPresentationValidationOptions::default().presentation_verifier_options(presentation_verifier_options);
+  let presentation: DecodedJwtPresentation =
+    JwtPresentationValidator::new().validate(&presentation_jwt, &holder, &presentation_validation_options)?;
+
+  // Concurrently resolve the issuers' documents.
+  let jwt_credentials = &presentation.presentation.verifiable_credential;
```
PhilippGackstatter, line 202:
```suggestion
  let jwt_credentials: &OneOrMany<Jwt> = &presentation.presentation.verifiable_credential;
```

examples/0_basic/6_create_vp.rs
```diff
@@ -174,46 +177,58 @@ async fn main() -> anyhow::Result<()> {
...
+      .validate::<_, Object>(jwt_vc, issuer_document, &validation_options, FailFast::FirstError)
+      .unwrap();
+
+    // Check that the presentation holder matches the credential subject.
+    CredentialValidator::check_subject_holder_relationship(
+      &decoded_credential.credential,
+      &holder_did.to_url().into(),
+      SubjectHolderRelationship::AlwaysSubject,
+    )?;
+  }
```
PhilippGackstatter, line 227:
```suggestion
  }

```

Nit: Add a newline here so this section looks less crowded :)

examples/0_basic/6_create_vp.rs
```diff
@@ -174,46 +177,58 @@ async fn main() -> anyhow::Result<()> {
...
+    let decoded_credential: DecodedJwtCredential<Object> = credential_validator
+      .validate::<_, Object>(jwt_vc, issuer_document, &validation_options, FailFast::FirstError)
+      .unwrap();
+
+    // Check that the presentation holder matches the credential subject.
+    CredentialValidator::check_subject_holder_relationship(
+      &decoded_credential.credential,
+      &holder_did.to_url().into(),
+      SubjectHolderRelationship::AlwaysSubject,
+    )?;
```
PhilippGackstatter, line 226: I think we should make this part of credential validation as an option in `CredentialValidationOptions`. I.e. we add an option `subject_holder_check: Option<(Url, SubjectHolderRelationship)>`, and if set, we do this check here inside the `CredentialValidator`. Otherwise this feels pretty manual and like a low-level check here, while all other low-level checks are already done inside `CredentialValidator::validate`.

We would only set this option if the credential being validated was part of a presentation.

identity_credential/src/presentation/jwt_presentation.rs
```diff
@@ -28,7 +28,7 @@ use super::JwtPresentationOptions;
 
 /// Represents a bundle of one or more [`Credential`]s expressed as [`Jwt`]s.
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
-pub struct JwtPresentation<T = Object> {
+pub struct JwtPresentation<CRED = Jwt, T = Object> {
```
PhilippGackstatter, line 31:
```suggestion
pub struct Presentation<CRED = Jwt, T = Object> {
```

This should just be our implementation of the `Presentation` data model, right? Also rename the file if you agree.

PhilippGackstatter (approved):

bindings/wasm/src/credential/jwt_credential_validation/untyped_credential.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::credential::WasmCredential;
+use crate::credential::WasmJwt;
+
+#[wasm_bindgen(js_name = UntypedCredential)]
+pub struct WasmUntypedCredentialContainer(UntypedCredential);
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum UntypedCredential {
```
abdulmth, line 17: I thought we agreed on `UnknownCredential`, no?

identity_credential/src/validator/vp_jwt_validation/jwt_presentation_validator.rs
```diff
@@ -62,39 +56,36 @@ where
...
-  /// - the signatures and some properties of the constituent credentials (see [`CredentialValidator`]).
   ///
   /// Validation is done with respect to the properties set in `options`.
   ///
   /// # Warning
-  /// The lack of an error returned from this method is in of itself not enough to conclude that the presentation can be
-  /// trusted. This section contains more information on additional checks that should be carried out before and after
-  /// calling this method.
+  /// * This method does NOT validate the constituent credentials, nor the relationship between the
+  /// credentials' issuers and the presentation holder.
```
abdulmth, line 64:
```suggestion
  /// * This method does NOT validate the constituent credentials.
```

abdulmth (commented):
Looks good to me, thanks for making the changes.  One thing that I would prefer to have is to add a specific method like `getRaw` on the `UntypedCredential` that returns a string and has documentation, just to make it stand out since `toJSON` is everywhere.

identity_credential/src/validator/vp_jwt_validation/jwt_presentation_validator.rs
PhilippGackstatter: It doesn't hurt to leave that extra bit of information in, actually. Why would you like to remove it?

abdulmth: I thought since we have it now in the options of the credential validator, it's not part of the presentation validator. But it's fine to keep it.

PR #1189, opened 2023-06-22
Add `resolve_multiple` to Resolver.

abdulmth:
# Description of change
Allow resolving multiple DID documents in parallel.

The method first removes any did duplicates before resolving, since duplicates can be common, for example if a verificable presentation has multiple credentials issued by the same DID. The order is then re-established for convenience and to avoid the need of filtering to retrieve the correct DID. To allow this, the documents must be clonable. 

https://github.com/iotaledger/identity.rs/pull/1183

## Type of change

- [ ] Documentation Fix

### DIFF
--- a/identity_resolver/src/resolution/resolver.rs
+++ b/identity_resolver/src/resolution/resolver.rs
@@ -106,13 +106,8 @@ where
   /// * If the resolution process of any DID fails.
   ///
   /// ## Note
-  /// * The order of the documents in the returned vector matches that in `dids`.
-  /// * If `dids` contains duplicates, these will be resolved only once and the resolved document
-  /// is copied into the returned vector to match the order of `dids`.
-  pub async fn resolve_multiple<D: DID>(&self, dids: &[D]) -> Result<Vec<DOC>>
-  where
-    DOC: Clone,
-  {
+  /// * If `dids` contains duplicates, these will be resolved only once.
+  pub async fn resolve_multiple<D: DID>(&self, dids: &[D]) -> Result<HashMap<D, DOC>> {
     let futures = FuturesUnordered::new();
     // Create set to remove duplicates to avoid unnecessary resolution.
     let dids_set: HashSet<D> = dids.iter().cloned().collect();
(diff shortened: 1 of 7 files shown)


identity_resolver/src/resolution/resolver.rs
```diff
@@ -95,6 +98,45 @@ where
...
+    for did in dids {
+      let doc: DOC = documents
+        .get(did)
+        .cloned()
+        // Error instead of `unwrap`.
+        .ok_or(Error::new(ErrorCause::UnspecificError {
+          message: "Internal Error".to_owned(),
+        }))?;
+      ordered_documents.push(doc);
+    }
```
PhilippGackstatter, line 137: Why do we need to return it in order? The Credential validator has no assumption on the order. It will iterate the returned Vec and `find` the corresponding document. We can just use FuturesUnordered and return the vec, without sorting it imo. That also means we don't have to clone the document which is avoids the copying.
If you agree, remove the `UnspecificError` variant.

PhilippGackstatter (changes requested):

identity_resolver/src/resolution/resolver.rs
PhilippGackstatter: The method will be used most of the time for validation and there we don't need the order. We can state in the documentation what the assumption is, this is a common thing.
If it's an expected thing in JS, then we should prepare that sorted return type in the bindings. I'd suggest to return a map, mapping DID to document, so we avoid the duplicates in the vector. I would really like to avoid unnecessary copies of kilobyte-large documents.

abdulmth: I updated the rust to return a hashmap and did the sorting in the bindings.

PhilippGackstatter (approved):
I like the map return type. Since we no longer need to pass a slice of documents to the validate function of the presentation validator after #1186 is merged, I assume this map will actually make things easier. I assume that in the loop of credential validations the map will make obtaining the proper document quite convenient and I would also suggest to return a map in Wasm (which was actually what I meant originally), so we'll have the same convenience there, but we'll see how that works out when we update the  `6_create_vp` example to use `resolve_multiple`. We should do that after merging, also so that we have at least a public usage and see for ourselves how well it works.

PR #1207, opened 2023-07-17
Rename `CredentialValidator` to `JwtCredentialValidator`

abdulmth:


bindings/wasm/docs/api-reference.md
PhilippGackstatter, line 1: Btw, you can probably save the work of building the docs and we can just do it now whenever we release a new version and then move the docs to the wiki repo.

PhilippGackstatter (approved):

PR #1209, opened 2023-07-28
Improve `Proof` 

abdulmth:
# Description of change

Improve the `Proof` type in `Credential` and `Presentation`. 
`Proof` now must include a `type` property to enable matching different proof types during validation.


bindings/wasm/src/credential/credential.rs
```diff
@@ -195,18 +196,25 @@ impl WasmCredential {
...
   #[wasm_bindgen]
   pub fn properties(&self) -> Result<MapStringAny> {
     MapStringAny::try_from(&self.0.properties)
   }
+
+  /// Sets the proof property of the {@link Credential}.
+  ///
+  /// Note that this proof is not related to JWT.
+  #[wasm_bindgen(js_name = "setProof")]
+  pub fn set_proof(&mut self, proof: Option<WasmProof>) {
```
PhilippGackstatter, line 215: We don't have setters for any other property, why for proof?
We expect them to be set in the constructor. Why not add it to `ICredential`?

identity_credential/src/credential/proof.rs
```diff
@@ -0,0 +1,96 @@
...
+/// can be utilized to implement standards or user-defined proofs. The presence of the
+/// `type` field is necessary to accommodate different types of cryptographic proofs.
+///
+/// Note that this proof is not related to JWT and can be used in combination or as an alternative
+/// to it.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct Proof {
+  /// Type of proof.
+  #[serde(rename = "type")]
+  pub _type: String,
```
PhilippGackstatter, line 21:
```suggestion
  pub type_: String,
```

Don't use `_type`, that is an implicit "ignore this variable" instruction. We don't want the compiler to not inform us if that field is unused. Use `type_` instead.

identity_credential/src/credential/proof.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[cfg(test)]
+mod tests {
+  use super::Proof;
+  use crate::credential::Credential;
+  use identity_core::common::Object;
+  use identity_core::convert::FromJson;
+
+  #[test]
+  fn test_from_json() {
```
PhilippGackstatter, line 43:
```suggestion
  fn test_proof_from_json() {
```

identity_credential/src/credential/proof.rs
```diff
@@ -0,0 +1,96 @@
...
+    assert_eq!(proof._type, "test-proof");
+    let value = proof
+      .properties
+      .get(&"signature".to_owned())
+      .expect("property in proof doesn't exist");
+    assert_eq!(value, "abc123");
+  }
+
+  #[test]
+  fn test_credential_from_json() {
```
PhilippGackstatter, line 61:
```suggestion
  fn test_credential_with_proof_from_json() {
```

Nit: Be more specific in what this test really tests.

bindings/wasm/src/credential/credential.rs
abdulmth: well, the way I thought a proof can be made is by constructing the credential without proof, creating the signature of the resulted JSON or whatever and then you set the proof. So, it's important to have a credential without proof and then be able to set the proof. Also, the credential builder in Rust doesn't allow to set the proof.

PhilippGackstatter: Fair point. Let's keep the setter. Still we should add it to `ICredential`, just to have the option I think. One of the examples in the spec that can be set as the `proof` was a proof-of-work, which may or may not be independent of the Credential. Then it's more convenient to set it all in the constructor right away.

bindings/wasm/src/credential/proof.rs
```diff
@@ -0,0 +1,45 @@
...
+impl WasmProof {
+  #[wasm_bindgen(constructor)]
+  pub fn constructor(_type: String, properties: JsValue) -> Result<WasmProof> {
+    let properties: Object = properties.into_serde().wasm_result()?;
+    Ok(WasmProof(Proof::new(_type, properties)))
+  }
+
+  /// Returns the type of proof.
+  #[wasm_bindgen(js_name = "type")]
+  pub fn _type(&self) -> String {
```
PhilippGackstatter, line 33:
```suggestion
  pub fn type_(&self) -> String {
```

bindings/wasm/src/credential/proof.rs
```diff
@@ -0,0 +1,45 @@
...
+///
+/// Note that this proof is not related to JWT and can be used in combination or as an alternative
+/// to it.
+#[wasm_bindgen(js_name = Proof)]
+pub struct WasmProof(pub(crate) Proof);
+
+#[wasm_bindgen(js_class = Proof)]
+impl WasmProof {
+  #[wasm_bindgen(constructor)]
+  pub fn constructor(_type: String, properties: JsValue) -> Result<WasmProof> {
```
PhilippGackstatter, line 26:
```suggestion
  pub fn constructor(type_: String, properties: JsValue) -> Result<WasmProof> {
```

identity_credential/src/credential/proof.rs
```diff
@@ -0,0 +1,99 @@
...
+  pub type_: String,
+
+  /// Properties of the proof, entries depend on the proof type.
+  #[serde(flatten)]
+  pub properties: Object,
+}
+
+impl Proof {
+  /// Creates a new `Proof`.
+  pub fn new(_type: String, properties: Object) -> Proof {
```
PhilippGackstatter, line 30:
```suggestion
  pub fn new(type_: String, properties: Object) -> Proof {
```

PhilippGackstatter (approved):
LGTM!

Please don't use `_type` anywhere.

PR #1231, opened 2023-09-05
Change `verifiable_credential` to type `Vec<CRED>` in `Presentation`

nanderstabel:
# Description of change
This change ensures that `verifiable_credential` in `Presentation` will serialize into a `[T]` instead of a `T`, which was previously the case when `verifiable_credential` would only hold one single credential.

This previous serialization is in line with how `credentialSubject` can be serialized: https://www.w3.org/TR/vc-data-model/#credential-subject, however this is not explicitly stated for `verifiableCredential`: https://www.w3.org/TR/vc-data-model/#presentations-0.
The following examples that are used in the specification seem to imply that 'single credentials' should be serialized as `[<credential>]`:
* https://www.w3.org/TR/vc-data-model/#example-a-simple-example-of-a-verifiable-presentation
* https://www.w3.org/TR/vc-data-model/#example-usage-of-the-termsofuse-property-by-a-holder
* https://www.w3.org/TR/vc-data-model/#example-a-verifiable-presentation-that-supports-cl-signatures

The W3C VC Data Model 2.0 specification fortunately is more specific in this regard: https://www.w3.org/TR/vc-data-model-2.0/#presentations-0:
> **verifiableCredential**
    The verifiableCredential [property](https://www.w3.org/TR/vc-data-model-2.0/#dfn-property) MAY be present. The value MUST be an array of one or more [verifiable credentials](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential), or of data derived from [verifiable credentials](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) in a cryptographically [verifiable](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verify) format.

Therefore the introduction of the change in this PR.

Important detail: 
Both W3C VC Data Model versions state that: `MAY be present` and `**If** present...`, which imply that `verifiableCredential` is an optional property. I fail to think of any use case where one would have a presentation without any credentials included but this optional status of `verifiableCredential` is included in this PR as well.

n/a

## Type of change

- [ ] Documentation Fix

* Added unit test: `test_presentation_deserialization` which tests deserializing from an example verifiable presentation (with some minor adjustments): https://www.w3.org/TR/vc-data-model/#example-a-simple-example-of-a-verifiable-presentation
* Added unit test: `test_presentation_deserialization_without_credentials` which tests deserializing a verifiable presentation as well, but one without any `verifiableCredential`.
* Added unit test: `test_presentation_builder_valid_without_credentials` which tests building a Presentation without any credentials set.

## Change checklist

### DIFF
--- a/identity_credential/src/error.rs
+++ b/identity_credential/src/error.rs
@@ -45,6 +45,11 @@ pub enum Error {
   #[error("could not convert JWT to the VC data model: {0}")]
   InconsistentCredentialJwtClaims(&'static str),
 
+  /// Caused when deserializing a Presentation with an empty array for the
+  /// `verifiableCredential` property.
+  #[error("empty verifiableCredential array")]
+  EmptyVerifiableCredentialsArray,
+
   /// Caused when attempting to convert a JWT to a `Presentation` that has conflicting values
   /// between the registered claims and those in the `vp` object.
   #[error("could not convert JWT to the VP data model: {0}")]
--- a/identity_credential/src/error.rs
+++ b/identity_credential/src/error.rs
@@ -47,8 +47,8 @@ pub enum Error {
 
   /// Caused when deserializing a Presentation with an empty array for the
   /// `verifiableCredential` property.
-  #[error("empty verifiableCredential array")]
-  EmptyVerifiableCredentialsArray,
+  #[error("empty verifiableCredential array in presentation")]
+  EmptyVerifiableCredentialArray,
 
   /// Caused when attempting to convert a JWT to a `Presentation` that has conflicting values
   /// between the registered claims and those in the `vp` object.
--- a/identity_credential/src/presentation/presentation.rs
+++ b/identity_credential/src/presentation/presentation.rs
@@ -39,7 +39,7 @@ pub struct Presentation<CRED, T = Object> {
   pub types: OneOrMany<String>,
   /// Credential(s) expressing the claims of the `Presentation`.
   #[serde(default = "Default::default", rename = "verifiableCredential")]
-  pub verifiable_credential: OneOrMany<CRED>,
+  pub verifiable_credential: Vec<CRED>,
   /// The entity that generated the `Presentation`.
   pub holder: Url,
   /// Service(s) used to refresh an expired [`Credential`] in the `Presentation`.
--- a/identity_credential/src/presentation/presentation.rs
+++ b/identity_credential/src/presentation/presentation.rs
@@ -38,6 +38,7 @@ pub struct Presentation<CRED, T = Object> {
   #[serde(rename = "type")]
   pub types: OneOrMany<String>,
   /// Credential(s) expressing the claims of the `Presentation`.
+  #[rustfmt::skip]
   #[serde(default = "Default::default", rename = "verifiableCredential", skip_serializing_if = "Vec::is_empty")]
   pub verifiable_credential: Vec<CRED>,
   /// The entity that generated the `Presentation`.
--- a/identity_credential/src/presentation/presentation.rs
+++ b/identity_credential/src/presentation/presentation.rs
@@ -244,7 +244,7 @@ mod tests {
       .source()
       .unwrap()
       .to_string(),
-      "empty verifiableCredential array in presentation"
+      crate::error::Error::EmptyVerifiableCredentialArray.to_string()
     );
   }
 }
--- a/identity_credential/src/presentation/presentation.rs
+++ b/identity_credential/src/presentation/presentation.rs
@@ -66,7 +66,7 @@ where
   let verifiable_credentials = Vec::<T>::deserialize(deserializer)?;
 
   (!verifiable_credentials.is_empty())
-    .then(|| verifiable_credentials)
+    .then_some(verifiable_credentials)
     .ok_or_else(|| de::Error::custom(Error::EmptyVerifiableCredentialsArray))
 }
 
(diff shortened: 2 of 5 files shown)


identity_credential/src/presentation/presentation.rs
```diff
@@ -38,8 +38,9 @@ pub struct Presentation<CRED, T = Object> {
   #[serde(rename = "type")]
   pub types: OneOrMany<String>,
   /// Credential(s) expressing the claims of the `Presentation`.
-  #[serde(default = "Default::default", rename = "verifiableCredential")]
-  pub verifiable_credential: OneOrMany<CRED>,
+  #[rustfmt::skip]
+  #[serde(default = "Default::default", rename = "verifiableCredential", skip_serializing_if = "Vec::is_empty")]
+  pub verifiable_credential: Vec<CRED>,
```
PhilippGackstatter, line 43: Thanks for opening the PR!

With this change we would have at least these differences to before:
1. When deserializing, an empty array in JSON would be allowed.
2. When constructing a presentation, an empty vec would be allowed to be set, which it was previously not.

Case 2 is fine, because that's our "new way" to serialize a presentation without the `verifiableCredential` field, which we probably have no use case for, but can allow without harm.

Case 1 I think we should disallow with a `deserialize_with = "path"`. We implement the VC Data Model 1.1 and it says:
> If present, the value of the verifiableCredential [property](https://www.w3.org/TR/vc-data-model/#dfn-property) MUST be constructed from one or more [verifiable credentials](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-credentials)

I interpret that as: If the field is present, then it must be a non-empty array.

@nanderstabel Do you have any thoughts on that?

nanderstabel: Yes I agree with you! Disallowing an empty array value is more in line with the spec so I took you suggestion and added
`deserialize_verifiable_credential`. 

I also added a new corresponding `Error` variant: `EmptyVerifiableCredentialsArray`. Not sure whether it adheres to any
naming convention you may have for the different Errors so please let me know if you would like me to change it!

identity_credential/src/error.rs
```diff
@@ -45,6 +45,11 @@ pub enum Error {
   #[error("could not convert JWT to the VC data model: {0}")]
   InconsistentCredentialJwtClaims(&'static str),
 
+  /// Caused when deserializing a Presentation with an empty array for the
+  /// `verifiableCredential` property.
+  #[error("empty verifiableCredential array")]
+  EmptyVerifiableCredentialsArray,
```
PhilippGackstatter, line 51:
```suggestion
  #[error("empty verifiableCredential array in presentation")]
  EmptyVerifiableCredentialArray,
```

Nit: Would clarify where the error occured (in the presentation) and loose the `s` since the field doesn't have it :)

identity_credential/src/presentation/presentation.rs
```diff
@@ -143,3 +157,88 @@ where
...
+        "https://www.w3.org/2018/credentials/examples/v1"
+      ],
+      "holder": "did:test:abc1",
+      "type": "VerifiablePresentation"
+    }))
+    .is_ok());
+  }
+
+  #[test]
+  fn test_presentation_deserialization_with_empty_credentials_array() {
```
PhilippGackstatter, line 231: Since we have a dedicated error type for this case, it would be nice to check for it instead of the less precise `is_err`. If the deserialization fails for any other reason in the future than the empty array, it would still pass.

PhilippGackstatter (approved):
Looks good to me. Thanks again :pray:.

I'll approve this with additional comments to mean that I leave it to your judgement whether to implement them or not.

abdulmth (approved):
Looks good to me, thanks for putting  in the effort.

identity_credential/src/error.rs
nanderstabel: I've adjusted the test to test for the specific error. I feel there must be a prettier way to do it but doing a string compare does the job as well.

identity_credential/src/presentation/presentation.rs
```diff
@@ -143,3 +157,94 @@ where
...
+        "holder": "did:test:abc1",
+        "type": "VerifiablePresentation",
+        "verifiableCredential": []
+      }))
+      .unwrap_err()
+      .source()
+      .unwrap()
+      .to_string(),
+      "empty verifiableCredential array in presentation"
+    );
```
PhilippGackstatter, line 248:
```suggestion
    assert_eq!(
      Presentation::<()>::from_json_value(json!({
        "@context": [
          "https://www.w3.org/2018/credentials/v1",
          "https://www.w3.org/2018/credentials/examples/v1"
        ],
        "holder": "did:test:abc1",
        "type": "VerifiablePresentation",
        "verifiableCredential": []
      }))
      .unwrap_err()
      .source()
      .unwrap()
      .to_string(),
      crate::Error::EmptyVerifiableCredentialArray.to_string()
    );
```

I think the serde error erases the type, right? So there's probably no better way to test.

Except you could compare against the string repr of the error, that would be more resilient.

PhilippGackstatter (commented):
The Wasm examples CI workflow ran for 2:50h but didn't finish until then so I cancelled it. Had this before at some point but a rerun did the trick back then. I haven't restarted in case you want to fix the comment, otherwise just rerun the workflow.

identity_credential/src/presentation/presentation.rs
nanderstabel: ahh yeah indeed, thanks for the suggestion :)

nanderstabel:
@PhilippGackstatter @abdulmth thank you both for reviewing! I made the last suggested change to the unit test. Hopefully the CI workflow won't have any issues :)

PhilippGackstatter:
@nanderstabel Our convention is to let the PR creator hit the merge button, so feel free to go ahead whenever you want. Thanks again :rocket:!

nanderstabel:
Great, thanks again!

PR #1235, opened 2023-09-14
Fix issuer claim check in VC

abdulmth:
When JWT VC is validated, `check_consistency` will be called that checks if values are duplicated in the JWT claims and the VC itself in the `vc` claim. These values will be copied afterward when `Credential` is created after decoding. 

This works fine for all the claims except `iss`. It's not checked for consistency and it will be copied to the `properties` field in `Credential` instead of `issuer`. This PR fixes that and adds some tests. 

I expect similar issues in JWT VP. I will take a look and eventually open another PR.

### DIFF
--- a/identity_credential/src/credential/jwt_serialization.rs
+++ b/identity_credential/src/credential/jwt_serialization.rs
@@ -449,7 +449,7 @@ mod tests {
       }
     }"#;
 
-    // `sub`, `jti`, `iss`, `nbf` are duplicated in `vc`.
+    // `sub`, `exp`, `jti`, `iss`, `nbf` are duplicated in `vc`.
     let claims_json: &str = r#"
     {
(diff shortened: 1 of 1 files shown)


PhilippGackstatter (approved):

identity_credential/src/credential/jwt_serialization.rs
```diff
@@ -409,4 +427,248 @@ mod tests {
...
+        .try_into_credential();
+    assert!(matches!(
+      credential_from_claims_result.unwrap_err(),
+      Error::InconsistentCredentialJwtClaims("inconsistent credential id")
+    ));
+  }
+
+  #[test]
+  fn inconsistent_subject() {
+    // issuer is inconsistent (15 instead of 14).
```
abdulmth, line 561:
```suggestion
```

eike-hass (approved):

PR #1237, opened 2023-09-15
Allow custom JWT claims for credentials

abdulmth:


identity_storage/src/storage/jwk_document_ext.rs
```diff
@@ -105,6 +106,7 @@ pub trait JwkDocumentExt: private::Sealed {
     storage: &Storage<K, I>,
     fragment: &str,
     options: &JwsSignatureOptions,
+    custom_claims: Option<Object>,
```
PhilippGackstatter, line 109: Add documentation for this parameter in the trait.

```
/// The `custom_claims` can be used to set additional claims on the resulting JWT.
```

PhilippGackstatter (approved):
Looks good to me, even though I wish there was a nicer way to facilitate these custom claims without polluting the API with something that is rarely going to be used by the average caller.

eike-hass (approved):

PR #1238, opened 2023-09-19
Add dedicated EdDSA verifier crate

PhilippGackstatter:
# Description of change

Add a dedicated EdDSA verifier crate which contains a single verifier for curve `Ed25519` but could be extended non-breakingly to also support `Ed448` later. Other verifier crates could be nicely added alongside it.

This PR also makes the validators fully agnostic over the verifier and not rely on the EdDSA JWS Verifier as the default. Previously the verifier was required even when not using it due to how we used generics. That was changed and the validation methods that don't rely on a verifier were moved to "Util" namespaces. The same applies to the Wasm bindings.

n/a

## Type of change

- [ ] Documentation Fix

Existing tests & clippy.

## Change checklist


examples/Cargo.toml
```diff
@@ -7,7 +7,8 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.62"
-identity_iota = { path = "../identity_iota", features = ["memstore", "domain-linkage", "stronghold"] }
```
abdulmth, line 10: Why the `stronghold` feature is removed, now the example doesn't work anymore. Not sure why CI didn't catch it, maybe because of `all-features` flag?

examples/Cargo.toml
```diff
@@ -7,7 +7,8 @@ publish = false
...
+identity_eddsa_verifier = { path = "../identity_eddsa_verifier" }
```
abdulmth, line 10: Why don't we include it in `identity_iota`?

abdulmth (changes requested):
Conflicts and some comments.

examples/Cargo.toml
PhilippGackstatter: The idea is to be fully agnostic over the verifier and also the storage (which is why stronghold was removed).
Storage and Verifier are things users have to select for themselves so while we offer some defaults, we don't necessarily have to ship them as part of the identity_iota crate.
The whole topic is very debatable, and we can reasonably go in either direction.

examples/Cargo.toml
PhilippGackstatter: Added stronghold crate to examples explicitly.

abdulmth (approved):

eike-hass (approved):

PR #1239, opened 2023-09-20
Allow custom `kid` to be set in JWS

PhilippGackstatter:
# Description of change

- Allow custom `kid` to be set in JWS.
- Allow a custom DID Url to be passed to the verification because the assumption that JWS kid == method ID does not always hold.
- Remove leftover `crits` parameter in Wasm bindings that was removed some time ago on the Rust side.
- Add `new Timstamp()` constructor, so this does not return a null pointer.
- Add tests for parsing the option types.
- Set the kid of a newly generated JWK to its SHA256 base64-encoded thumbprint, as recommended by did-core:
    - > It is RECOMMENDED that JWK kid values are set to the public key fingerprint.
      > https://www.w3.org/TR/did-core/#dfn-publickeyjwk

part of #1103 

## Type of change

Added new tests.

## Change checklist


abdulmth (approved):
Looks good to me, maybe it would be worth it to add the same Rust test to JS, maybe also test the case when `kid` isn't set if not tested somewhere else.

bindings/wasm/src/did/jws_verification_options.rs
```diff
@@ -65,4 +66,9 @@ interface IJwsVerificationOptions {
 
     /** Verify the signing verification method relationship matches this.*/
     readonly methodScope?: MethodScope;
+
+    /** The DID URl of the method, whose JWK should be used to verify the JWS.
```
eike-hass, line 70:
```suggestion
    /** The DID URL of the method, whose JWK should be used to verify the JWS.
```

eike-hass (approved):

PR #1245, opened 2023-09-27
Allow arbitrary JWS header parameters

abdulmth:
# Description of change
Now `JwsSignatureOptions` has the option to set arbitrary header parameters https://datatracker.ietf.org/doc/html/rfc7519#section-5

### DIFF
--- a/identity_jose/src/jws/header.rs
+++ b/identity_jose/src/jws/header.rs
@@ -185,10 +185,7 @@ mod tests {
       "tst-value".to_owned()
     );
 
-    assert_eq!(
-      header1.custom().unwrap().get("test-bool").unwrap().as_bool().unwrap(),
-      false
-    );
+    assert!(!header1.custom().unwrap().get("test-bool").unwrap().as_bool().unwrap());
     assert!(header1.has("test"));
     assert!(!header1.has("invalid"));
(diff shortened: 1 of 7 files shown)


identity_jose/src/jws/header.rs
```diff
@@ -77,20 +85,54 @@ impl JwsHeader {
...
+            return false;
+          }
+        } else {
+          return true;
+        }
+      }
+    } else {
+      return true;
+    }
+    true
```
PhilippGackstatter, line 135:
```suggestion
  /// Returns `true` if none of the fields are set in both `self.custom` and `other.custom`.
  fn is_custom_disjoint(&self, other: &JwsHeader) -> bool {
    match (&self.custom, &other.custom) {
      (Some(self_custom), Some(other_custom)) => {
        for self_key in self_custom.keys() {
          if other_custom.contains_key(self_key) {
            return false;
          }
        }
        true
      }
      _ => true,
    }
  }
```

This is a bit easier to read and we don't evaluate `if let Some(custom) = other.custom() {` in every loop iteration.

PhilippGackstatter (approved):

eike-hass (approved):

Issue #1150, opened 2023-03-28
[Task] Investigate potential for deadlocks or panics with the functionality introduced in #1148


olivereanderson:
# Description
#1148 introduces asynchronous methods on `CoreDocument` and `IotaDocument` which means that we had to introduce locking internally. Currently we use `toki::sync::RwLock`(https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) because it has both `async` read/write and `blocking_read` / `blocking_write` i.e. it can be used in both async (concurrent) and blocking contexts. 

I am however not sure whether this solution is safe with regards to panics and/or deadlocks occurring. In particular we should investigate  what happens under the following conditions: 

Use [`Promise.any`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any) or [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) with two promises. The first should be a promise returned from an  `async` method that writes to a `CoreDocument` (this could be `generateMethod` for instance) and the second promise should be constructed from an async javascript function that internally calls any non-async method on the same document (say `CoreDocument.id`).  This is interesting because if the first promise yields mid-execution while holding a lock the second promise that now gets to run will call a method on the document which in turn calls [blocking_read](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#method.blocking_read). According to the documentation it should panic in an asynchronous context, but I am not sure what happens when said asynchronous context is controlled by Javascript.  Maybe something like [this](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html) can be injected into one of the asynchronous method used to artificially introduce a yield when investigating this. 

## What to do In the case of panics and/or Deadlocks? 
I can think of the following (hopefully possible) fixes with different trade-offs. 

-  Replace all blocking reads and writes with async versions which means that most methods on the document
will need to be made `async` (in the bindings).
- Make all non-async methods fallible by replacing `blocking_read` (resp. `blocking_write`) with `try_read`  (resp. `try_write`).  
- Make all methods that require mutable access to the document non-async and replace all uses of `blocking_write` with `try_write`.  

I think making more methods async (which is definitely a breaking change) is the best fix if it works. 

## Motivation
Users of the bindings should not run into surprises when using common Javascript functionality such as `Promise.any` / `Promise.all`. 

## Resources
[...]

PR #1258, opened 2023-10-18
Fix wasm panic caused by a race condition in `IotaDocument` and `CoreDocument`

abdulmth:
# Description of change

fixes #https://github.com/iotaledger/identity.rs/issues/1150

This PR change the methods requiring read/write to be fallible if the lock is acquired to prevent uncontrolled crashes, since blocking is not supported for `RwLock` in wasm.

### DIFF
--- a/bindings/wasm/tests/storage.ts
+++ b/bindings/wasm/tests/storage.ts
@@ -472,3 +472,61 @@ describe("#OptionParsing", function() {
             });
         });
 });
+
+describe("#documents throw error on concurrent access", async function() {
+    const wait: any = (ms: any) => new Promise(r => setTimeout(r, ms));
+
+    class MyJwkStore extends JwkMemStore {
+        constructor() {
+            super();
+        }
+
+        async generate(keyType: string, algorithm: JwsAlgorithm) {
+            await wait(10000);
+            return await super.generate(keyType, algorithm);
+        }
+    }
+    it("CoreDocument", async () => {
+        const document = new CoreDocument({ id: "did:example:123" });
+        const storage = new Storage(new MyJwkStore(), new KeyIdMemStore());
+        const insertPromise = document.generateMethod(
+            storage,
+            JwkMemStore.ed25519KeyType(),
+            JwsAlgorithm.EdDSA,
+            "#key-1",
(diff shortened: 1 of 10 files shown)


bindings/wasm/tests/storage.ts
```diff
@@ -472,3 +472,61 @@ describe("#OptionParsing", function() {
...
+
+        const idPromise = wait(10).then((_value: any) => {
+            return document.id();
+        });
+
+        try {
+            await Promise.all([insertPromise, idPromise]);
+        } catch (e: any) {
+            assert.equal(e.name, "TryLockError");
+        }
```
PhilippGackstatter, line 508: You should do something like this here, otherwise the test could silently pass without throwing:

```ts
    let resolvedToError = false
    try {
      await Promise.all([insertPromise, idPromise]);
    } catch (e: any) {
      resolvedToError = true
      assert.equal(e.name, "TryLockError");
    }
    assert.ok(resolvedToError, "Promise.all did not throw an error")
```

bindings/wasm/tests/storage.ts
```diff
@@ -472,3 +472,61 @@ describe("#OptionParsing", function() {
             });
         });
 });
+
+describe("#documents throw error on concurrent access", async function() {
```
PhilippGackstatter, line 476:
```suggestion
describe("#Documents throw error on concurrent synchronous access", async function() {
```

PhilippGackstatter (approved):
Looks good to me. I think that's the most reasonable way to deal with this.

If I do a Promise.all on multiple promises each of which does a `createJws` (so an async read), that should be fine because it would use asynchronous rather than synchronous locking, right? That's how I understood #1150.

bindings/wasm/tests/storage.ts
PhilippGackstatter: Also applies to the other test.

abdulmth:
> If I do a Promise.all on multiple promises each of which does a createJws (so an async read), that should be fine because it would use asynchronous rather than synchronous locking, right?

yes that's true.

eike-hass (approved):

Issue #1265, opened 2023-11-28
[Task] Validate origin for domain linkage


eike-hass:
## Description
We should verify during creating and parsing of the domain linkage service endpoint that the origin filed only contains origins and not full URL.

## Motivation
Bring our implementation in line with the specification and reduce the potential for unexpected behavior 

## Resources
https://identity.foundation/.well-known/resources/did-configuration/#origins
https://html.spec.whatwg.org/multipage/browsers.html#origin
https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #1267, opened 2023-11-29
Validate domain-linkage URL making sure they only include an origin

UMR1352:
# Description of change
See issue #1265

Fixes issue #1265 

## Type of change

- [ ] Documentation Fix

Added a new test that makes sure only URLs with no path, no query and no fragment are accepted

## Change checklist


UMR1352:
@abdulmth I'm not exactly satisfied by these changes, but I wanted to ask your opinion first. What I'd like to do is extract the URL validation logic used here into a new type `DomainLinkageOrigin` and use that instead of an `Url`. This could definitely be more type-safe and just as ergonomic as `Url` if provided with the right interface. What do you say?

identity_credential/src/domain_linkage/domain_linkage_configuration.rs
```diff
@@ -128,6 +128,9 @@ mod __fetch_configuration {
       if domain.scheme() != "https" {
         return Err(DomainLinkageError("domain` does not use `https` protocol".into()));
       }
+      if domain.path() != "/" || domain.query().is_some() || domain.fragment().is_some() {
+        return Err(DomainLinkageError("domain mut not include any path, query of fragment".into()));
```
abdulmth, line 132:
```suggestion
        return Err(DomainLinkageError("domain must not include any path, query or fragment".into()));
```

identity_credential/src/domain_linkage/domain_linkage_credential_builder.rs
```diff
@@ -139,6 +144,18 @@ mod tests {
...
     assert!(matches!(err, Error::DomainLinkageError(_)));
   }
+  #[test]
+  fn test_builder_origin_is_a_url() {
+    let issuer: CoreDID = "did:example:issuer".parse().unwrap();
+    let err: Error = DomainLinkageCredentialBuilder::new()
+      .issuance_date(Timestamp::now_utc())
+      .expiration_date(Timestamp::now_utc())
+      .issuer(issuer)
+      .origin(Url::parse("https://example.com/foo?bar=420#baz").unwrap())
```
abdulmth, line 154: I would add other test cases for `"https://example.com/?bar=420#baz"` and `"https://example.com/#baz"`

identity_credential/src/domain_linkage/domain_linkage_configuration.rs
```diff
@@ -128,6 +128,9 @@ mod __fetch_configuration {
       if domain.scheme() != "https" {
         return Err(DomainLinkageError("domain` does not use `https` protocol".into()));
       }
+      if domain.path() != "/" || domain.query().is_some() || domain.fragment().is_some() {
```
abdulmth, line 131: Not sure about the part `domain.path() != "/"` would it return false if the URL was `http://www.example.com/` and not `http://www.example.com` ?  we can probably also add positive tests for these cases too.

abdulmth (commented):
I think adding a new type for internal use is a bit too much (exposing it would be a breaking change, no?). We can just add a function like `isOriginOnly` or something similar to avoid the duplication, but it looks good otherwise.

identity_credential/src/domain_linkage/utils.rs
```diff
@@ -0,0 +1,19 @@
+use identity_core::common::Url;
```
abdulmth, line 1:
```suggestion
// Copyright 2020-2023 IOTA Stiftung
// SPDX-License-Identifier: Apache-2.0

use identity_core::common::Url;
```

abdulmth (commented):
I think one more place to add this check is `LinkedDomainService::check_structure()` since the  endpoints must also be origins only otherwise the service is invalid.

abdulmth (approved):

PR #1268, opened 2023-11-30
Support Selective Disclosure SD-JWT

abdulmth:
Add Rust and Wasm support for verifying selective disclosures JWT and key binding JWT using https://github.com/iotaledger/sd-jwt-payload.


identity_credential/src/validator/jwt_credential_validation/jwt_credential_validator.rs
UMR1352, line 1: I'm not too convinced with how the validator has changed. Having to explicitly pass the optional `SdObjectDecoder` doesn't feel really confortable. Don't you think it might be worth having `SdObjectDecoder` work behind the scene? The APIs for `JwtCredentialValidator` remains the same, but internally use `SdObjectDecoder` if the JWT received as input is a SD-JWT. If a user knows for sure that they are working with a SD-JWT then they can use the `SdObjectDecoder` directly, maybe through some utility functions or even an extension trait provided by the library to make things more ergonomic. 
Probably we can discuss a better way to handle this. What concerns me is that this won't scale well if we want to handle the validation of even more JWT types (say ZK-JWT for instance), as it could require to add yet another optional parameter to every method call.

identity_credential/src/validator/sd_jwt/kb_validation_options.rs
```diff
@@ -0,0 +1,67 @@
...
+  /// Options which affect the verification of the signature on the KB-JWT.
+  pub jws_options: JwsVerificationOptions,
+  /// Declares that the KB-JWT is considered invalid if the `iat` value in the claims is
+  /// earlier than this timestamp.
+  pub earliest_issuance_date: Option<Timestamp>,
+  /// Declares that the KB-JWT is considered invalid if the `iat` value in the claims is
+  /// later than this timestamp.
+  /// Uses the current timestamp during validation if not set.
+  pub latest_issuance_date: Option<Timestamp>,
+}
```
UMR1352, line 27: Maybe you might want to avoid serializing to `null` in case of a `None` value

identity_credential/src/validator/jwt_credential_validation/jwt_credential_validator.rs
UMR1352: I see now you have created an `SdJwtValidator` that only validates SD-JWT encoded credential. Even more of a reason to avoid polluting the APIs of `JwtCredentialValidator` if the two have different concerns.

abdulmth: The library's API didn't really change, since none of the public methods changed. `validate_extended` is not public.
I agree with you that this does not scale well. But If we want to separate the concerns, we pretty much have to duplicate the majority of the `JwtCredentialValidator`. 

> Don't you think it might be worth having SdObjectDecoder work behind the scene?

It doesn't work, it's tightly coupled with the signature. You can not reuse `JwtCredentialValidator` if you do the decoding yourself. You have to verify the signature first, which is a deep down part of the `JwtCredentialValidator`.

So all in all, we either have to duplicate `JwtCredentialValidator`, or we have to rewrite it completely to make it more generic (not solved problem), or we keep the current solution and make it a future problem :D, But for ZK, I don't think it would use any of the `JwtCredentialValidator`. I don't think it would have a big overlap for that to make sense.

My concern is only that the `sd_jwt_payload` is not an optional dependency if users only want `JwtCredentialValidator`. But I kept all selective disclosure concerns under a flag, in case we want to rewrite or rethink the architecture of the internal implementation of the `JwtCredentialValidator`.

UMR1352: @abdulmth the last thing you mentioned is definitely something worth discussing. I couldn't have thought about it, thanks for pointing it out.

UMR1352 (approved):

identity_credential/src/validator/jwt_credential_validation/jwt_credential_validator.rs
```diff
@@ -286,56 +290,66 @@ impl<V: JwsVerifier> JwtCredentialValidator<V> {
...
+    //     })?)
+    //     .to_string();
+    //     CredentialJwtClaims::from_json(&decoded).map_err(|err| {
+    //       JwtValidationError::CredentialStructure(crate::Error::JwtClaimsSetDeserializationError(err.into()))
+    //     })?
+    //   } else {
+    //     CredentialJwtClaims::from_json_slice(&claims).map_err(|err| {
+    //       JwtValidationError::CredentialStructure(crate::Error::JwtClaimsSetDeserializationError(err.into()))
+    //     })?
+    //   };
```
UMR1352: forgot to remove this commented code

UMR1352:
We might want to further split some validation steps into smaller reusable and composable functions which probably better belong to `JwtCredentialValidationUtils` rather than `JwtCredentialValidator`

bindings/wasm/src/iota/iota_document.rs
```diff
@@ -704,7 +704,39 @@ impl WasmIotaDocument {
   ///
   /// Upon success a string representing a JWS encoded according to the Compact JWS Serialization format is returned.
   /// See [RFC7515 section 3.1](https://www.rfc-editor.org/rfc/rfc7515#section-3.1).
+  ///
+  /// @deprecated Use `createJws()` instead.
```
abdulmth, line 708: typo in the latest release. I added an ew method and deprecated this one

abdulmth:
> We might want to further split some validation steps into smaller reusable and composable functions which probably better belong to JwtCredentialValidationUtils rather than JwtCredentialValidator

We can do that in a separate PR, this one is getting a bit too big. 

I already adapted your [PR](https://github.com/iotaledger/identity.rs/pull/1274) here, we can close that one I think.

eike-hass (commented):
looks really good a couple minor suggestions. Please also add the new example to `bindings/wasm/examples/src/tests` to run it in CI.

bindings/wasm/src/credential/credential.rs
```diff
@@ -215,6 +218,26 @@ impl WasmCredential {
   pub fn set_proof(&mut self, proof: Option<WasmProof>) {
     self.0.set_proof(proof.map(|wasm_proof| wasm_proof.0))
   }
+
+  /// Serializes the `Credential` as a JWT claims set
+  /// in accordance with [VC Data Model v1.1](https://www.w3.org/TR/vc-data-model/#json-web-token).
```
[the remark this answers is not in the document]
abdulmth: They don't have it in the Data Model 2.0, it's in [Securing Verifiable Credentials using JOSE and COSE](https://www.w3.org/TR/vc-jose-cose/). 
I'm not sure what we can do here, it affects our JWT implementation for credentials everywhere, not only for SD JWT.

abdulmth:
> Would there ever be an additional Proof in the credential? IS that something that need to demonstrate?

It's possible to add a proof property to a credential, but how is that relevant to SD JWT? We use JWT to validate the signature here so we don't need a proof property since it's not relevant to SD JWT 🤷🏻

bindings/wasm/examples/src/1_advanced/6_sd_jwt.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    // ===========================================================================
+    // Step 3: Issuer sends the JWT and the disclosures to the holder.
+    // ===========================================================================
+
+    // One way to send the JWT and the disclosures, is by creating an SD-JWT with all the
+    // disclosures.
+    const strDisclosures = disclosures.map(disclosure => disclosure.toEncodedString());
+
+    let sdJwt = new SdJwt(jws.toString(), strDisclosures, undefined).presentation();
```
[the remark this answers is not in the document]
abdulmth: Done, I kept it originally to hint that users can optionally add the KB. but it's demonstrated further down in the code so it's fine.

eike-hass:
> > Would there ever be an additional Proof in the credential? IS that something that need to demonstrate?
> 
> It's possible to add a proof property to a credential, but how is that relevant to SD JWT? We use JWT to validate the signature here so we don't need a proof property since it's not relevant to SD JWT 🤷🏻
That makes sense, thank you! Then I think there is nothing for us to demonstrate or additionally consider.

bindings/wasm/src/credential/credential.rs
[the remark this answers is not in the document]
eike-hass: Sorry I wasn't clear. I was going for do we have any dependencies on Data Model 1.1 that we should abstract out already.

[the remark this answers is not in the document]
abdulmth: I don't think so, it can deal with any JSON value, so that wouldn't be an issue AFAIK.

eike-hass (approved):
Looks good, awesone job 💪

PR #1273, opened 2024-01-08
Feat/status list 2021

UMR1352:
# Description of change
Implementation of [StatusList2021](https://www.w3.org/TR/2023/WD-vc-status-list-20230427)

## Type of change
- [ ] Documentation Fix

Unit and integration tests

## Change checklist


examples/1_advanced/7_status_list_2021.rs
```diff
@@ -0,0 +1,72 @@
...
+use identity_iota::credential::status_list_2021::StatusList2021Entry;
+use identity_iota::credential::status_list_2021::StatusPurpose;
+use identity_iota::credential::Credential;
+use identity_iota::credential::Issuer;
+use identity_iota::credential::JwtCredentialValidatorUtils;
+use identity_iota::credential::JwtValidationError;
+use identity_iota::credential::StatusCheck;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
```
abdulmth, line 20: I suggest making the example include the entire use case, similar to other examples. Like creating DID, issuing a credential ... etc

examples/1_advanced/7_status_list_2021.rs
```diff
@@ -0,0 +1,72 @@
...
+  // }
+  let revocation_entry = StatusList2021Entry::new(
+    status_list_credential.id.clone().unwrap(),
+    StatusPurpose::Revocation,
+    420,
+  );
+  credential.credential_status = Some(revocation_entry.into());
+
+  // To revoke this credential we set the status of the 420th entry
+  status_list_credential.update_status_list(|status_list| status_list.set(420, true))?;
```
abdulmth, line 61: I suggest to re-expose the functions in the status list instead of the closure.

identity_credential/src/revocation/status_list_2021/status_list.rs
```diff
@@ -0,0 +1,144 @@
...
+  }
+  /// Returns the number of entries
+  #[allow(clippy::len_without_is_empty)]
+  pub const fn len(&self) -> usize {
+    self.0.len() * 8
+  }
+  /// Returns the status of the entry at `index` without bound checking
+  /// ## Panic:
+  /// * if `index` is greater than or equal to `self.len()`
+  pub const fn get_unchecked(&self, index: usize) -> bool {
```
abdulmth, line 49: I think the unchecked methods aren't really necessary.

identity_credential/src/revocation/status_list_2021/status_list.rs
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      self.0[i] &= 0b0111_1111 >> offset
+    }
+  }
+  /// Returns the status of the `index`-th entry, if it exists
+  pub fn get(&self, index: usize) -> Option<bool> {
+    (index < self.len()).then_some(self.get_unchecked(index))
+  }
+  /// Sets the status fo the `index`-th entry to `value`
+  pub fn set(&mut self, index: usize, value: bool) {
```
abdulmth, line 69: Better make this method fallible than silently not doing anything if index is out of bounds.

abdulmth (changes requested):
It looks great, thanks for working on it, some points that I noticed. This is not the full review, I will take a deeper look later.

identity_credential/src/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+impl StatusList2021Credential {
+  /// Returns the inner "raw" [`Credential`]
+  pub fn into_inner(self) -> Credential {
+    self.0
+  }
+  /// Returns the purpose of this status list
+  pub fn purpose(&self) -> StatusPurpose {
+    let subject = StatusList2021CredentialSubject::try_from_credential(&self.0).unwrap(); // Safety: `Self` has already been validated as a valid StatusList2021Credential
```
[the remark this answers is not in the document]
UMR1352: Yeah, it would do that and even more annoying it would require to clone data because you cannot create self referential types

identity_credential/src/credential/status.rs
```diff
@@ -7,11 +7,63 @@ use serde::Serialize;
...
+    match self {
+      #[cfg(feature = "revocation-bitmap")]
+      Self::StatusList2021(s) => s.r#type(),
+      Self::Other(s) => s.r#type(),
+    }
+  }
+}
+
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
-pub struct Status<T = Object> {
```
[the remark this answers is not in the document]
UMR1352: That's true, how should I address this?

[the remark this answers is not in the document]
abdulmth: Maybe change the name of the current `Status` to `StatusType` or something similar and keep the custom one as `Status`. it doesn't look great but it's better than breaking IMO. Maybe @eike-hass can have an opinion.

identity_credential/src/revocation/status_list_2021/status_list.rs
```diff
@@ -0,0 +1,144 @@
...
+
+use std::fmt::Display;
+
+use thiserror::Error;
+
+const DEFAULT_LIST_SIZE: usize = 16 * 1024;
+
+#[derive(Debug, Error, Eq, PartialEq)]
+/// [`std::error::Error`] type representing an invalidly encoded [`StatusList2021`]
+pub struct InvalidEncodedStatusList(String);
```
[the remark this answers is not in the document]
UMR1352: I've a different opinion on this. I think error types should be hand tailored to fit like a glove the failure they attempt to model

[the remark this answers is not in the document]
abdulmth: If we do that, we would have way too many error types in our crates, which would be very annoying. You can add to the documentation of the function, what error would be thrown at what point.

examples/1_advanced/7_status_list_2021.rs
UMR1352: @abdulmth checkout the new commit. Instead of exposing `StatusList2021`'s API, I've added a method to `StatusList2021Credential` that handles the revocation process as a whole, without the user having to manually create and set a `StatusList2021Entry` status

identity_credential/src/credential/status.rs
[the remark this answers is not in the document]
UMR1352: @abdulmth, the latest commit refactor things a bit (basically reverting all changes to `Status`). It's a bit annoying because the status of a parsed credential loses its type-safeness but at least this won't cause a breaking change

identity_credential/src/revocation/status_list_2021/entry.rs
```diff
@@ -0,0 +1,136 @@
...
+{
+    "id": "https://example.com/credentials/status/3#94567",
+    "type": "StatusList2021Entry",
+    "statusPurpose": "revocation",
+    "statusListIndex": "94567",
+    "statusListCredential": "https://example.com/credentials/status/3"
+}"#;
+
+  #[test]
+  fn entry_deserialization_works() {
```
abdulmth, line 130: It could be nice to check the rest of the values and add tests where the serialization fails, like when the `type` is wrong.

identity_credential/src/revocation/status_list_2021/entry.rs
```diff
@@ -0,0 +1,136 @@
...
+    self.status_purpose
+  }
+
+  /// Returns the index of this entry.
+  pub const fn index(&self) -> usize {
+    self.status_list_index
+  }
+
+  /// Returns the referenced [`StatusList2021Credential`]'s [`Url`].
+  pub const fn credential(&self) -> &Url {
```
abdulmth, line 111: Rename it to something more descriptive. `status_list_credential` for example.

identity_credential/src/revocation/status_list_2021/entry.rs
```diff
@@ -0,0 +1,136 @@
...
+
+use crate::credential::Status;
+
+use super::credential::StatusPurpose;
+
+const CREDENTIAL_STATUS_TYPE: &str = "StatusList2021Entry";
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, Serialize)]
+#[serde(transparent)]
+struct EntryType(&'static str);
```
abdulmth, line 16: It could have been easier to use serde's `deserialize_with` and check the string equality there, instead of introducing this struct. That would be much more readable IMO.

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,221 @@
...
+    use crate::error::WasmResult;
+    json.into_serde::<WasmStatusList2021Credential>().wasm_result()
+  }
+
+  #[wasm_bindgen(js_name = "toJSON")]
+  pub fn to_json(&self) -> Result<JsValue> {
+    use crate::error::WasmResult;
+    JsValue::from_serde(self).wasm_result()
+  }
+}
```
abdulmth, line 145:
```suggestion
```
Use `impl_wasm_json` and `impl_wasm_clome` instead.

abdulmth (changes requested):
Some more comments, I also would suggest moving the files of the old bitmap revocation method to a new directory and feature gate the status list into its own feature since it's not the same as bitmap, but it's under its feature.

abdulmth:
Also, please change the test number since the SD-JWT exampled is merged now, also Add the tests to the README files inside the examples directories, also to the node and cypress tests so they run for CI. Check the sd-jwt PR to see where they should be added.

identity_credential/src/revocation/status_list_2021/status_list.rs
```diff
@@ -0,0 +1,160 @@
...
+      Err(StatusListError::IndexOutOfBounds)
+    }
+  }
+
+  /// Attempts to parse a [`StatusList2021`] from a string, following the
+  /// [StatusList2021 expansion algorithm](https://www.w3.org/TR/2023/WD-vc-status-list-20230427/#bitstring-expansion-algorithm).
+  pub fn try_from_encoded_str(s: &str) -> Result<Self, StatusListError> {
+    use flate2::read::GzDecoder;
+    use identity_core::convert::Base;
+    use identity_core::convert::BaseEncoding;
```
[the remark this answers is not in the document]
UMR1352: I generally prefer not to pollute the module's namespace with stuff I only need in one function, but if that's the convention I'll follow it

identity_credential/src/revocation/status_list_2021/status_list.rs
```diff
@@ -0,0 +1,160 @@
...
+    use flate2::write::GzEncoder;
+    use flate2::Compression;
+    use identity_core::convert::Base;
+    use identity_core::convert::BaseEncoding;
+
+    let compressed_status_list = {
+      use std::io::Write;
+
+      let mut compressor = GzEncoder::new(vec![], Compression::best());
+      compressor.write_all(&self.0).expect("Out of memory");
```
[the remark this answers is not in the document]
UMR1352: Actually it won't even return an error, it will straight up panic if out of memory (although it's not even mentioned in the docs). Check out the [vec's write implementation](https://doc.rust-lang.org/src/std/io/impls.rs.html#395-398)

identity_credential/src/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,356 @@
...
+
+impl StatusList2021Credential {
+  /// Returns the inner "raw" [`Credential`].
+  pub fn into_inner(self) -> Credential {
+    self.0
+  }
+
+  /// Returns the id needed to fetch this credential.
+  pub fn id(&self) -> &Url {
+    self.0.id.as_ref().unwrap()
```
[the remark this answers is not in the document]
UMR1352: That's my bad, what I wanted to actually return is `credentialSubject.id` which is certainly there.

identity_credential/src/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    Ok(entry)
+  }
+
+  /// Sets the `index`-th entry to `value`
+  pub(crate) fn set_entry(&mut self, index: usize, value: bool) -> Result<(), StatusListError> {
+    let mut status_list = self.status_list()?;
+    status_list.set(index, value)?;
+    let OneOrMany::One(subject) = &mut self.0.credential_subject else {
+      unreachable!(); // Safety: already parsed as a valid `StatusList2021Credential`
```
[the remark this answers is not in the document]
UMR1352: But that part can't fail, that's the point of wrapping a `Credential` into `StatusList2021Credential` . But I see your standing I'll try to improve that part

identity_credential/src/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,356 @@
...
+    let _ = subject.properties.insert(
+      "encodedList".to_owned(),
+      serde_json::Value::String(status_list.into_encoded_str()),
+    );
+
+    Ok(())
+  }
+
+  /// Returns the status of the `index-th` entry.
+  pub fn entry(&self, index: usize) -> Result<bool, StatusListError> {
```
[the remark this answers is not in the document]
UMR1352: Done something similar to what you suggested

identity_credential/src/revocation/status_list_2021/credential.rs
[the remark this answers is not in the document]
UMR1352: I've refactored things a bit. `StatusList2021Credential` now parses the credential instead of validating it, extracting the credentialSubject.

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
UMR1352: Those macros only work with new type structs, cannot use them in this case.

identity_credential/src/revocation/status_list_2021/entry.rs
UMR1352: If I want to avoid `EntryType` I'll need a `&'static str type_` field in `StatusList2021Entry`. To deserialize that I can use a function with the signature:
`fn<'de, D: Deserializer>(d: D) -> Result<&'static, D::Error>`. It will work but there's an issue: `StatusList2021Entry` implements `Deserialize<'static>` instead of using any lifetime, this makes it impossible to deserialize for any `&str` other than `&'static str`, which is a no go.
Going the `EntryType` way solves this, another solution could be to simply clone and use `String` instead of borrowing from the deserializer.  Lemme know what do you think the best approach would be.

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
```diff
@@ -0,0 +1,157 @@
...
+    // Create signed JWT credential.
+    const credentialJwt = await issuerDocument.createCredentialJwt(
+        issuerStorage,
+        issuerFragment,
+        credential,
+        new JwsSignatureOptions(),
+    );
+    console.log(`Credential JWT > ${credentialJwt.toString()}`);
+
+    // Validate the credential using the issuer's DID Document.
```
eike-hass, line 113: Let's add a comment like we did in the Rust examples why we disable StatusChecks

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
```diff
@@ -0,0 +1,157 @@
...
+        FailFast.FirstError,
+    );
+
+    // ===========================================================================
+    // Revocation of the Verifiable Credential.
+    // ===========================================================================
+
+    // Update the status list credential.
+    // This revokes the credential's unique index.
+    statusListCredential.setCredentialStatus(credential, CREDENTIAL_INDEX, true);
```
eike-hass, line 129: I wonder if should (de-)serialize the StatutsList credential to make clear it would need to be retrieved

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
```diff
@@ -0,0 +1,157 @@
...
+    // Credential verification now fails.
+    try {
+        jwtCredentialValidator.validate(
+            credentialJwt,
+            issuerDocument,
+            validationOptions,
+            FailFast.FirstError,
+        );
+        JwtCredentialValidator.checkStatusWithStatusList2021(credential, statusListCredential, StatusCheck.Strict);
+        console.log("Revocation Failed!");
```
eike-hass, line 140: This is a bid ambiguous

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
```diff
@@ -0,0 +1,157 @@
...
+    } catch (e) {
+        console.log(`Error during validation: ${e}`);
+    }
+}
+
+// A custom JWS Verifier capabale of verifying EdDSA signatures with curve Ed25519.
+class Ed25519JwsVerifier implements IJwsVerifier {
```
eike-hass, line 147: Why do we need a custom Verifier here?

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,251 @@
...
+  pub fn subject_id(mut self, id: String) -> Result<WasmStatusList2021CredentialBuilder> {
+    let id = Url::parse(id).wasm_result()?;
+    self.0 = self.0.subject_id(id);
+
+    Ok(self)
+  }
+
+  /// Sets the expiration date of the credential.
+  #[wasm_bindgen(js_name = "expirationDate")]
+  pub fn expiration_date(mut self, time: WasmTimestamp) -> WasmStatusList2021CredentialBuilder {
```
eike-hass, line 207: Difficult topic, but we hardcode a VC Data Model 1.1 assumption here, could we also hardcode the 2.0 `validFrom`/`validUntil` or would that break in verification later?

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  /// Sets the given credential's status using the `index`-th entry of this status list.
+  /// Returns the created `credentialStatus`.
+  #[wasm_bindgen(js_name = "setCredentialStatus")]
+  pub fn set_credential_status(
+    &mut self,
+    credential: &mut WasmCredential,
+    index: usize,
+    flag: bool,
```
eike-hass, line 131: in Rust we call in value, is there a reason to differentiate?

eike-hass (changes requested):
Overall looking very good 💪 Left a view comments. Not everything needs to be addressed, but would like to have discussion on those topics

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
UMR1352: Copied this from `vc_revocation` example. How can I improve it? Shall I add a comment specifying that since the credential has been revoked `checkStatusWithStatusList2021` should throw an exception?

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
UMR1352: Changed this to `value`.

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
abdulmth: I don't think it's needed. Just use `new EdDSAJwsVerifier()` which is provided by the bindings instead.

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
```diff
@@ -0,0 +1,152 @@
...
+            validationOptions,
+            FailFast.FirstError,
+        );
+        /// Since the credential has been revoked, this validation step will throw an error.
+        JwtCredentialValidator.checkStatusWithStatusList2021(
+            credential,
+            refetchedStatusListCredential,
+            StatusCheck.Strict,
+        );
+        console.log("Revocation Failed!");
```
eike-hass, line 147:
```suggestion
        // In case the revocation failed for some reason we will hit this point
        console.log("Revocation Failed!");
```

eike-hass (approved):

bindings/wasm/src/credential/revocation/status_list_2021/entry.rs
```diff
@@ -0,0 +1,67 @@
...
+
+  /// Returns the referenced {@link StatusList2021Credential}'s url.
+  #[wasm_bindgen]
+  pub fn credential(&self) -> String {
+    self.0.status_list_credential().to_string()
+  }
+
+  /// Downcasts {@link this} to {@link Status}
+  #[wasm_bindgen(js_name = "toStatus")]
+  pub fn to_status(self) -> Result<JsValue> {
```
abdulmth, line 61: The TS function signature is `  toStatus(): any;` which is wrong in this case.

Add an external type to `src/credential/types.rs` 

```rust
  #[wasm_bindgen(typescript_type = "Status")]
  pub type WasmStatus;
``` 

and change the function to 
```rust
  #[wasm_bindgen(js_name = "toStatus")]
  pub fn to_status(self) -> Result<WasmStatus> {
    Ok(
      JsValue::from_serde(&Status::from(self.0))
        .wasm_result()?
        .unchecked_into::<WasmStatus>(),
    )
  }
```

bindings/wasm/src/credential/revocation/status_list_2021/entry.rs
```diff
@@ -0,0 +1,67 @@
...
+
+  /// Returns this `credentialStatus`'s `id`.
+  #[wasm_bindgen]
+  pub fn id(&self) -> String {
+    self.0.id().to_string()
+  }
+
+  /// Returns the purpose of this entry.
+  #[wasm_bindgen]
+  pub fn purpose(&self) -> WasmStatusPurpose {
```
abdulmth, line 43: the function signature is compiled to 
```typescript
/**
* Returns the {@link StatusPurpose} of this {@link StatusList2021Credential}.
* @returns {number}
*/
  purpose(): number;
``` 

Which is not correct. There's a limitation of wasm bindgen when it comes to enums. This is why we usually convert Rust's enums to JS classes. For example `WasmMethodScope`. 

It's also generally a good idea to take a look at the file `/node/identity_wasm.d.ts` to check if the compiled output of the bindings matches what you expect.

bindings/wasm/src/credential/revocation/status_list_2021/status_list.rs
```diff
@@ -0,0 +1,62 @@
...
+    self.0.get(index).map_err(|e| JsError::new(&e.to_string()).into())
+  }
+
+  /// Sets the value of the `index`-th entry.
+  #[wasm_bindgen]
+  pub fn set(&mut self, index: usize, value: bool) -> Result<()> {
+    self
+      .0
+      .set(index, value)
+      .map_err(|e| JsError::new(&e.to_string()).into())
```
abdulmth, line 45: If you add your error type to `impl_wasm_error_from!` in `error.rs` you probably can just use `.wasm_result()` on it.

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,251 @@
...
+
+  /// Returns the {@link StatusPurpose} of this {@link StatusList2021Credential}.
+  #[wasm_bindgen]
+  pub fn purpose(&self) -> WasmStatusPurpose {
+    self.inner.purpose().into()
+  }
+
+  /// Returns the state of the `index`-th entry, if any.
+  #[wasm_bindgen]
+  pub fn entry(&self, index: usize) -> Result<WasmCredentialStatus> {
```
abdulmth, line 151: Same thing here, resulted return type is `number`

bindings/wasm/src/credential/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,251 @@
...
+  pub fn set_credential_status(
+    &mut self,
+    credential: &mut WasmCredential,
+    index: usize,
+    value: bool,
+  ) -> Result<WasmStatusList2021Entry> {
+    let entry = self
+      .inner
+      .set_credential_status(&mut credential.0, index, value)
+      .map_err(|e| JsValue::from(JsError::new(&e.to_string())))?
```
abdulmth, line 136: same thing here and in some other places. `wasm_result` can be used.

bindings/wasm/src/credential/revocation/status_list_2021/entry.rs
UMR1352: That's weird, I've checked the locally generated `/node/identity_wasm.d.ts` and `StatusPurpose` is always used. I've tried building a couple times always with the same outcome.

examples/1_advanced/8_status_list_2021.rs
```diff
@@ -0,0 +1,178 @@
...
+  // ===========================================================================
+  // Revocation of the Verifiable Credential.
+  // ===========================================================================
+
+  // The issuer retrieves the status list credential.
+  let mut status_list_credential =
+    serde_json::from_str::<StatusList2021Credential>(status_list_credential_json.as_str()).unwrap();
+
+  // Set the value of the chosen index entry to true to revoke the credential
+  status_list_credential.set_credential_status(&mut credential, credential_index, true)?;
```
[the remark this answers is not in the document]
UMR1352: It's slightly more complex than that. A `StatusListCredential` has a purpose that cannot be changed after its creation. Therefore a `StatusListCredential` can only set or unset the status of a credential to its `statusPurpose`. If the `set_credential_status` method signature becomes:
```rust
fn set_credential_status(&mut self, c: &mut Credential, i: usize, status: CredentialStatus) -> Result<()>
```
we have lost some type correctness as we have to check that the passed `status` matches the `StatusListCredential`'s, thus the user is more prone to misusing the API.

identity_credential/Cargo.toml
```diff
@@ -49,6 +51,7 @@ default = ["revocation-bitmap", "validator", "credential", "presentation", "doma
 credential = []
 presentation = ["credential"]
 revocation-bitmap = ["dep:dataurl", "dep:flate2", "dep:roaring"]
+status-list-2021 = ["revocation-bitmap", "dep:serde-aux"]
```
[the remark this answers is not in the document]
UMR1352: It doesn't but it has a dependency to the `revocation-bitmap` feature. I didn't wanna play too much around feature-gates so the idea was to enable `revocation-bitmap` automatically if `status-list-2021` is enabled. Should I change this behaviour?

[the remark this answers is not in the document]
abdulmth: We probably can leave it as is for now.

bindings/wasm/src/credential/revocation/status_list_2021/entry.rs
abdulmth: I updated wasm bindgen on my machine and it seems to work fine, I think they fixed it in one of the latest versions. That's actually pretty nice since it's very useful for what we usually use. Great that we found out about this 👍🏻 You can ignore my comments regarding this.

identity_credential/src/revocation/status_list_2021/credential.rs
```diff
@@ -0,0 +1,406 @@
...
+    StatusList2021::try_from_encoded_str(&self.subject.encoded_list)
+  }
+
+  /// Sets the credential status of a given [`Credential`],
+  /// mapping it to the `index`-th entry of this [`StatusList2021Credential`].
+  pub fn set_credential_status(
+    &mut self,
+    credential: &mut Credential,
+    index: usize,
+    value: bool,
```
abdulmth, line 124:
```suggestion
    revoked_or_suspended: bool,
```

bindings/wasm/examples/src/1_advanced/7_status_list_2021.ts
```diff
@@ -0,0 +1,153 @@
...
+    StatusList2021Credential,
+    StatusList2021CredentialBuilder,
+    StatusList2021Entry,
+    StatusPurpose,
+    Storage,
+} from "@iota/identity-wasm/node";
+import { Client, MnemonicSecretManager, Utils } from "@iota/sdk-wasm/node";
+import { API_ENDPOINT, createDid } from "../util";
+
+export async function statusList2021() {
```
abdulmth, line 24: All the comments from previous reviews on the rust example should also be reflected on the Wasm example. We try to keep the rust and TS example as similar as we can.

bindings/wasm/src/credential/revocation/status_list_2021/entry.rs
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /// Returns the index of this entry.
+  #[wasm_bindgen]
+  pub fn index(&self) -> usize {
+    self.0.index()
+  }
+
+  /// Returns the referenced {@link StatusList2021Credential}'s url.
+  #[wasm_bindgen]
```
abdulmth, line 55:
```suggestion
  #[wasm_bindgen(js_name = "statusListCredential"]
```

examples/1_advanced/8_status_list_2021.rs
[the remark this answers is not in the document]
abdulmth: We could've combined them to `RevokedOrSuspended` since that represents the `1` value better and the purpose is not relevant anyways. I added a suggestion to change the `value` parameter to something more descriptive if you want to keep it the way it is.

abdulmth (approved):
Last minor comments

identity_credential/src/revocation/status_list_2021/credential.rs
abdulmth: I would say yes

abdulmth (approved):

PR #1274, opened 2024-01-11
decouple JwtValidator from SdJwtValidator

UMR1352:


abdulmth:
done in https://github.com/iotaledger/identity.rs/pull/1268

PR #1296, opened 2024-02-05
Update `sd-jwt-payload` dependency

abdulmth:
Update the dependency `sd-jwt-payload` to the version `0.2.0` and adapt its Wasm bindings.


eike-hass (approved):
LGTM! Could you provide a more complete title and description so it can serve a reference through the changelog?

abdulmth:
@eike-hass I updated the description. But I don't think this PR should be included in the Changelog, since it's part of introducing the SD-JWT feature to the `1.1` release.

PR #1304, opened 2024-02-14
Support multiple IOTA networks in the `Resolver`

abdulmth:
# Description of change
Add the method `attach_multiple_iota_handlers` to the `Resolver` which allows using multiple clients depending on the network the client uses. 

Having the input as a vec of tuple `([network name], Client)` instead of only clients, is to avoid async network calls just to have the network name. The use can do that manually to have better transparency. 

I think this approach is fine. But it might be improved in 2.0 since we can have breaking changes.

### DIFF
--- a/identity_resolver/src/error.rs
+++ b/identity_resolver/src/error.rs
@@ -68,4 +68,7 @@ pub enum ErrorCause {
     /// The method that is unsupported.
     method: String,
   },
+  /// No client attached to the specific network.
+  #[error("none of the attached clients support the netowk {0}")]
+  UnsupportedNetowrk(String),
 }
--- a/identity_resolver/src/error.rs
+++ b/identity_resolver/src/error.rs
@@ -69,6 +69,6 @@ pub enum ErrorCause {
     method: String,
   },
   /// No client attached to the specific network.
-  #[error("none of the attached clients support the netowk {0}")]
-  UnsupportedNetowrk(String),
+  #[error("none of the attached clients support the network {0}")]
+  UnsupportedNetwork(String),
 }
--- a/identity_resolver/src/resolution/resolver.rs
+++ b/identity_resolver/src/resolution/resolver.rs
@@ -286,7 +286,7 @@ mod iota_handler {
     ///
     /// # Arguments
     ///
-    /// * `clients` - A vector of tuples where each tuple contains the name of the network name and its corresponding
+    /// * `clients` - A collection of tuples where each tuple contains the name of the network name and its corresponding
     ///   client.
     ///
     /// # Examples
@@ -301,8 +301,11 @@ mod iota_handler {
     ///
     /// # Note
     ///
-    /// Using `attach_iota_handler` or `attach_handler` for the IOTA method would override all
+    /// - Using `attach_iota_handler` or `attach_handler` for the IOTA method would override all
     /// clients added in this method.
+    /// - This function does not validate the provided configuration. Ensure that the provided
+    /// network name corresponds with the client, possibly by using `client.network_name()`.
+    ///
     pub fn attach_multiple_iota_handlers<CLI>(&mut self, clients: Vec<(&'static str, CLI)>)
     where
       CLI: IotaClientExt + Send + Sync + 'static,
@@ -316,9 +319,9 @@ mod iota_handler {
           let client: &CLI = future_client
             .as_ref()
             .iter()
-            .find(|(netowrk, _)| *netowrk == did_network)
+            .find(|(network, _)| *network == did_network)
             .map(|(_, client)| client)
-            .ok_or(crate::Error::new(ErrorCause::UnsupportedNetowrk(
+            .ok_or(crate::Error::new(ErrorCause::UnsupportedNetwork(
               did_network.to_string(),
             )))?;
           client
--- a/identity_resolver/src/resolution/resolver.rs
+++ b/identity_resolver/src/resolution/resolver.rs
@@ -358,17 +358,14 @@ where
 
 #[cfg(test)]
 mod tests {
-  use identity_iota_core::block::address::Address;
   use identity_iota_core::block::output::AliasId;
   use identity_iota_core::block::output::AliasOutput;
   use identity_iota_core::block::output::OutputId;
   use identity_iota_core::block::protocol::ProtocolParameters;
-  use identity_iota_core::IotaClientExt;
   use identity_iota_core::IotaDID;
   use identity_iota_core::IotaDocument;
   use identity_iota_core::IotaIdentityClient;
   use identity_iota_core::IotaIdentityClientExt;
-  use iota_sdk::client::secret::SecretManager;
 
   use super::*;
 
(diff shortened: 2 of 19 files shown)


identity_resolver/src/resolution/resolver.rs
```diff
@@ -277,6 +279,57 @@ mod iota_handler {
...
+            )))?;
+          client
+            .resolve_did(&did)
+            .await
+            .map_err(|err| crate::Error::new(ErrorCause::HandlerError { source: Box::new(err) }))
+        }
+      };
+
+      self.attach_handler(IotaDID::METHOD.to_owned(), handler);
+    }
```
UMR1352, line 332:
```suggestion
    pub fn attach_multiple_iota_handlers<CLI, I>(&mut self, clients: I)
    where
      CLI: IotaClientExt + Send + Sync + 'static,
      I: IntoIterator<Item = (&'static str, CLI)>,
    {
      let clients = Arc::new(clients.into_iter().collect::<HashMap<_, _>>());

      let handler = move |did: IotaDID| {
        let clients = clients.clone();
        async move {
          let did_network = did.network_str();
          let client: &CLI = clients
            .get(did_network)
            .ok_or(crate::Error::new(ErrorCause::UnsupportedNetowrk(
              did_network.to_string(),
            )))?;
          client
            .resolve_did(&did)
            .await
            .map_err(|err| crate::Error::new(ErrorCause::HandlerError { source: Box::new(err) }))
        }
      };

      self.attach_handler(IotaDID::METHOD.to_owned(), handler);
    }
```
Make the input generic over the actual collection, use an HashMap instead of checking all entries one by one

identity_resolver/src/resolution/resolver.rs
```diff
@@ -277,6 +279,57 @@ mod iota_handler {
...
       self.attach_handler(IotaDID::METHOD.to_owned(), handler);
     }
+
+    /// Convenience method for attaching multiple handlers responsible for resolving IOTA DIDs
+    /// on multiple networks.
+    ///
+    ///
+    /// # Arguments
+    ///
+    /// * `clients` - A vector of tuples where each tuple contains the name of the network name and its corresponding
```
UMR1352, line 289:
```suggestion
    /// * `clients` - A collection of tuples where each tuple contains the name of the network name and its corresponding
```

UMR1352 (changes requested):
Nice job tackling this one!

identity_resolver/src/error.rs
```diff
@@ -68,4 +68,7 @@ pub enum ErrorCause {
     /// The method that is unsupported.
     method: String,
   },
+  /// No client attached to the specific network.
+  #[error("none of the attached clients support the netowk {0}")]
+  UnsupportedNetowrk(String),
```
UMR1352, line 73:
```suggestion
  #[error("none of the attached clients support the network {0}")]
  UnsupportedNetwork(String),
```

UMR1352 (changes requested):

eike-hass:
> to avoid async network calls just to have the network name.. The user can do that manually to have better transparency.
As I understand it the network info is being cached, so if the user initialized the client before it should already be cached. If they didn't, we would need to wait for the info, but would be sure that the client can reach their node.
The only I think explicit naming would be advantageous is to detect misconfiguration (user wanted shimmer mainnet, but passed a testnet client), but to detect that we would need to also resolve the network info.
What do you guys think?

UMR1352:
@eike-hass that would be ideal, but the network info needs to be fetched asynchronously. We'd have to change the method to an async one in order to check that the couple `(network, client)` is valid.

abdulmth:
Whether it's cashed or not, getting the network name is async, and it can throw all the IO errors, which is probably not expected in such an attach method, and as Enrico said the method needs to be async. Also misconfiguration can be easily detected with one resolution on that network, or avoided by using `client.get_network_name` to set the network name in the tuple.

eike-hass:
Understood, but as we newly introduce the method, we could make it async, no? The argument to keep it consistent with other `attach_handler` is very valid. Might be a case to make all async with a breaking update though.

eike-hass:
> Also misconfiguration can be easily detected with one resolution on that network, or avoided by using client.get_network_name to set the network name in the tuple.

Can we demonstrate this in an example?

identity_resolver/src/resolution/resolver.rs
abdulmth: I think collecting the elements back into a `Hashmap` defeats the purpose of using generics here, and I prefer the cleaner function signature since users will most likely create the vector specifically for this function, Also the hash map will not bring any measurable performance benefits since it's very unlikely that the user will add more than 2 or 3 clients.

UMR1352: I agree that using an HashMap won't result in a performance improvement (probably it deteriorates it if the number of entries is very small) but I still think it is the right collection for the job as it has an implicit "lookup" implication - which is what the method does. I also agree that users will probably simply use a Vec 99% of the time but it doesn't hurt to account for other possible collections. I'm down for whatever you choose

identity_resolver/src/resolution/resolver.rs
```diff
@@ -277,6 +279,59 @@ mod iota_handler {
...
+
+      let handler = move |did: IotaDID| {
+        let future_client = arc_clients.clone();
+        async move {
+          let did_network = did.network_str();
+          let client: &CLI = future_client
+            .as_ref()
+            .iter()
+            .find(|(network, _)| *network == did_network)
+            .map(|(_, client)| client)
```
UMR1352, line 322:
```suggestion
            .find_map(|(network, client)| (*network == did_network).then_some(client))
```

eike-hass:
> > Also misconfiguration can be easily detected with one resolution on that network, or avoided by using client.get_network_name to set the network name in the tuple.
> 
> Can we demonstrate this in an example?

As discussed out-of-band an example is not needed

UMR1352 (approved):

identity_resolver/src/resolution/resolver.rs
```diff
@@ -277,6 +279,58 @@ mod iota_handler {
...
+    /// // Assume `smr_client` and `iota_client` are instances IOTA clients `iota_sdk::client::Client`.
+    /// attach_multiple_iota_handlers(vec![("smr", smr_client), ("iota", iota_client)]);
+    /// ```
+    ///
+    /// # See Also
+    /// - [`attach_handler`](Self::attach_handler).
+    ///
+    /// # Note
+    ///
+    /// - Using `attach_iota_handler` or `attach_handler` for the IOTA method would override all
```
eike-hass, line 304:
```suggestion
    /// - Using `attach_iota_handler` after `attach_handler` (or the other way around) would override all
```

identity_resolver/src/resolution/resolver.rs
```diff
@@ -277,6 +279,58 @@ mod iota_handler {
...
+    /// clients added in this method.
```
eike-hass, line 305:
```suggestion
    /// previously added clients.
```

eike-hass (approved):

identity_resolver/src/resolution/resolver.rs
abdulmth: This is not accurate, it's using `attach_iota_handler` or `attach_handler` after `attach_multiple_iota_handlers` would overwrite the clients.

eike-hass: gotcha, let's go with that

Issue #1302, opened 2024-02-12
[Request] Retrieve public key from `JwkStorage` by `KeyId`


daniel-mader:
## Description
Add public key retrieval to the `JwkStorage` trait by supplying a `KeyId` and receiving a JWK with public params. Proposed function signature (based on the other functions such as `generate`, `insert`, `sign`): `async fn get(&self, key_id: &KeyId) -> KeyStorageResult<Jwk>`.

## Motivation
When using identity.rs' interfaces to manage DID documents and keys for other DID methods (such as `did:key` or `did:jwk`), direct access to the public key parameters through the Storage interface is desired to create the method id. Other options are retrieving the key material directly from the actual storage implementation, this would make the usage of the Storage interface inconvenient though.

## Requirements
1. expose a `get(key_id)` function in `JwkStorage` trait
2. implement that function in `identity_stronghold`

## Open questions (optional)
n/a

## Are you planning to do it yourself in a pull request?
No

PR #1311, opened 2024-02-20
Add `get_public_key` for `StrongholdStorage`

abdulmth:
Retrieve public keys from `StrongholdStorage` by `KeyId`.

closes https://github.com/iotaledger/identity.rs/issues/1302

This function can't be added to the trait since that would be a breaking change and it doesn't seem necessary to force other implementations to add this function.


eike-hass (approved):
I like the idea of adding it to the concrete Stronghold implementation, rather then the storage interface/trait!

UMR1352 (approved):

PR #1334, opened 2024-03-12
Allow arbitrary verification methods

UMR1352:
# Description of change
Add support for arbitrary (custom) verification methods.

Fixes issue #1322

## Type of change
- [ ] Documentation Fix

New unit tests for the custom verification method data.

## Change checklist

### DIFF
--- a/identity_verification/src/verification_method/method.rs
+++ b/identity_verification/src/verification_method/method.rs
@@ -20,6 +20,7 @@ use crate::verification_method::MethodBuilder;
 use crate::verification_method::MethodData;
 use crate::verification_method::MethodRef;
 use crate::verification_method::MethodType;
+use crate::CustomMethodData;
 use identity_did::CoreDID;
 use identity_did::DIDUrl;
 use identity_did::DID;
@@ -247,8 +248,8 @@ impl KeyComparable for VerificationMethod {
 }
 
 // Horrible workaround for a tracked serde issue https://github.com/serde-rs/serde/issues/2200. Serde doesn't "consume"
-// the input when deserializing flattened enums (MethodData in this case) cousing duplication of data (in this case
-// it ends up in the properties object). This workaround simply remove the duplication.
+// the input when deserializing flattened enums (MethodData in this case) causing duplication of data (in this case
+// it ends up in the properties object). This workaround simply removes the duplication.
 #[derive(Deserialize)]
 struct _VerificationMethod {
   #[serde(deserialize_with = "deserialize_id_with_fragment")]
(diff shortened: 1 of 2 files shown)


lucagiorgino (commented):
Hi 👋🏻, 
I checked the code and didn't find anything worth mentioning. I was able to test this new custom type for my needs, and it works properly. 

The only issue I noticed is that the document is not displayed in the Explorer (`undefined` is displayed) . However, this is unrelated to the code. It's likely that the Explorer needs to be updated to use the next version of the library to reconstruct the DID document correctly.

Example: https://explorer.shimmer.network/testnet/addr/rms1pqqcmsxfsupcmyjreysaewnynlvgu25tdm4csp628ejmak9xscqavlzhfvn?tab=DID

bindings/wasm/src/verification/wasm_method_data.rs
```diff
@@ -98,3 +100,28 @@ impl From<MethodData> for WasmMethodData {
     WasmMethodData(data)
   }
 }
+
+/// A custom verification method data format.
+#[wasm_bindgen(js_name = CustomMethodData, inspectable)]
+pub struct WasmCustomMethodData(pub(crate) CustomMethodData);
```
abdulmth, line 106: Add `impl_wasm_clone!` and `impl_wasm_json!`

identity_verification/src/verification_method/method.rs
```diff
@@ -245,3 +245,40 @@ impl KeyComparable for VerificationMethod {
     self.id()
   }
 }
+
+// Horrible workaround for a tracked serde issue https://github.com/serde-rs/serde/issues/2200
```
abdulmth, line 249: Maybe add more explanation why this is needed and how it's used.

eike-hass:
> Hi 👋🏻, I checked the code and didn't find anything worth mentioning. I was able to test this new custom type for my needs, and it works properly.
> 
> The only issue I noticed is that the document is not displayed in the Explorer (`undefined` is displayed) . However, this is unrelated to the code. It's likely that the Explorer needs to be updated to use the next version of the library to reconstruct the DID document correctly.
> 
> Example: https://explorer.shimmer.network/testnet/addr/rms1pqqcmsxfsupcmyjreysaewnynlvgu25tdm4csp628ejmak9xscqavlzhfvn?tab=DID

Your assumption is correct, the explorer needs to be updated to the upcoming version of the library to support arbitrary verification methods. Not the prettiest error state though 🙈

lucagiorgino (approved):

identity_verification/src/verification_method/method.rs
```diff
@@ -245,3 +245,42 @@ impl KeyComparable for VerificationMethod {
...
+    let _VerificationMethod {
+      id,
+      controller,
+      type_,
+      data,
+      mut properties,
+    } = value;
+    let data_json = serde_json::to_value(&data).unwrap();
+    let data_type = data_json.as_object().unwrap().into_iter().next().unwrap().0;
+    properties.remove(data_type);
```
wulfraem, line 276: Hm, I guess we can just ignore this rant here, but couldn't bring myself to not complain about the `.unwrap()` here. :D

Could this be refactored into a match/if-let version that just removes the type (also adressed by its name instead of its relative position in the deserialized JSON) and does nothing if for whatever reasons the field couldn't be found?

Sorry for stirring up a (most probably useless) discussion but I've to be reaaally careful with `.unwrap()`s in library code. ^^

wulfraem (changes requested):

wulfraem (approved):

identity_verification/src/verification_method/method.rs
wulfraem: Looks good to me. ^^

PR #1381, opened 2024-06-18
chore(grpc): enhance domain grpc services

itsyaasir:
# Description of change

## Type of change

- [ ] Documentation Fix

## Change checklist


UMR1352 (approved):
Looking good! Do we really care about the order in which domains are validated? If that's not the case let's validate them parallely

bindings/grpc/src/services/domain_linkage.rs
wulfraem, line 1: Most probably bike shedding, I guess, but would it make sense to align the `use` statements and proto file includes a bit? Skimming through the existing service files, we might derive the following rules:

- proto file derived modules should start with "_", e.g.

```rs
#[allow(clippy::module_inception)]
mod _endpoint {
  tonic::include_proto!("endpoint");
}
```

- types from those are imported with the local prefix, **but** sorted into the non-std, non-local imports

The second point is not really beautiful, but ensures, that automatically adding missing imports (e.g. via vscode) don't get added to the "external" block and have to be moved to an internal block below it, and have to be rewritten (in case we'd go for a `self::` or `crate::` prefix).

Open for ideas here, @UMR1352, @itsyaasir. ^^

Just noticed, that we have `use crate` and `use domain_linkages` import leading to the same set of types, so started comparing the other imports a bit. ^^;;

examples/1_advanced/6_domain_linkage.rs
```diff
@@ -69,7 +71,7 @@ async fn main() -> anyhow::Result<()> {
 
   // The DID should be linked to the following domains.
   let domain_1: Url = Url::parse("https://foo.example.com")?;
-  let domain_2: Url = Url::parse("https://bar.example.com")?;
+  let domain_2: Url = Url::parse("https://foo.example.com")?;
```
wulfraem, line 74: Should both domains be identical here?

wulfraem (changes requested):

eike-hass (approved):

wulfraem (approved):

PR #1435, opened 2024-11-22
Versioning of on-chain `Identity`

UMR1352:
# Description of change
Add support for versioning on-chain `Identity`s in order to allow the creation of backwards incompatible changes to `Identity`.

# Requirements
Should be merged after #1430 

## Type of change
- [ ] Documentation Fix


identity_iota_core/packages/iota_identity/sources/identity.move
```diff
@@ -23,6 +24,10 @@ module iota_identity::identity {
     const EInvalidThreshold: u64 = 2;
     /// The controller list must contain at least 1 element.
     const EInvalidControllersList: u64 = 3;
+    /// Idenity cannot does not need an upgrade.
+    const ENoUpgrade: u64 = 4;
```
wulfraem, line 28: Hm, would something like `EAlreadyUpgraded` make more sense here? For some reason, I find "no upgrade" a bit confusing. But this could also be due to me being confused, or that the name might make more sense for me with future updates. ^^

Also, the comment has some artifacts from updating the wording. ^^

identity_iota_core/packages/iota_identity/sources/proposals/upgrade.move
```diff
@@ -0,0 +1,9 @@
+module iota_identity::upgrade_proposal {
```
wulfraem, line 1: License header is missing. ;)

wulfraem (changes requested):
Would adding a test for an upgrade make sense here?

UMR1352:
> Would adding a test for an upgrade make sense here?

It definitely would but it's not so easy to do. Especially with the limited time we have RN

identity_iota_core/packages/iota_identity/sources/identity.move
UMR1352: The thing is that if you can't upgrade it means that a next version of the package doesn't exist rather than "this identity has already upgraded". I fixed the comment wording but kept the name.

wulfraem: Ah, makes sense. Understood the check `assert!(self.version < PACKAGE_VERSION, ENoUpgrade);` to prevent the update if the identity is already at version `PACKAGE_VERSION`.

wulfraem (approved):

PR #1454, opened 2024-11-27
Borrow an `Identity`'s `ControllerCap` in order to perform operations on sub-Identities

UMR1352:
# Description of change
Introduces a new proposal type `ControllerExecution` that allows controllers of `Identity` **A** to perform controller-specific operations on `Identity` **B**, where **A** is a controller of **B**. To achieve this, `ControllerExecution` lets its executor borrow a specific (selected by its ID) `ControllerCap` for the duration of the transaction. _Move_ code ensures the borrowed `ControllerCap` is correctly returned to its place.

# Motivation
This is needed in order to enable an _"identity owns identity"_ scenario.

## Type of change
- [ ] Documentation Fix

Added an integration test.


identity_iota_core/src/rebased/migration/identity.rs
```diff
@@ -343,17 +349,69 @@ pub async fn get_identity(
     return Ok(None);
   };
 
-  let did = IotaDID::from_alias_id(&object_id.to_string(), client.network());
-  let (id, multi_controller, created, updated) = match unpack_identity_data(&did, &data)? {
-    Some(data) => data,
-    None => {
-      return Ok(None);
-    }
+  let content = data
```
wulfraem, line 352: Was this due to a merge conflict or is there a reason, we can't keep

```rs
  let did = IotaDID::from_alias_id(&object_id.to_string(), client.network());
  let Some((id, multi_controller, created, updated)) = unpack_identity_data(&did, &data)? else {
    return Ok(None);
  };

  let did_doc =
    IotaDocument::from_iota_document_data(multi_controller.controlled_value(), true, &did, created, updated)
      .map_err(|e| Error::DidDocParsingFailed(e.to_string()))?;
```

Also, it seems another merge conflict undid the `let Some(...) = ... else {};` change we did for a previous commit, so the version in the code block above is the intended version. >.>

wulfraem (changes requested):

identity_iota_core/src/rebased/migration/identity.rs
UMR1352: Yeah merge issues :(

UMR1352: Fixed

wulfraem (approved):

PR #1471, opened 2024-12-02
Add devnet package information

UMR1352:
# Description of change
Embed devnet's `IotaIdentity` package metadata into the library and adapt `NetworkName` to work with `ChainIdentifiers` instead of HRP.

### DIFF
--- a/identity_iota_core/src/rebased/iota/well_known_networks.rs
+++ b/identity_iota_core/src/rebased/iota/well_known_networks.rs
@@ -1,7 +1,7 @@
 use iota_sdk::types::base_types::ObjectID;
 use phf::{phf_map, Map};
 
-/// A Mapping `NetworkID` -> metadata needed by the library.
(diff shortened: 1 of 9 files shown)


identity_iota_core/src/rebased/iota/well_known_networks.rs
```diff
@@ -0,0 +1,73 @@
+use iota_sdk::types::base_types::ObjectID;
+use phf::{phf_map, Map};
+
+/// A Mapping `NetworkID` -> metadata needed by the library.
```
wulfraem, line 4: Heavily nitpicking.

Maybe we should use something like "`network_id`" or just "network id" here. Was searching for the type/enum "`NetworkID`" in the code after seeing this. :D

```suggestion
/// A Mapping `network_id` -> metadata needed by the library.
```

wulfraem (approved):
LGTM, just nitpicking. ^^

But the CI clippy might have a word with you. ;)

Issue #1470, opened 2024-12-02
Make sure, that `get_coin_for_transaction` returns a coin object with enoug balance


wulfraem:
## Description
When the function `get_coin_for_transaction` retrieves a `Coin` object, we have to make sure, that this coin has a balance big enough, that we can actually use it.

## Motivation
Getting a `Coin` object with low balance might lead to our transaction failing. Even worse, if we get the same `Coint` object for every `get_coin_for_transaction` call, this would lock out the current user from making writing operations to the chain. Also the current implementation is not paging through results, which should be changed.

## To-do list
Add more if points are missing

- [ ] use paging when fetching coins
- [ ] add checks for a minimum balance (value TBD)
- [ ] add tests

## Change checklist
Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [ ] The feature or fix is implemented in Rust and across all bindings whereas possible.
- [ ] The feature or fix has sufficient testing coverage
- [ ] All tests and examples build and run locally as expected
- [ ] Every piece of code has been document according to the documentation guidelines.
- [ ] If conceptual documentation (mdbook) and examples highlighting the feature exist, they are properly updated.
- [ ] If the feature is not currently documented, a documentation task Issue has been opened to address this.

PR #1472, opened 2024-12-03
Enhance coin retrieval logic in Iota client

itsyaasir:
# Description of change

- Introduced a constant `MINIMUM_BALANCE` to define the minimum balance required for transactions.
- Updated `get_coin_for_transaction` method to iterate through paginated coin data, selecting the coin with the highest balance that meets the minimum requirement.
- Improved error handling to provide clearer messages when no suitable coins are found.

fixes issue #1470 

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/identity_iota_core/src/rebased/client/full_client.rs
+++ b/identity_iota_core/src/rebased/client/full_client.rs
@@ -263,7 +263,6 @@ impl<S> IdentityClient<S> {
     const LIMIT: usize = 10;
     let mut cursor = None;
 
-    // Keep track of the highest balance coin found
     let mut best_coin: Option<Coin> = None;
 
     loop {
@@ -280,7 +279,6 @@ impl<S> IdentityClient<S> {
         };
       }
 
-      // Find the coin with the highest balance in this page
       let page_best_coin = coins.data.into_iter().max_by_key(|coin| coin.balance);
 
       if let Some(coin) = page_best_coin {
@@ -288,7 +286,6 @@ impl<S> IdentityClient<S> {
           return Ok(coin);
         }
 
-        // Update best_coin if this coin has a higher balance
         if best_coin.as_ref().map_or(true, |best| coin.balance > best.balance) {
           best_coin = Some(coin);
         }
@@ -301,7 +298,6 @@ impl<S> IdentityClient<S> {
       cursor = coins.next_cursor;
     }
 
-    // We've searched all pages and found no coin with sufficient balance
     Err(Error::GasIssue(format!(
       "no coin found with minimum required balance of {}",
       MINIMUM_BALANCE
--- a/identity_iota_core/src/rebased/client/full_client.rs
+++ b/identity_iota_core/src/rebased/client/full_client.rs
@@ -263,8 +263,6 @@ impl<S> IdentityClient<S> {
     const LIMIT: usize = 10;
     let mut cursor = None;
 
-    let mut best_coin: Option<Coin> = None;
-
     loop {
       let coins = self
         .coin_read_api()
@@ -272,23 +270,18 @@ impl<S> IdentityClient<S> {
         .await?;
 
       if coins.data.is_empty() {
-        return if let Some(coin) = best_coin {
-          Ok(coin)
-        } else {
-          Err(Error::GasIssue("no coins found for address".to_string()))
-        };
+        return Err(Error::GasIssue("no coins found for address".to_string()));
       }
 
-      let page_best_coin = coins.data.into_iter().max_by_key(|coin| coin.balance);
-
-      if let Some(coin) = page_best_coin {
+      if let Some(coin) = coins.data.into_iter().max_by_key(|coin| coin.balance) {
         if coin.balance >= MINIMUM_BALANCE {
           return Ok(coin);
         }
 
-        if best_coin.as_ref().map_or(true, |best| coin.balance > best.balance) {
-          best_coin = Some(coin);
-        }
+        return Err(Error::GasIssue(format!(
+          "highest balance coin ({}) is below minimum required balance of {}",
+          coin.balance, MINIMUM_BALANCE
+        )));
       }
 
       if !coins.has_next_page {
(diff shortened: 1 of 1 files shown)


identity_iota_core/src/rebased/client/full_client.rs
```diff
@@ -253,18 +259,49 @@ impl<S> IdentityClient<S> {
...
+      let coins = self
+        .coin_read_api()
+        .get_coins(self.sender_address(), None, cursor, Some(LIMIT))
+        .await?;
+
+      if coins.data.is_empty() {
+        return if let Some(coin) = best_coin {
+          Ok(coin)
+        } else {
+          Err(Error::GasIssue("no coins found for address".to_string()))
```
wulfraem, line 278: Could you add the address to the error message? This might be helpful in case multiple accounts are used. ^^

identity_iota_core/src/rebased/client/full_client.rs
```diff
@@ -253,18 +259,49 @@ impl<S> IdentityClient<S> {
...
+      cursor = coins.next_cursor;
+    }
 
-    coins
-      .data
-      .into_iter()
-      .next()
-      .ok_or_else(|| Error::GasIssue("could not find coins".to_string()))
+    Err(Error::GasIssue(format!(
+      "no coin found with minimum required balance of {}",
```
wulfraem, line 302: Could you add the address to the error message? This might be helpful in case multiple accounts are used. ^^

identity_iota_core/src/rebased/client/full_client.rs
```diff
@@ -253,18 +259,49 @@ impl<S> IdentityClient<S> {
     Ok(budget)
   }
 
-  async fn get_coin_for_transaction(&self) -> Result<iota_sdk::rpc_types::Coin, Error> {
-    let coins = self
-      .coin_read_api()
-      .get_coins(self.sender_address(), None, None, None)
-      .await
-      .map_err(|err| Error::GasIssue(format!("could not get coins; {err}")))?;
+  async fn get_coin_for_transaction(&self) -> Result<Coin, Error> {
```
wulfraem, line 262: Hm, if I get the current behavior correct, we are doing the following (correct me, if I'm wrong ^^):

- page through all coins, do this per page
  - if we didn't get coins for current page, check if we have a best coin
    - if we have one, use it
    - if we don't, throw an error
  - if we got coins for current page:
    - get the coin object with the most funds on this page
    - if that coin object is above `MINIMUM_BALANCE`, use it
    - if not, keep it as the current best coin for later

I'm not sure if we should try to let the flow use a coin object, that is below `MINIMUM_BALANCE`. I would have expected that, as we are suspecting this (best) coin to not have enough funds, we would be returning an error with the proper minimum required balance here instead of trying to use it. This is of course a valid possible strategy for using coin object, I just wouldn't have expected it to be the default one. ^^

Also related to the flow, could we actually reach the

```rs
      if coins.data.is_empty() {
        return if let Some(coin) = best_coin {
          Ok(coin)
```

part without some anomaly going on? Most probably overlooking something here, but the following paragraphs describe how I see the flow. ^^;;

In theory, the `!coins.has_next_page` check should prevent us from getting an empty result page for pages 2 and so on. And when the first page is empty, `best_coin` should be empty as well.

The only way to reach an empty result page, that I can see, is to page through the coins, while coins from the "upcoming" result page are burned (e.g. last page has one coin, and this gets burned while we try to fetch its page). Then the `best_coin` would be used.

In all other cases, paging should lead to the `break` due to `!coins.has_next_page` and throw an error. And here I am still unsure if we should use the coin object with funds, that are below the minimum, as written above. ;)

wulfraem (changes requested):

identity_iota_core/src/rebased/client/full_client.rs
```diff
@@ -253,18 +259,42 @@ impl<S> IdentityClient<S> {
...
+
+      if let Some(coin) = coins.data.into_iter().max_by_key(|coin| coin.balance) {
+        if coin.balance >= MINIMUM_BALANCE {
+          return Ok(coin);
+        }
+
+        return Err(Error::GasIssue(format!(
+          "highest balance coin ({}) is below minimum required balance of {}",
+          coin.balance, MINIMUM_BALANCE
+        )));
```
wulfraem, line 284: Do we need this part here? ^^

Thinking about a case, where no coin object on the first page meetings the required `MINIMUM_BALANCE`. Then we wouldn't be returning with the `Ok` from line 278.

In this case, I'd expect the logic to page through the results until a page with a coin with enough funds is found. But with the current lines 281-284 we'd be canceling the paging early.

So, shouldn't the error in line 294 be enough, as it should be triggered after we'd paged through all results and still haven't found a matching coin? ^^

identity_iota_core/src/rebased/client/full_client.rs
wulfraem: And the address? :D 

Same for error message in line 273 ("no coins found for address"). ^^

wulfraem (changes requested):

identity_iota_core/src/rebased/client/full_client.rs
```diff
@@ -47,6 +48,11 @@ use crate::rebased::Error;
 use super::get_object_id_from_did;
 use super::IdentityClientReadOnly;
 
+/// The minimum balance required to execute a transaction.
+///
+/// TBD: A temporary value,
```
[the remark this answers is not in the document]
UMR1352: Honestly I have no idea. Most of the transactions we run are pretty small and end up in the first fee bucket (1000NANO). I'd say we could you 10e6 NANOs which makes it 0.001 IOTA.

identity_iota_core/src/rebased/client/full_client.rs
```diff
@@ -253,18 +259,41 @@ impl<S> IdentityClient<S> {
...
+          "no coins found for address {}",
+          self.sender_address()
+        )));
+      }
+
+      if let Some(coin) = coins.data.into_iter().max_by_key(|coin| coin.balance) {
+        if coin.balance >= MINIMUM_BALANCE {
+          return Ok(coin);
+        }
+      }
```
UMR1352, line 283: See if you like this. It removes one level of nesting
```suggestion
      let Some(coin) = coins.data.into_iter().max_by_key(|coin| coin.balance) else {
        return Err(Error::GasIssue(format!(
          "no coins found for address {}",
          self.sender_address()
        )));
      };
      
      if coin.balance >= MINIMUM_BALANCE {
        return Ok(coin);
      }
```

UMR1352 (approved):

wulfraem (approved):

PR #1496, opened 2025-01-06
Support %-encoded characters in DID URL

### LINKED ISSUE #1492: DIDUrl implementation doesn't support %-encoded path and fragment
## Bug description
A valid DID URL that uses %-encoded characters in either its path or fragment cannot be parsed from its string representation.

UMR1352:
# Description of change
Allow the use of %-encoded characters in DID URLs.

Closes #1492 

## Type of change
- [ ] Documentation Fix

Updated unit tests.

### DIFF
--- a/identity_did/src/did_url.rs
+++ b/identity_did/src/did_url.rs
@@ -537,7 +537,7 @@ pub(crate) const fn is_char_fragment(ch: char) -> bool {
 pub(crate) fn is_valid_percent_encoded_char(s: &str) -> bool {
   let mut chars = s.chars();
   let Some('%') = chars.next() else { return false };
-  chars.take(2).all(|c| c.is_ascii_hexdigit() && !c.is_ascii_lowercase())
+  s.len() == 3 && chars.take(2).all(|c| c.is_ascii_hexdigit())
 }
 
 pub(crate) fn is_valid_url_segment<F>(segment: &str, char_predicate: F) -> bool
--- a/identity_did/src/did_url.rs
+++ b/identity_did/src/did_url.rs
@@ -537,7 +537,7 @@ pub(crate) const fn is_char_fragment(ch: char) -> bool {
 pub(crate) fn is_valid_percent_encoded_char(s: &str) -> bool {
   let mut chars = s.chars();
   let Some('%') = chars.next() else { return false };
-  s.len() == 3 && chars.take(2).all(|c| c.is_ascii_hexdigit())
+  s.len() >= 3 && chars.take(2).all(|c| c.is_ascii_hexdigit())
 }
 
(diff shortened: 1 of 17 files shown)


identity_did/src/did_url.rs
```diff
@@ -535,6 +534,33 @@ pub(crate) const fn is_char_fragment(ch: char) -> bool {
   is_char_path(ch) || ch == '?'
 }
 
+pub(crate) fn is_valid_percent_encoded_char(s: &str) -> bool {
+  let mut chars = s.chars();
+  let Some('%') = chars.next() else { return false };
+  chars.take(2).all(|c| c.is_ascii_hexdigit() && !c.is_ascii_lowercase())
```
wulfraem, line 540: Hm, is there a reason, that we disallow lowercase hex encoding here?

> The uppercase hexadecimal digits 'A' through 'F' are equivalent to
   the lowercase digits 'a' through 'f', respectively.  If two URIs
   differ only in the case of hexadecimal digits used in percent-encoded
   octets, they are equivalent.  For consistency, URI producers and
   normalizers should use uppercase hexadecimal digits for all percent-
   encodings.

(from [RFC](https://www.rfc-editor.org/rfc/rfc3986#section-2.1))

Quick check in browser also considers them equal:

```js
decodeURIComponent('%3b') === decodeURIComponent('%3B')
> true
```

wulfraem, line 540: I think, we should check the length of `chars.take(2)`, as this would return an iterator with a length of 0 or 1, making wrongly percent encoded elements at the end of an input valid, e.g. `relative_url.set_path(Some("/p%A"))` or `relative_url.set_path(Some("/p%"))`.

> Creates an iterator that yields the first n elements, or fewer if the underlying iterator ends sooner.

(from [docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take))

wulfraem (changes requested):

identity_did/src/did_url.rs
```diff
@@ -535,6 +534,33 @@ pub(crate) const fn is_char_fragment(ch: char) -> bool {
   is_char_path(ch) || ch == '?'
 }
 
+pub(crate) fn is_valid_percent_encoded_char(s: &str) -> bool {
+  let mut chars = s.chars();
+  let Some('%') = chars.next() else { return false };
+  s.len() == 3 && chars.take(2).all(|c| c.is_ascii_hexdigit())
```
wulfraem, line 540: Maybe

```suggestion
  s.len() >= 3 && chars.take(2).all(|c| c.is_ascii_hexdigit())
```

as we could have any remaining characters still in the string due to calling it with `segment[i..]`. ^^

eike-hass (approved):

PR #1499, opened 2025-01-15
WASM Move Calls

### LINKED ISSUE #1449: WASM32 - Implement MoveCalls in Typescript
The traits `AssetMoveCalls`, `IdentityMoveCalls` and `MigrationMoveCalls` are defined in _identity_iota_interaction/src/move_call_traits.rs_

We need implementations for these traits using the _Typescript IOTA Client SDK_ similar to those implementations using
the Rust SDK contained in  _identity_iota_core/src/iota_interaction_rust/asset_move_calls.rs_, _.../migration_move_calls.rs_ and _...identity_move_calls.rs_.

The best solution to implement this, might be to implement the move calls in Typescript and call the TS functions via wasm32-bindgen generated bindings.

UMR1352:
# Description of change
Port all calls to move functions to TS using IOTA's TS SDK and bind these move calls through WASM into `iota_interaction_ts` package.

Closes #1449 

## Type of change
- [ ] Documentation Fix

Huge trust me bro as the project is not yet compilable.


bindings/wasm/iota_interaction_ts/src/identity_move_calls.rs
```diff
@@ -44,7 +229,31 @@ impl IdentityMoveCalls for IdentityMoveCallsTsSdk {
...
+      .map_err(WasmError::from)?;
+
+    // Use cell to move `intent_fn` inside `closure` without actually moving it.
+    // This ensures that `closure` is an `impl Fn(..)` instead of `impl FnOnce(..)` like `intent_fn`.
+    let wrapped_intent_fn = Cell::new(Some(intent_fn));
+    let closure = |tx_builder: WasmTransactionBuilder, args: WasmTxArgumentMap| {
+      let mut builder = TransactionBuilderTsSdk::new(tx_builder);
+      let args = serde_wasm_bindgen::from_value(args.into()).expect("failed to convert JS argument map");
+      wrapped_intent_fn.take().unwrap()(&mut builder, &args);
+    };
```
UMR1352, line 249: Check this out! :sunglasses:

chrisgitiota:
I'll definitely need some more hours to deep dive into this PR especially for my personal favorite [check this out](https://github.com/iotaledger/identity.rs/pull/1499#discussion_r1916221291) :+1: :star_struck:

One request from my side would be to change the target branch to `feat/identity-rebased-alpha`. The branch `feat/wasm32-develop-alpha` has already been merged into it and will be deleted probably next week (it only still exists, because I wanted to maintain the purged commit messages for a while).

UMR1352:
Here some more context to better explain the hack used in the implementation of user-driven move-calls.

WASM bindgen allows to pass Rust closures to JS, but it only accepts `Fn` or `FnMut` closures. The intent function we receive from the user is neither of these two types and is in fact a `FnOnce` closure. The reason behind this is that `FnOnce` better fit the use case we have with "intent function" i.e., a **pure** function that is called exactly once.
Therefore, in order to make WASM bindgen happy, the intent function we receive has to become either a `Fn` or `FnMut`. Since we don't really care about modifying the captured context, `Fn` is a better fit.
Changing the type of intent function is not straightforward, the function traits cannot be manually implemented in stable Rust and using `std::mem::transmute` can't really help in this case.
The only way around this is to construct a new closure that implements `Fn` and that calls our `intent_fn` for us. 
Though, this is not as straight forward as doing:
```rust
let f = || intent_fn(...);
```
since `intent_fn` is a `FnOnce`,  invoking it with `(...)` also consumes it (see [`FnOnce::call_once`](https://doc.rust-lang.org/std/ops/trait.FnOnce.html#tymethod.call_once) and therefore the compiler will make `f: FnOnce` as well.

In order for `f` to be `Fn` it has to access all captured variables through shared references only, but `FnOnce` cannot be called through a reference. 

`Cell` to the rescue! By wrapping `intent_fn` in `Cell` we can move it inside closure `f` through a shared reference:
```rust
let wrapped_fn = Cell::new(Some(intent_fn)); // We wrap it in an Option to leverage its Default implementation.
let f = || {
  wrapped_fn
    .take() // only requires shared access to self and moves out its content i.e. Option<impl FnOnce>
    .unwrap() // returns an impl FnOnce
    (...); // invokes and consumes `intent_fn`
};
```
The compiler will infer `impl Fn` for `f` and we can then pass it to WASM bindgen.

wulfraem (approved):

chrisgitiota (approved):

PR #1526, opened 2025-02-12
Re-add Resolver API for WASM builds

wulfraem:
Re-adds examples for did resolution via `Resolver` API:
- `0_basic/2_resolve_did.ts`
- `1_advanced/4_custom_resolution.ts`

`Resolver` API is re-added, but the `attach_iota_handler` and `attach_multiple_iota_handlers` parts are left out of the WASM build.

The Stardust implementation was already not using those two and was directly using a client with respective traits to resolve the DID. Updated implementation follows the same principle, but uses the new API for this.

### DIFF
--- a/bindings/wasm/identity_wasm/examples/src/0_basic/2_resolve_did.ts
+++ b/bindings/wasm/identity_wasm/examples/src/0_basic/2_resolve_did.ts
@@ -25,8 +25,8 @@ export async function resolveIdentity() {
     const iotaClient = new IotaClient({ url: NETWORK_URL });
     const network = await iotaClient.getChainIdentifier();
     const storage = getMemstorage();
-    const [unpublished, vmFragment1] = await createDocumentForNetwork(storage, network);
     const identityClient = await getClientAndCreateAccount(storage);
+    const [unpublished] = await createDocumentForNetwork(storage, network);
 
     // create new identity for this account and publish document for it
     const { output: identity } = await identityClient
--- a/bindings/wasm/identity_wasm/examples/src/0_basic/2_resolve_did.ts
+++ b/bindings/wasm/identity_wasm/examples/src/0_basic/2_resolve_did.ts
@@ -1,51 +1,63 @@
 // Copyright 2020-2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-// TODO:
-// - [ ] clarify if we need/want a resolver example
-// - [ ] clarify if we need/want the AliasOutput -> ObjectID example
-
-
 import {
     CoreDocument,
     DIDJwk,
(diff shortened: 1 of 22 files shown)


bindings/wasm/identity_wasm/examples/src/0_basic/2_resolve_did.ts
```diff
@@ -1,56 +1,84 @@
...
+    const identityClientReadOnly = await IdentityClientReadOnly.createWithPkgId(iotaClient, IDENTITY_IOTA_PACKAGE_ID);
+    // In this case we will only be resolving `IotaDocument` instances, as we don't pass a `handler` configuration.
+    // Therefore we can limit the type of the resolved documents to `IotaDocument` when creating the new resolver as well.
+    const resolverWithReadOnlyClient = new Resolver<IotaDocument>({ client: identityClientReadOnly });
+
+    // And resolve as before.
+    const resolvedViaReadOnly = await resolverWithReadOnlyClient.resolve(did.toString());
+    console.log(`resolverWithReadOnlyClient ${did.toString()} resolves to:\n ${JSON.stringify(resolvedViaReadOnly, null, 2)}`);
+
+    // As our `Resolver<IotaDocument>` instance will only return `IotaDocument` instances, we can directly work with them, w.g.
```
UMR1352, line 77:
```suggestion
    // As our `Resolver<IotaDocument>` instance will only return `IotaDocument` instances, we can directly work with them, e.g.
```

bindings/wasm/identity_wasm/src/resolver/wasm_did_resolution_handler.rs
```diff
@@ -0,0 +1,45 @@
...
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::JsFuture;
+
+use crate::error::JsValueResult;
+use crate::iota::PromiseIotaDocument;
+use crate::iota::WasmIotaDID;
+
+#[wasm_bindgen(typescript_custom_section)]
+const WASM_DID_DID_RESOLUTION_HANDLER: &str = r#"
+interface WasmDidDidResolutionHandler {
```
UMR1352, line 18: Is the double `did` in `WasmDidDidResolutionHandler` intended? Can we rename it to `WasmDidResolutionHandler`?

UMR1352 (approved):

bindings/wasm/identity_wasm/src/resolver/wasm_did_resolution_handler.rs
wulfraem: Ah, definitively not, took me even a bit to find the duplicate here. >.<

Fixed. ^^

PR #1568, opened 2025-02-27
Move: Bare Identity, Removed DeactivateDid Proposal, DID Deletion, Identity Deletion

UMR1352:
# Description of change
- Made Identity's DID Document optional.
- Removed `Proposal<DeactivateDid>`, replacing it with `Proposal<UpdateValue>` with `value = Some([])`.
- Added DID Document deletion through `Proposal<UpdateValue>` with `value = None`. Once an Identity's DID is deleted that Identity won't be able to set another DID Document, enforced through the use of the new Identity flag `deleted_did`.
- Added `Proposal<Delete>` to allow the deletion of an Identity.


identity_iota_core/tests/e2e/identity.rs
```diff
@@ -127,6 +129,8 @@ async fn approving_proposal_works() -> anyhow::Result<()> {
     anyhow::bail!("the proposal is executed");
   };
 
+  dbg!(&proposal);
+
```
wulfraem, line 133: Is this leftover from testing or do we still need it? ^^

```suggestion
```

identity_iota_core/tests/e2e/identity.rs
```diff
@@ -103,6 +103,8 @@ async fn approving_proposal_works() -> anyhow::Result<()> {
     .await?
     .output;
 
+  dbg!("identity created");
+
```
wulfraem, line 107: Is this leftover from testing or do we still need it? ^^

```suggestion
```

wulfraem (changes requested):

wulfraem (approved):

eike-hass (approved):

PR #1527, opened 2025-02-13
Several renamed types in iota_interaction_ts and iota_interaction_rust

chrisgitiota:
# Description of change

This PR comprises the following renamings

bindings/wasm/iota_interaction_ts:
* Renamed `bindings::IotaTransactionBlockResponseAdapter` to `bindings::WasmIotaTxBlockResponseAdapter`

identity_iota_core/src/iota_interaction_rust and
bindings/wasm/iota_interaction_ts:
* Renamed `AdapterNativeResponse` to `NativeTransactionBlockResponse`

## Links to any relevant issues
fixes issue #1514

## Type of change
Add an `x` to the boxes that are relevant to your changes.

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix
- [x] Refactoring  (a non-breaking change which enhances code quality)

## How the change has been tested
* Current E2E tests for non wasm32 targets run but without success
* Successfull wasm32 and non wasm32 build for identity_iota_core:
  wasm32 build: `cargo build --package identity_iota_core --lib --target wasm32-unknown-unknown --no-default-features --features iota-client --features revocation-bitmap`
* NPM build iota_interaction_ts:
   In folder bindings/wasm/iota_interaction_ts:
   * `npm install`
   * `npm run build:nodejs`
* NPM build identity_wasm:
   In folder bindings/wasm/identity_wasm:
   * Prerequisite: NPM build iota_interaction_ts (see above)
   * `npm install`
   * `npm run build:nodejs`

Issue #1514, opened 2025-01-31
[Task] Make IotaTransactionBlockResponse related types exported from iota_interaction_ts and iota_interaction_rust easier to distinguish


chrisgitiota:
## Description
In the current implementation of iota_interaction_ts we have following `pub use`s in bindings/wasm/iota_interaction_ts/src/lib.rs:

```rust
pub use iota_client_ts_sdk::IotaTransactionBlockResponseProvider as IotaTransactionBlockResponseAdapter;

// ...

pub use bindings::IotaTransactionBlockResponseAdapter as AdapterNativeResponse;
```

The name `IotaTransactionBlockResponseAdapter` is used for `bindings::IotaTransactionBlockResponseAdapter` and also as an alias for the exportet `iota_client_ts_sdk::IotaTransactionBlockResponseProvider`.

This does not result in name clashes but is very confusing.

PR #1572, opened 2025-02-28
Rename types in iota interaction ts and iota interaction rust

chrisgitiota:
# Description of change
This PR comprises the following renamings

bindings/wasm/iota_interaction_ts:
* Renamed `bindings::IotaTransactionBlockResponseAdapter` to `bindings::WasmIotaTransactionBlockResponseWrapper` (name is also used for the TS class in iota_client_helpers.ts)

identity_iota_core/src/iota_interaction_rust and
bindings/wasm/iota_interaction_ts:
* Renamed `AdapterNativeResponse` to `NativeTransactionBlockResponse`

* fixes issue #1514
* Replaces closed PR #1527

## Type of change
- [ ] Documentation Fix
- [x] Refactoring  (a non-breaking change which enhances code quality)

* Current E2E tests for non wasm32 targets run as expected
* Successful wasm32 and non wasm32 build for identity_iota_core:
  wasm32 build: `cargo build --package identity_iota_core --lib --target wasm32-unknown-unknown --no-default-features --features iota-client --features revocation-bitmap`
* NPM build iota_interaction_ts:
   In folder bindings/wasm/iota_interaction_ts:
   * `npm install`
   * `npm run build:nodejs`
* NPM build identity_wasm:
   In folder bindings/wasm/identity_wasm:
   * Prerequisite: NPM build iota_interaction_ts (see above)
   * `npm install`
   * `npm run build:nodejs`
* Successfully ran all identity_wasm examples
   In folder bindings/wasm/identity_wasm:
   * Prerequisite: NPM build identity_wasm (see above)
   * npm run example:node <example-name-goes-here>

### DIFF
--- a/bindings/wasm/identity_wasm/src/rebased/identity.rs
+++ b/bindings/wasm/identity_wasm/src/rebased/identity.rs
@@ -9,7 +9,7 @@ use identity_iota::iota::rebased::migration::OnChainIdentity;
 use identity_iota::iota::rebased::transaction::TransactionInternal;
 use identity_iota::iota::rebased::transaction::TransactionOutputInternal;
 use identity_iota::iota::IotaDocument;
-use iota_interaction_ts::AdapterNativeResponse;
+use iota_interaction_ts::NativeTransactionBlockResponse;
 use tokio::sync::RwLock;
 use wasm_bindgen::prelude::*;
 
@@ -218,7 +218,7 @@ impl WasmTransactionOutputInternalOnChainIdentity {
   }
 
   #[wasm_bindgen(getter)]
-  pub fn response(&self) -> AdapterNativeResponse {
+  pub fn response(&self) -> NativeTransactionBlockResponse {
     self.0.response.clone_native_response()
   }
 }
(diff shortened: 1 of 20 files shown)


chrisgitiota:
For unclear reasons the `typescript_type` name and the `pub type` name defined in wasm_bindgen extern "C" scopes must be equal. Otherwise, the JS constructor will not be found in the generated js code.

The reason for this is unclear at the moment. The macro parameter `typescript_type` should "allows us to use typescript declarations in typescript_custom_section as arguments for rust functions". This would imply IMO that the names can be different.

The contrary observed behavior is, that this only works if the `typescript_type` name eqals the `pub type`. Even using wasm_bindgen(js_name = "...") doesn't help. The needed constructor function results from the combination of the 'typescript_type' and 'constructor' wasm_bindgen macro parameters, which may cause this limitation.

An assumption is that this might also be caused by using typescript instead of JS which is not fully supported by wasm-bindgen.

eike-hass (approved):

wulfraem (approved):

Issue #1554, opened 2025-02-26
WASM: use wasm-bindgen `unchecked_return_type` to produce more idiomatic types


UMR1352:
## Description
Replace complex types like `UpdateDidProposalInternalTransactionOutput` with `TransactionOutput<Proposal<UpdateDid> | undefined>`.

## Motivation
wasm-bindgen doens't support generic types, which makes the generated code look very un-idiomatic.

PR #1599, opened 2025-03-06
Better TS types

UMR1352:
# Description of change
Leveraged `wasm_bindgen(unchecked_return_type = ...)` to provide better types in TS bindings.

Closes issue #1554

### DIFF
--- a/bindings/wasm/identity_wasm/src/rebased/identity.rs
+++ b/bindings/wasm/identity_wasm/src/rebased/identity.rs
@@ -87,12 +87,18 @@ impl WasmOnChainIdentity {
     WasmCreateUpdateDidProposalTx::new(self, updated_doc.clone(), expiration_epoch)
   }
 
-  #[wasm_bindgen(js_name = deactivateDid)]
+  #[wasm_bindgen(
+    js_name = deactivateDid,
+    unchecked_return_type = "TransactionInternal<Proposal<UpdateDid> | ProposalOutput<UpdateDid>>",
+  )]
   pub fn deactivate_did(&self, expiration_epoch: Option<u64>) -> WasmCreateUpdateDidProposalTx {
     WasmCreateUpdateDidProposalTx::deactivate(self, expiration_epoch)
   }
 
-  #[wasm_bindgen(js_name = updateConfig)]
+  #[wasm_bindgen(
+    js_name = updateConfig,
+    unchecked_return_type = "TransactionInternal<Proposal<ConfigChange> | ProposalOutput<ConfigChange>>",
+  )]
   pub fn update_config(
     &self,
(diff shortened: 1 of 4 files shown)


bindings/wasm/identity_wasm/src/rebased/identity.rs
```diff
@@ -75,7 +75,10 @@ impl WasmOnChainIdentity {
     serde_wasm_bindgen::to_value(proposals).map_err(wasm_error)
   }
 
-  #[wasm_bindgen(js_name = updateDidDocument)]
+  #[wasm_bindgen(
+    js_name = updateDidDocument,
+    unchecked_return_type = "TransactionInternal<Proposal<UpdateDid> | ProposalOutput<UpdateDid>>",
```
wulfraem, line 80: Would it make sense to add a type to `bindings/wasm/identity_wasm/lib/proposal.ts`, that reflects the outcome of a call, that returns a proposal or its result? Then we could shorten the definitions here a bit and would avoid having the action type twice, while implicitly requiring both places to have the same type. E.g. like:

```ts
// bindings/wasm/identity_wasm/lib/proposal.ts
export type ProposalOrOutput<A extends Action> = Proposal<A> | ProposalOutput<A>;
```

```rs
/// bindings/wasm/identity_wasm/src/rebased/identity.rs
#[wasm_bindgen(
  js_name = updateDidDocument,
  unchecked_return_type = "TransactionInternal<ProposalOrOutput<UpdateDid>>",
)]
```

Or are we expecting to have situations where we could have mixed types of return values? E.g.:

```ts
TransactionInternal<Proposal<UpdateDid> | ProposalOutput<ConfigChange>>",
```

bindings/wasm/identity_wasm/lib/transaction_internal.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright 2021-2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import { IotaTransactionBlockResponse } from "@iota/iota-sdk/client";
+import { IdentityClient } from "~identity_wasm";
+
+export interface TransactionInternalOutput<T> {
```
wulfraem, line 7: Do we need the "internal" wording in TypeScript or should it just be `Transaction` here?

wulfraem (changes requested):

bindings/wasm/identity_wasm/lib/transaction_internal.ts
UMR1352: I don't like it but we've used the Internal wording in the Rust APIs as well so we better stick to it

eike-hass (approved):

PR #1655, opened 2025-04-25
Add support for `DelegationToken` and `ControllerToken`

### LINKED ISSUE #1440: Add library support for `ControllerCap`, `DelegationToken` and related operations
## Description
Enhance the operations users can perform though our library with regards to `ControllerCap` and `DelegationToken` related operations

## Motivation
Simplifying Identity's AC related operation without having user resort to making _Move_ calls themselves.

## To-do list
- [ ] Handling of `DelegationToken`
- [ ] Code a way to reason about `ControllerCap` and `DelegationToken` generically (enum?)
- [ ] Update APIs that add controllers to also require a `can_delegate` parameter
- [ ] Revocation and unrevocation of `DelegationToken`

UMR1352:
# Description of change
All `Identity`'s APIs require a token that authenticates the caller as an `Identity`'s controller. Access to such APIs though is not strictly limited to controllers but it can also be granted to other entities through the *delegation* of a controller's token `ControllerCap`. Delegating a controller's access requires minting a new kind of token called `DelegationToken`, that allows its bearer to act in the stead of the controller that minted it, with either full or limited permissions.

Now `ControllerToken` doesn't represent just a `ControllerCap` - as it was before this PR - but it represents *either* a `ControllerCap` or a `DelegationToken`.

Resolves #1440 

## Type of change
- [ ] Documentation Fix

New E2E test for all operations that use `DelegationToken` + doc tests.

### DIFF
--- a/identity_iota_core/packages/iota_identity/sources/proposals/config.move
+++ b/identity_iota_core/packages/iota_identity/sources/proposals/config.move
@@ -1,108 +1,108 @@
 // Copyright (c) 2024 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-module iota_identity::config_proposal {
-    use iota_identity::multicontroller::Multicontroller;
-    use iota_identity::controller::DelegationToken;
-    use iota::vec_map::VecMap;
+module iota_identity::config_proposal;
 
-    const ENotMember: u64 = 0;
-    const EInvalidThreshold: u64 = 1;
+use iota::vec_map::VecMap;
+use iota_identity::controller::DelegationToken;
+use iota_identity::multicontroller::Multicontroller;
 
-    public struct Modify has store, drop {
-        threshold: Option<u64>,
-        controllers_to_add: VecMap<address, u64>,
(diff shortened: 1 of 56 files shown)


identity_iota_core/packages/iota_identity/sources/proposals/config.move
```diff
@@ -21,59 +21,38 @@ public fun propose_modify<V>(
...
-    };
-    let (_, mut voting_powers) = controllers_to_add.into_keys_values();
-    let mut voting_power_increase = 0;
-    while (!voting_powers.is_empty()) {
-        let voting_power = voting_powers.pop_back();
-
-        voting_power_increase = voting_power_increase + voting_power;
-    };
-    voting_powers.destroy_empty();
+    });
```
wulfraem: Most probably missing something, but was wondering why this part isn't using `fold!` to accumulate the max_votes, similar to how the next two sections do.

wulfraem (changes requested):

PR #1664, opened 2025-05-07
Feat/package registry

UMR1352:
# Description of change
Introduce `PackageRegistry`, a singleton Map that keeps track of the published IotaIdentity for each network (devnet, testnet, mainnet, and custom networks).
Identity's transactions and operations use this mechanism to retrieve the package ID for the network the executing client is connected to without relying on any other input from the client itself.


identity_iota_core/src/rebased/iota/package.rs
```diff
@@ -0,0 +1,226 @@
...
+use crate::rebased::client::CoreClientReadOnly;
+use crate::rebased::Error;
+
+pub(crate) const MAINNET_CHAIN_ID: &str = "6364aad5";
+
+type PackageRegistryLock = RwLockReadGuard<'static, PackageRegistry>;
+type PackageRegistryLockMut = RwLockWriteGuard<'static, PackageRegistry>;
+
+static IDENTITY_PACKAGE_REGISTRY: LazyLock<RwLock<PackageRegistry>> = LazyLock::new(|| {
+  let move_lock_content = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/packages/iota_identity/Move.lock"));
```
wulfraem, line 28: As we're compiling for windows as well in the CI, should we consider Windows paths here as well?

> The provided path is interpreted in a platform-specific way at compile time. So, for instance, an invocation with a Windows path containing backslashes \ would not compile correctly on Unix.

[`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html)

identity_iota_core/src/rebased/iota/package.rs
```diff
@@ -0,0 +1,226 @@
...
+  pub published_version: u64,
+}
+
+impl Metadata {
+  /// Create a new [Metadata] assuming a newly published package.
+  pub(crate) fn from_package_id(package: ObjectID) -> Self {
+    Self {
+      original_published_id: package,
+      latest_published_id: package,
+      published_version: 1,
```
wulfraem, line 97: Just wondering if we should set this to 1 here.

We don't seem to be using the version at the moment, but I'm guessing it is parsed and might be used in the future, so we parse it from the `.lock` file.

`IdentityClientReadOnly::new_with_pkg_id` uses the current function (`from_package_id`) to allow passing a package into the client, but then it is always set to 1. Was just wondering if we currently have the two possible setups:

- we parsed the `.lock` file and `published_version` is valid
- we used `new_with_pkg_id` and it is set to 1 and might be invalid

So I was thinking, I'm either missing an update for the 1, are planning to add this in a follow-up issue, or if we should make `published_version` optional and just set it to `None` here.

identity_iota_core/src/rebased/client/read_only.rs
```diff
@@ -70,97 +69,106 @@ impl IdentityClientReadOnly {
...
+      let package_id = package_registry
+        .package_id(&network)
+        .ok_or_else(|| {
+        Error::InvalidConfig(format!(
+          "no information for a published `iota_identity` package on network {network}; try to use `IdentityClientReadOnly::new_with_package_id`"
+        ))
+      })?;
+      let maybe_alias = package_registry
+        .chain_alias(&chain_id)
+        .and_then(|alias| NetworkName::try_from(alias).ok());
```
wulfraem, line 110: Could we move this to the default case in the match, as it seems to be only used there?

identity_iota_core/src/rebased/iota/package.rs
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  #[tokio::test]
+  async fn can_connect_to_mainnet() -> anyhow::Result<()> {
+    let iota_client = IotaClientBuilder::default().build_mainnet().await?;
+    let _identity_client = IdentityClientReadOnly::new(iota_client).await?;
+
+    Ok(())
+  }
+}
```
wulfraem, line 226: Seeing these aligned that neatly, I was wondering if we could add a more convenient way for local development: using `IotaClientBuilder::default().build_localnet().await?;` and pulling the package id from the `.lock` file instead of the ENV.

But I guess, this might become annoying and prone to errors, as every PR seems to overwrite the `localnet` values in the `.lock` file anyway. >.<

identity_iota_core/src/rebased/migration/registry.rs
```diff
@@ -58,3 +116,36 @@ pub async fn lookup(id_client: &IdentityClientReadOnly, alias_id: ObjectID) -> R
...
+    .event_api()
+    .query_events(event_filter, None, Some(1), false)
+    .await
+    .map_err(|e| Error::Client(e.into()))?
+    .data;
+  let event = if !returned_events.is_empty() {
+    returned_events.swap_remove(0)
+  } else {
+    return Err(Error::NotFound(format!(
+      "No MigrationRegistyrCreated event on network {}",
```
wulfraem, line 143: ^^

```suggestion
      "No MigrationRegistryCreated event on network {}",
```

identity_iota_core/src/rebased/client/read_only.rs
```diff
@@ -70,97 +69,106 @@ impl IdentityClientReadOnly {
...
-      }
-    }
+  /// Attempts to create a new [`IdentityClientReadOnly`] from a given [`IotaClient`].
+  ///
+  /// # Failures
+  /// This function fails if the provided `iota_client` is connected to an unrecognized
+  /// network.
+  ///
+  /// # Notes
+  /// When trying to connect to a local or unofficial network prefer using
```
wulfraem, line 86: Maybe?

```suggestion
  /// When trying to connect to a local or unofficial network, prefer using
```

wulfraem (changes requested):

identity_iota_core/src/rebased/iota/package.rs
UMR1352: That would be great and I tried doing that only to realize that I cannot read files when targeting WASM.

wulfraem: Ah yeah, makes sense. >.<

wulfraem (approved):

PR #1674, opened 2025-06-05
Package's history

UMR1352:
# Description of change
Make the client be aware of the entire package's history for the chain it is connected to. This is needed to allow operations for which the package-id part of an object's type cannot be assumed to be the current active package ID.

### DIFF
--- a/identity_iota_core/src/rebased/client/read_only.rs
+++ b/identity_iota_core/src/rebased/client/read_only.rs
@@ -56,7 +56,7 @@ impl IdentityClientReadOnly {
   pub fn package_id(&self) -> ObjectID {
     *self
       .package_history
-      .first()
+      .last()
       .expect("at least one package exists in history")
   }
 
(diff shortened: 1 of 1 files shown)


identity_iota_core/src/rebased/client/read_only.rs
```diff
@@ -58,8 +53,11 @@ impl IdentityClientReadOnly {
   /// Returns `iota_identity`'s package ID.
   /// The ID of the packages depends on the network
   /// the client is connected to.
-  pub const fn package_id(&self) -> ObjectID {
-    self.iota_identity_pkg_id
+  pub fn package_id(&self) -> ObjectID {
+    *self
+      .package_history
+      .first()
```
wulfraem, line 59: Most probably a dumb question, but why is the first value used in this `package_id`?

`package_history` comes from `PackageRegistry`, where `.last` is used to fetch the value for its own `package_id` function.

Also, `new_with_pkg_id` is using `insert_new_package_version` with the comment "force it at the top of the list", while using `Vec::push` ("Appends an element to the back of a collection.").

Most probably missing some flipping of the order or just looking at the wrong elements. >.<

wulfraem (changes requested):
Most probably just a bit of confusion on my side, and a nitpick suggestion. :D

identity_iota_core/src/rebased/client/read_only.rs
wulfraem: Ah, phew, was really assuming, I'm overlooking a reverse somewhere, but couldn't find it. :D

PR #1679, opened 2025-06-12
WASM override default sync contructors

### LINKED ISSUE #1596: add synchronous contstructors, that throw an error and point to the asynchronous `create` functions for the two identity clients

UMR1352:
# Description of change
`IdentityClient` can only be built though an async operation but the default constructor - inherited by Object - allows to create an instance, although an invalid one. This behavior is deceptive and hence we override the constructor with one that always fails with an error that mentions the correct way to instantiate the client.

Closes #1596

### DIFF
--- a/bindings/wasm/identity_wasm/src/rebased/wasm_identity_client_read_only.rs
+++ b/bindings/wasm/identity_wasm/src/rebased/wasm_identity_client_read_only.rs
@@ -51,6 +51,11 @@ pub struct WasmIdentityClientReadOnly(pub(crate) IdentityClientReadOnly);
 // builder related functions
 #[wasm_bindgen(js_class = IdentityClientReadOnly)]
 impl WasmIdentityClientReadOnly {
+  #[wasm_bindgen(constructor, skip_typescript)]  
+  pub fn _new() -> JsError {
+    JsError::new("cannot build an instance of `IdentityClientReadOnly` through its default sync constructor. Use `IdentityClientReadOnly.create` instead.")
+  }
+
   #[wasm_bindgen(js_name = create)]
   pub async fn new(iota_client: WasmIotaClient) -> Result<WasmIdentityClientReadOnly, JsError> {
     let inner_client = IdentityClientReadOnly::new(iota_client).await?;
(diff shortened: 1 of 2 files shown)


bindings/wasm/identity_wasm/src/rebased/wasm_identity_client_read_only.rs
```diff
@@ -51,6 +51,11 @@ pub struct WasmIdentityClientReadOnly(pub(crate) IdentityClientReadOnly);
 // builder related functions
 #[wasm_bindgen(js_class = IdentityClientReadOnly)]
 impl WasmIdentityClientReadOnly {
+  #[wasm_bindgen(constructor, skip_typescript)]
+  pub fn _new() -> JsError {
+    JsError::new("cannot build an instance of `IdentityClientReadOnly` through its default sync constructor. Use `IdentityClientReadOnly.create` instead.")
```
wulfraem, line 56: Today I learned: js/ts constructors can be even weirder, than expected. :D

ts forbids most of the unrelated return values in constructors (while js is still pretty open about them >.<), but still allows to _return_ an `Error`. oO

And that's what happens with the current implementation. An `Error` is _returned_ but not _thrown_.

`wasm-bindgen` wasn't too happy about the `()` when changing return type to `Result<(), JsError>` (`error: The constructor for class `IdentityClientReadOnly` tries to return a JS primitive type, which would cause the return value to be ignored. Use a builder instead (remove the `constructor` attribute).`). So changed it to `WasmIdentityClientReadOnly`, which should reflect the expected behavior of a constructor.

We could also add a deprecation warning, even if the wording is a bit off here, but it would show a warning in a ts IDE and possibly avoid errors even before compilation (had to re-enable ts for the function for this).

So ended up with this:

```suggestion
  /// @deprecated Use `IdentityClientReadOnly.create` instead.
  #[wasm_bindgen(constructor, skip_typescript)]
  pub fn _new() -> Result<WasmIdentityClientReadOnly, JsError> {
    Err(JsError::new("cannot build an instance of `IdentityClientReadOnly` through its default sync constructor. Use `IdentityClientReadOnly.create` instead."))
  }
```

Not sure about re-enabling ts and the deprecation warning, you can drop it if you like, but the same pattern should be applied for both clients. ^^

PR #1703, opened 2025-08-07
Automatic package history management with build.rs

UMR1352:
# Description of change
build.rs runs when iota_identity's Move.lock changes and generates the source file package_registry.rs


UMR1352:
Ehm @wulfraem I know this is dumb, let's consider this more of a research thingy I don't actually expect to merge this

PR #1713, opened 2025-08-27
Fix `SdJwtVcPresentationBuilder::conceal`

UMR1352:
# Description of change
Fixes a bug that made `SdJwtVcPresentationBuilder::conceal` not working properly.

### DIFF
--- a/identity_credential/src/sd_jwt_vc/presentation.rs
+++ b/identity_credential/src/sd_jwt_vc/presentation.rs
@@ -23,9 +23,11 @@ impl SdJwtVcPresentationBuilder {
   /// Prepare a presentation for a given [`SdJwtVc`].
   pub fn new(token: SdJwtVc, hasher: &dyn Hasher) -> Result<Self> {
     let SdJwtVc {
-      sd_jwt,
-      parsed_claims: vc_claims,
+      mut sd_jwt,
+      parsed_claims: mut vc_claims,
     } = token;
+    // Make sure to set the parsed claims back into the SD-JWT Token.
+    std::mem::swap(sd_jwt.claims_mut(), &mut vc_claims.sd_jwt_claims);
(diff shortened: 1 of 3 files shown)


identity_credential/src/sd_jwt_vc/presentation.rs
```diff
@@ -23,9 +23,11 @@ impl SdJwtVcPresentationBuilder {
   /// Prepare a presentation for a given [`SdJwtVc`].
   pub fn new(token: SdJwtVc, hasher: &dyn Hasher) -> Result<Self> {
     let SdJwtVc {
-      sd_jwt,
-      parsed_claims: vc_claims,
+      mut sd_jwt,
+      parsed_claims: mut vc_claims,
     } = token;
+    // Make sure to set the parsed claims back into the SD-JWT Token.
```
wulfraem, line 29: Hm, the code looks correct and seems to work, but could we add a small note **why**, we have to swap the claims between `new` and `finish`? TBH, a bit confused about this. >.<

Also, pretty-sure-that-it-does-not-apply-but-wanted-to-ask-as-it-would-be-nagging-on-me-otherwise-question: The warning at `SdJwt::claims_mut` does not apply, right?

```Rust
  /// ## Warning
  /// Modifying the claims might invalidate the signature.
  /// Use this method carefully.
```

wulfraem (approved):
Great stuff. ^^

Just a small documentation update (that might be helpful in the future) and a just-to-make-sure question. :D

identity_credential/src/sd_jwt_vc/presentation.rs
UMR1352: In this case no, it won't invalidate the signature as we just swap them without modifying their order nor their values. Thanks for the review !

PR #1827, opened 2026-06-15
AA-enabled Identity

UMR1352:


identity_iota_core/packages/iota_identity/sources/identity_v2/identity_v2.move
```diff
@@ -0,0 +1,651 @@
...
+#[error(code = 3)]
+const EInsufficientApprovals: vector<u8> = b"Insufficient approvals for this transaction";
+#[error(code = 4)]
+const EInvalidControllerSignature: vector<u8> = b"Invalid controller signature";
+#[error(code = 5)]
+const EUnsupportedKeyType: vector<u8> = b"Unsupported key type";
+#[error(code = 6)]
+const ETransactionDigestMismatch: vector<u8> = b"Transaction digest mismatch";
+#[error(code = 7)]
+const EInsufficientPermissions: vector<u8> = b"Controller does not have sufficient permissions";
```
wulfraem, line 37: Would it make sense to split this error into `EInsufficientPermissions` and `EInsufficientWeight`? Both values behave differently, and knowing which of them is off might help to find out why an action was rejected.

identity_iota_core/packages/iota_identity/sources/identity_v2/config.move
```diff
@@ -0,0 +1,239 @@
...
+        assert!(!seen.contains(addr), EControllersMustNotContainDuplicated);
+        seen.push_back(*addr);
+    });
+}
+
+public fun is_subset_of(permissions: u64, other_permissions: u64): bool {
+    (permissions & other_permissions.bitwise_not()) == 0
+}
+
+public macro fun can_propose_tx(): u64 {
```
wulfraem, line 193: Readability wise, would it make sense to move the permission related components into their own file? IMO `permission.can_propose_tx()` might be clearer than the current `config.can_propose_tx()` (initially thought, it would check the current config, whether the caller can propose a tx or something and not return the value for such a check >.<).

While on that topic already (^^;;), would have expected a function starting with `can_` to return a boolean value. Maybe we could drop the `can_` to give it a more "enum-like" feeling, resulting in `permission.propose_tx()`. This would also fit the existing `admin()` function naming.

identity_iota_core/packages/iota_identity/sources/identity_v2/config.move
```diff
@@ -0,0 +1,239 @@
...
+    1 << 9
+}
+
+public macro fun can_transfer_asset(): u64 {
+    1 << 10
+}
+
+public macro fun admin(): u64 {
+    1 << 63
+}
```
wulfraem, line 239: Was wondering if we should add something like `all_permissions`, that returns `std::u64::max_value!())`. Might be used in the migration and the v2 `new`.

identity_iota_core/packages/iota_identity/sources/identity_v2/identity_v2.move
```diff
@@ -0,0 +1,651 @@
...
+                assert_permissions(permissions, config::can_delete_did!());
+            } else if (move_call_is(&move_call, b"add_controller")) {
+                validate_add_controller_call(&move_call, inputs, permissions, weight);
+            } else if (move_call_is(&move_call, b"update_controller")) {
+                validate_update_controller_call(&move_call, inputs, config, permissions, weight);
+            } else if (move_call_is(&move_call, b"remove_controller")) {
+                validate_remove_controller_call(&move_call, inputs, config, permissions, weight);
+            } else if (move_call_is(&move_call, b"update_threshold")) {
+                assert_permissions(permissions, config::can_update_threshold!());
+            } else {}
```
wulfraem, line 564: Putting a reminder here, to revisit the question about whether we should add permission checks for executing transactions with the identity, e.g. control all owned assets or having all controllers being allowed to call all functions of all "owned" (actually owned by the identity or the identity being allowed to do something in a shared object) objects of the identity.

identity_iota_core/packages/iota_identity/sources/identity.move
```diff
@@ -688,6 +700,71 @@ module iota_identity::identity {
...
+            let proposal = aa_migration::new();
+            df::add(&mut self.id, AAMigrationKey {}, proposal);
+            df::borrow_mut(&mut self.id, AAMigrationKey {})
+        };
+
+        proposal.insert_controller(ctx.sender(), self.did_doc.voting_power(token.controller()));
+        cap.put_back(token, borrow);
+    }
+
+    public fun execute_aa_migration(
```
wulfraem, line 727: (Reminder) What about the objects owned by the identity?

IIRC, we can't just provide them as arguments here, so this should be responsibility of the Rust API, right?

Should work, but a check or something here would be really nice, but I don't think, that's an option. >.<

identity_iota_core/packages/iota_identity/sources/identity_v2/migration_tests.move
```diff
@@ -0,0 +1,113 @@
...
+
+const CONTROLLER_A: address = @0x1;
+const CONTROLLER_B: address = @0x2;
+const CONTROLLER_C: address = @0x3;
+
+// Creates an identity with 3 controllers threshold is 2:
+//    - Controller A: address 0x1, weight 1;
+//    - Controller B: address 0x2, weight 1;
+//    - Controller C: address 0x3, weight 2;
+fun make_identity(clock: &Clock, ctx: &mut TxContext) {
```
wulfraem: Maybe? ^^

```suggestion
/// Creates an identity with 3 controllers threshold is 2:
///    - Controller A: address 0x1, weight 1;
///    - Controller B: address 0x2, weight 1;
///    - Controller C: address 0x3, weight 2;
```

identity_iota_core/packages/iota_identity/sources/identity_v2/config.move
```diff
@@ -0,0 +1,239 @@
...
+
+/// Create a new `IdentityConfig`.
+/// - `addresses` and `weights` must have the same length.
+/// - `addresses` must not contain duplicates.
+/// - `threshold` must be greater than zero and less than or equal to the sum of all weights.
+public fun new(
+    addresses: vector<address>,
+    weights: vector<u64>,
+    permissions: vector<u64>,
+    threshold: u64,
```
wulfraem, line 69: Not sure if we actually need this, but wanted to pin it here before I forget it. :D 

Would it make sense to use separate thresholds per permission? But then again, this would probably require permission based weights as well.

This might allow distributing responsibilities and powers better, and for example deleting a DID might require a higher threshold, where all members must agree with the same weights, while updating DID documents could be fastlaned by responsible parties.

But it would complicate the API for creating identities and adding/updating controllers. One way would be to add a default threshold/weight and a map to override the values for both functions. E.g. call could provide a default threshold of 2, but require a much higher weight for deleting the identity. Similarly, a controller could have a default weight of 1, and a weight of 5 for the responsibility to update the DID document.

Just as mentioned, not sure if we need this, but wanted to chip better sooner than later, as adding it might require a separate migration. >.<

identity_iota_core/packages/iota_identity/sources/identity_v2/migration_registry.move
```diff
@@ -0,0 +1,42 @@
...
+
+public fun get(self: &AAMigrationRegistry, old_id: ID): Option<ID> {
+    if (self.migrated_identities.contains(old_id)) {
+        option::some(*self.migrated_identities.borrow(old_id))
+    } else {
+        option::none()
+    }
+}
+
+public fun delete_controller_cap(self: &AAMigrationRegistry, cap: ControllerCap) {
```
wulfraem, line 35: Is this function only used in tests?

wulfraem (changes requested):

